PR #1, opened 2022-06-30
Add ard folder and readme

willmcb:
Adds ADR director in `/docs` to hold ADRs

Copy to point devs to use ADR tools when creating new records 👍


willmcb:
> We can use `adr-tools` as a recommended way to add ADRs. Easy install with `brew install adr-tools` and then when a new ADR needs to be recorded we can use:
> 
> `adr new Integrate with Foobar` and will create a new, numbered ADR file and open it in your editor of choice (as specified by the VISUAL or EDITOR environment variable).
> 
> Also supports superseding previous ADRs. More details in their repo: https://github.com/npryce/adr-tools
> 
> By using that tool to initialise the ADR directory it already records a very first 0001 ADR with the basic template 🥳
> 
> We could keep all documentation together in the `/docs` directory. For example we can initialise the ADR dir as this: `adr init docs/architecture/decisions` or `adr init docs/architecture-decisions`

I think this sounds like a sensible approach - I will make the changes and add a note to the application README outlining the use of ADR tools 👍

zheileman (approved):
🎉

PR #2, opened 2022-07-04
CAD-325 Absolute minimum Hello world app

zheileman:
This is just a rack app (literally a one liner) that can be dockerized in order to have something basic for a CI/CD pipeline.

Run it locally with: `rackup hello.ru`
Or build in a container with: `docker-compose up`

It is intended to be disposable, this will be replaced with a proper rails app soon.


willmcb (approved):
Nice looks good - tested locally and works well 👍

PR #3, opened 2022-07-05
CAD-323 Add initial kubernetes manifest files

zheileman:
This is the basic configuration to be able to deploy the Hello World app.

I'm not adding a secrets file just yet as that can be a bit more involved with git-crypt, but that will be next, and will also allow us to set the http basic-auth credentials.

Note: once we have a Rails app we need to replace the port `9292` with the standard `3000`, but other than that, and maybe adding the RDS endpoint, this should all continue to be valid.


willmcb (approved):
LGTM 👍

PR #5, opened 2022-07-06
Update README with info about Docker and Kubernetes

zheileman:
Just to have some bits to structure the README, and we can add some more as we go. For example when we have git-crypt, or when the local docker containers get more complicated with local env configuration.

Happy to add/remove/change any other bits!


willmcb (approved):
👍 LGTM

PR #7, opened 2022-07-07
CAD-323 Add secrets manifest and git-crypt action

zheileman:
This is really the last missing bit to finalise ticket CAD-323. We needed a way to encrypt the `secrets.yml` manifest and to decrypt it when applying the manifests in the deploy pipeline.

We use `git-crypt` as it is a well tested solution across the organisation.

In the pipeline, a new step has been introduced in the deploy job that will perform the git-crypt unlock making use of a symmetric key exported and added to the repository secrets with name `GIT_CRYPT_KEY`.

Additionally, we are pinning 3rd party actions to their specific SHA, instead of a tag, for hardening purposes.


.github/workflows/test-build-deploy.yml
```diff
@@ -54,7 +54,7 @@ jobs:
 
       - name: Push to ECR
         id: ecr
-        uses: jwalton/gh-ecr-push@v1.3.5
+        uses: jwalton/gh-ecr-push@3197046958514d9f2b97073bb797ca426db8061f
```
willmcb, line 57: 👍

willmcb (approved):
👍  Nice - LGTM

PR #11, opened 2022-07-08
CAD-355 Some more tweaks and cleanup of scaffolding

zheileman:
Reorganised the `.env` files to follow more tightly the recommendations of how the different files override each other envs variables, according to their documentation.

Added a Brewfile which is similar to a Gemfile, doing `brew bundle` will install any package easily, for now I just added the `adr-tools` but will be more in the future.

Removed unneeded default files and disabled some Rails functionality that we are not going to use for now.

Exposed the RDS instance URL to the pods otherwise they can't connect to the database.

Added database to the docker-compose as well.

Some other miscellanea minor tweaks.


config/environments/test.rb
```diff
@@ -53,7 +53,7 @@
   config.active_support.disallowed_deprecation_warnings = []
 
   # Raises error for missing translations.
-  # config.i18n.raise_on_missing_translations = true
+  config.i18n.raise_on_missing_translations = true
```
willmcb, line 56: Nice 👍

willmcb (approved):
LGTM 👍

PR #12, opened 2022-07-08
CAD-355 Setup asset pipeline CSS and govuk-frontend

zheileman:
Added the govuk template markup and the npm package.

To test it, once you checkout the branch, run the following:

`bundle`
`yarn`

and then:

`bin/dev`

This will launch foreman with the puma server and the dartsass:watch processes, any changes to the CSS will compile in real time.

If all works you should see:

Screenshot 2022-07-08 at 17 30 35


willmcb (approved):
LGTM 👍

PR #13, opened 2022-07-11
Add govuk-frontend javascript as import-map

zheileman:
We are pinning the `govuk-frontend` JS to the `importmap.rb` and
also ensuring same version is pinned as a postinstall yarn script.

Added instructions on how to build and run the app to the README.


willmcb (approved):
👍

PR #14, opened 2022-07-12
CAD-360 Setup the test framework and linters

zheileman:
For now this includes:

* rspec, and coverage
* rubocop
* brakeman

It declares a default `rake` task to run all 3, and also the individual tasks.

Rubocop enabled with almost all cops enabled by default except some sensible defaults turned off. Also will use the additional cops from `rubocop-rails` and `rubocop-performance`.

Added info to the README.

Enabled the tests in the Github workflow.

I've not included cucumber here as to not overcomplicate things, I think cukes can be incorporated a bit later when we actually have some real pages or interactions.


lib/tasks/brakeman.rake
```diff
@@ -0,0 +1,5 @@
+task :brakeman do
+  sh <<end
+mkdir -p tmp && (brakeman --quiet --output tmp/brakeman.out --exit-on-warn && echo "no warnings or errors") || (cat tmp/brakeman.out; exit 1)
+end
```
willmcb, line 4: Is there an extra 'end' here?

spec/rails_helper.rb
```diff
@@ -0,0 +1,31 @@
+require_relative "../config/environment"
+abort("The Rails environment is running in production mode!") if Rails.env.production?
+
+require 'spec_helper'
+require 'rspec/rails'
+
+# Checks for pending migrations and applies them before tests are run.
+# If you are not using ActiveRecord, you can remove these lines.
+begin
```
willmcb, line 9: Nice

lib/tasks/brakeman.rake
zheileman: No, the first end is the closure to `<<end`, the second is the closure to the task itself 😄

willmcb (approved):
👍 LGTM

PR #15, opened 2022-07-12
Ensure env is correct for rspec and rubocop tasks

zheileman:
Follow-up to PR #14.

Some tasks rely on gems only included in specific Gemfile groups (like `development` and/or `test`).

On building the docker image, we exclude those groups from the Gemfile (as we run on `production`).

We need to guard these tasks against trying to require unavailable gems, or the build will fail.


willmcb (approved):
👍

PR #17, opened 2022-07-14
Add feature flag capability

willmcb:
Adds basic feature flag functionality to the application.

**Notes for reviewer**
- Currently there are only two environments so I have confined abilities to reflect that. As we add more ENVs we can add methods to encapsulate that we are running the code on them etc.
- I adapted [code from here](https://github.com/ministryofjustice/correspondence_tool_staff/blob/main/lib/feature_set.rb) to create this functionality. This code also has the functionality to pass a block in in a `safe?` method that will only run things on a safe environment. I have not added this into this PR as I thought it was probably not needed at this point. If we think this is something that would be useful I could add it in in a future PR.
- I have used the [config gem](https://github.com/rubyconfig/config), following the pattern [used in this app](https://github.com/ministryofjustice/correspondence_tool_staff) - which has created a set of settings files, we may view some of these as bloat, so happy to remove some of them if not needed. Currently the only one of the settings files used by this code is `config/settings.yml`

**Relevant ticket**
[CRIMAP](https://dsdmoj.atlassian.net/browse/CRIMAP-4)

**Manually testing the code:**
- you can change the `test_on` / `test_off` test features in the `config/settings.yml` I have added a little temporary logic toe the `index.html.erb` that we can remove when we build out this page. You will see the text "Features are enableable!" and "Features are disableable!" show and hide depending on the values in the `settings.yml`


zheileman (approved):
🎉

PR #18, opened 2022-07-18
CRIMAP-13 Add DfE Form Builder and configuration

zheileman:
In preparation for the step-by-step journey and the different form elements, first we add the Form Builder which is 100% compatible with GOV.UK Frontend.

https://govuk-form-builder.netlify.app

https://github.com/DFE-Digital/govuk-formbuilder

Some basic configuration to customise the default tags and size (these can be override on individual elements).

Also added some basic helpers to the `ApplicationHelper` to deal with the service name and the title of the pages.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -20,6 +20,9 @@ AllCops:
 ## Customization for this project ##
 ####################################
 
+# Disabled cops
+###############
+
 Style/Documentation:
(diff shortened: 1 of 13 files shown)


willmcb (approved):
LGTM 👍

willmcb (approved):
👍

PR #19, opened 2022-07-18
Add healthcheck endpoint

willmcb:
Add healthcheck endpoint to the application

At the moment it only checks there is a DB connection available, but it should be simple to adapt to add other checks as needed.

Are there other checks that we should add? Or other endpoints?

### DIFF
--- a/app/controllers/healthcheck_controller.rb
+++ b/app/controllers/healthcheck_controller.rb
@@ -1,5 +1,4 @@
-# rubocop:disable Rails/ApplicationController
-class HealthcheckController < ActionController::Base
+class HealthcheckController < BareApplicationController
   def show
     render json: { healthcheck: healthcheck_result }, status: http_status
   end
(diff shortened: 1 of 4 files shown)


app/controllers/healthcheck_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def green?
+    # add more checks as needed:
+    # database_connected? && redis_connected etc...
+    database_connected?
+  end
+
+  def database_connected?
+    ApplicationRecord.connection.select_value('SELECT 1') == 1
```
zheileman, line 24: I think we can use `ActiveRecord::Base.connection.active?` for the same purpose, that's what I've used in the past, although unsure what is more optimal or recommended tho 🤔

willmcb: That sounds cleaner to me I'll amend 👍

PR #22, opened 2022-07-19
CRIMAP-11 Add yes-no value object

zheileman:
This value object along with its custom Type can be used for basic yes-no step questions.

It will be used for the first time in a follow-up PR to implement the `client has a partner` question.


willmcb (approved):
Nice 👍

PR #25, opened 2022-07-19
CRIMAP-31 Add ping endpoint to expose build info

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/CRIMAP-31

This endpoint `/ping` exposes currently deployed build details and also can be used as a lightweight probe.

The ENV variables are exposed in the Dockerfile, and passed as `build-arg` when building the image in the deploy pipeline.

I've also pointed the k8s probes to the new /health endpoint created in PR #19.

### DIFF
--- a/app/controllers/healthcheck_controller.rb
+++ b/app/controllers/healthcheck_controller.rb
@@ -3,6 +3,10 @@ class HealthcheckController < BareApplicationController
     render json: { healthcheck: healthcheck_result }, status: http_status
   end
 
+  def ping
+    render json: build_args, status: :ok
+  end
+
   private
 
   def healthcheck_result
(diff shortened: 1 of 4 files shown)


app/controllers/healthcheck_controller.rb
```diff
@@ -3,14 +3,18 @@ def show
...
+
   private
 
   def healthcheck_result
     green? ? 'OK' : 'NOT OK'
   end
 
   def http_status
-    green? ? 200 : 503
+    green? ? :ok : :service_unavailable
```
willmcb, line 17: 👍

willmcb (approved):
👍

PR #26, opened 2022-07-19
CRIMAP-16 Update to new ingress controller

zheileman:
Ticket https://dsdmoj.atlassian.net/browse/CRIMAP-16

To improve security of the Cloud Platform they have introduced new v1 Ingress Controllers.

Tweaked the existing ingress manifest as per documentation here:
https://user-guide.cloud-platform.service.justice.gov.uk/documentation/other-topics/Switch-ingress-to-v1-ingress-controller.html#switch-ingress-to-the-new-nginx-ingress-controller-v1-2

And slack convo here:
https://mojdt.slack.com/archives/CH6D099DF/p1657794137979789

This would have incur a minor downtime but because this service is not live/production we can do this all in one go.

To not incur downtime there are specific instructions to do it in 2-steps creating a separate ingress controller.

### DIFF
--- a/config/kubernetes/staging/ingress.yml
+++ b/config/kubernetes/staging/ingress.yml
@@ -14,6 +14,7 @@ metadata:
         return 301 https://raw.githubusercontent.com/ministryofjustice/security-guidance/main/contact/vulnerability-disclosure-security.txt;
       }
 spec:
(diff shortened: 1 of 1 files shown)


willmcb (approved):
💥

PR #27, opened 2022-07-20
CRIMAP-6 Configure Sentry exceptions reporting

zheileman:
## Description of change
Configured Sentry exceptions reporting.
I've created a team `laa-crime-apply` and added the app project to that team in Sentry with name `apply-for-criminal-legal-aid`.

## Link to relevant ticket

https://dsdmoj.atlassian.net/browse/CRIMAP-6

## Notes for reviewer

As part of this I moved the 2 files in `lib/` to `app/lib` as that way Rails will load them automatically and we don't need to use an initialiser to require them.

Not intended to be report local exceptions, but to force to report an exception in local, you can expose the `SENTRY_DSN` ENV variable (found in secrets.yml, have to be base64-decoded), for example adding it to your `.env.development.local` file, and restart the server. Then force an exception, this will get reported to the sentry website.
Do not forget to remove the `SENTRY_DSN` locally!


docker-compose.yml
```diff
@@ -14,6 +14,8 @@ services:
   web:
     build: .
     environment:
+      ENV_NAME: staging # or `production` in the future
```
willmcb, line 17: 👍

willmcb (approved):
LGTM 👍

PR #28, opened 2022-07-21
CRIMAP-6 Add basic error handling and error pages

zheileman:
## Description of change
Handling of a few common errors, like invalid session or `ApplicationNotFound` (this one still not really in use until we have the applications dashboard functionality). No doubt there will be more.

The error pages copy is kind of placeholder, generic copy. This will have to revised at some point in the future.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-6

Example of an error, all are very similar just changing the copy:
Screenshot 2022-07-21 at 11 56 18

You can access each of the error pages by using their URL, for example:

http://localhost:3000/errors/not_found
http://localhost:3000/errors/application_not_found
http://localhost:3000/errors/invalid_session

In production, any URL that doesn't exist will trigger the not found error page. In non-production the exception is shown in the browser on purpose for debug purposes.

### DIFF
--- a/app/controllers/steps/base_step_controller.rb
+++ b/app/controllers/steps/base_step_controller.rb
@@ -1,5 +1,7 @@
 module Steps
   class BaseStepController < ApplicationController
+    before_action :check_crime_application_presence
+
     # rubocop:disable Rails/LexicallyScopedActionFilter
     before_action :update_navigation_stack, only: [:show, :edit]
     # rubocop:enable  Rails/LexicallyScopedActionFilter
--- a/app/views/errors/application_not_found.html.erb
+++ b/app/views/errors/application_not_found.html.erb
@@ -1,0 +1,11 @@
+<% title t('.page_title') %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= t '.heading' %></h1>
+
+    <p class="govuk-body-l"><%=t '.lead_text' %></p>
+
+    <%= link_button t('helpers.start_again'), root_path, class: 'govuk-!-margin-top-2' %>
+  </div>
+</div>
--- a/config/locales/en/errors.yml
+++ b/config/locales/en/errors.yml
@@ -5,6 +5,23 @@ en:
     page_title_prefix: "Error: "
     error_summary:
       heading: There is a problem on this page
+    invalid_session:
+      page_title: Sorry, you’ll have to start again
+      heading: Sorry, you’ll have to start again
+      lead_text: Your session automatically ends if you don’t use the service for XX minutes.
+      more_text: We do this for your security. Any unsaved details will be deleted.
+    application_not_found:
+      page_title: Application not found
+      heading: Sorry, this application no longer exists
+      lead_text: If you copied a web address, please check it’s correct.
+    not_found:
+      page_title: Page not found
+      heading: Page not found
+      lead_text: If you copied a web address, please check it’s correct.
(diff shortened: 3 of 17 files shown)


app/controllers/steps/base_step_controller.rb
```diff
@@ -1,5 +1,7 @@
 module Steps
   class BaseStepController < ApplicationController
+    before_action :check_crime_application_presence
+
     # rubocop:disable Rails/LexicallyScopedActionFilter
```
willmcb, line 5: Do you think it might be worth removing this cop? Or do you think it will come in useful elsewhere?

app/views/errors/application_not_found.html.erb
```diff
@@ -0,0 +1,11 @@
+<% title t('.page_title') %>
+
```
willmcb, line 2: I'm wondering if we could just have a generic view for - "Application not found" "not found" and "unhandled" as the ERB partials appear to be identical. Seems like we have two types of view error partial:

- concise
- verbose (i.e. has the `.more_text`) `<p>` tag

🤔  

I guess the benefit of having them separate is that the i18n works, but we could compose them down into either concise or verbose...

Just a thought its not much code to duplicate really.

willmcb (approved):
💥 looks good - made a few non-blocking comments - let me know if you want to chat about them 👍

config/locales/en/errors.yml
```diff
@@ -5,6 +5,23 @@ en:
     page_title_prefix: "Error: "
     error_summary:
       heading: There is a problem on this page
+    invalid_session:
+      page_title: Sorry, you’ll have to start again
+      heading: Sorry, you’ll have to start again
+      lead_text: Your session automatically ends if you don’t use the service for XX minutes.
```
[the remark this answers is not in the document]
zheileman: Yep exactly. No need having that now, but we can safely assume we would have like 60 mins, maybe 30 mins timeout etc. and will be configurable.

app/controllers/steps/base_step_controller.rb
zheileman: Oh, that cop is for the other `before_action` below that one, because rubocop is complaining that the actions :edit and :show are no declared in that controller (as this controller is suppose to be used in subclasses using inheritance).

I think ideally we might not have that many instances where this happen, so that's why I didn't disable this cop just yet unless it becomes too annoying. Another alternative is to declare the actions in this controller too, just empty or raise exception, mark as nocov and then implement in subclasses 🤔

app/views/errors/application_not_found.html.erb
zheileman: Yes, having separate views helps a lot with the localisation. Now they look similar, but it may be they require further customisation down the line, so didn't want to anticipate the future, for now is just easier having 1 view per error.

But definitely if we end up creating many more error pages, all almost identical, we can extract some bits DRY some bits 👍

app/controllers/steps/base_step_controller.rb
willmcb: That might be clearer to have them declared there and it gives people a sign post of where they are perhaps 🤔

app/views/errors/application_not_found.html.erb
willmcb: Makes sense! 👍

app/controllers/steps/base_step_controller.rb
zheileman: Yep, I will change that in my next PR so it is more clear 👍

PR #30, opened 2022-07-27
CRIMAP-55 Implement multi param dates (client DoB)

zheileman:
## Description of change
Thorough the application journey we are going to need date components in more than one step. This PR implements the handling of this special components, as these are multi param dates, divided in 3 parameters for day, month and year, and have a more complex than usual parsing/casting and validation.

I've split the work into a few logical commits to be easier to follow, with extra information in each of their commit messages.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-55
The date of birth is added to an existing step, client details, refer to this other ticket:
https://dsdmoj.atlassian.net/browse/CRIMAP-43

## Notes for reviewer
There is a lot of guidance about date components and validations in the govuk frontend documentation here:

https://design-system.service.gov.uk/components/date-input/

However please note that with the current DfE form builder it is not possible to set the error just in the field that produced it (for example, the day), it has to be in the whole component. But other than that the rest is as per guidelines, with perhaps error copy to be reviewed by content designer.

There are many different errors (refer to the errors.yml locales) but here I added a few screenshot examples:

Screenshot 2022-07-27 at 11 23 22
Screenshot 2022-07-27 at 11 23 31
Screenshot 2022-07-27 at 11 24 00


app/validators/multiparam_date_validator.rb
```diff
@@ -0,0 +1,73 @@
+class MultiparamDateValidator < ActiveModel::EachValidator
+  DATE_STRUCT = Struct.new('DateStruct', :day, :month, :year, keyword_init: true)
+
+  DEFAULT_OPTIONS = {
+    earliest_year: 1920,
```
willmcb, line 5: Is there anyone over 102 that might want legal aid? Pretty unlikely but you never know? perhaps 1900 might be better?

willmcb (approved):
LGTM 👍

app/validators/multiparam_date_validator.rb
zheileman: Haha! I considered between the MVP development time and the moment we go live with a broader group of providers (not just closed beta with a handful of them) that hypothetical person might be 104 years old by then.

I can change to 1900 to be safe, 122 years old people have all the right to apply 😄

willmcb: They do indeed 👍

PR #31, opened 2022-07-27
CRIMAP-44 Enter client's NINO step

willmcb:
## Description of change
Add form to collect client Nino details + specs / validations

atm validations for NINO format are in the form, but these could be moved to a custom validator at some future point if needed.

Also added some holding pages for the progression of the form steps (`nino_yes.html.erb` and `nino_no.html.erb`) that can be removed as we build out more steps.

## Link to relevant ticket

[CRIMAP](https://dsdmoj.atlassian.net/browse/CRIMAP-44)

## Notes for reviewer
- please check that our 'conventions' are followed 👍  
  
Screenshot 2022-07-27 at 17 35 33

Format Error:
Screenshot 2022-07-27 at 17 36 02

Blank error:
Screenshot 2022-07-27 at 17 37 16

Run through the stepped journey form  `/has_partner`:
- `/details` on completion now routes to `has_nino`
- check validations on `/has_nino`:
  - if radios are blank it errors saying choose one 
  - if yes selected and NINO is blank or incorrect then it will error saying that you need to enter a NINO in the right format (`AB123456C` is a valid NINO)

### DIFF
--- a/app/controllers/steps/client/has_nino_controller.rb
+++ b/app/controllers/steps/client/has_nino_controller.rb
@@ -10,12 +10,6 @@ module Steps
       def update
         update_and_advance(HasNinoForm, as: :has_nino)
       end
-
-      private
-
-      def decision_tree_class
-        Decisions::ClientDecisionTree
-      end
     end
   end
 end
--- a/app/controllers/steps/client/has_nino_controller.rb
+++ b/app/controllers/steps/client/has_nino_controller.rb
@@ -1,0 +1,21 @@
+module Steps
+  module Client
+    class HasNinoController < Steps::ClientStepController
+      def edit
+        @form_object = HasNinoForm.build(
+          current_crime_application
+        )
+      end
+
+      def update
+        update_and_advance(HasNinoForm, as: :has_nino)
+      end
+
+      private
+
+      def decision_tree_class
+        Decisions::ClientDecisionTree
+      end
+    end
+  end
+end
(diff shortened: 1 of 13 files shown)


app/controllers/steps/client/has_nino_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+        )
+      end
+
+      def update
+        update_and_advance(HasNinoForm, as: :has_nino)
+      end
+
+      private
+
+      def decision_tree_class
```
zheileman, line 16: Although completely harmless, there is no need to declare the decision tree in this controller, as it will be using the one implemented in the superclass `Steps::ClientStepController` 👍

config/locales/en/steps.yml
```diff
@@ -10,3 +10,7 @@ en:
           page_title: Enter your client’s details
           heading: Enter your client’s details
           full_name_legend: Full name
+      has_nino:
+        edit:
+          page_title: Enter your client’s NINO
+          has_nino: Do you have your client's National Insurance number?
```
zheileman, line 16: As we are using the legend as the page heading, this `has_nino` locale in reality is not being used so can be removed.

zheileman (commented):
🥳 🥇 

A couple harmless suggestions.

Also if you want, to further validate the form object, you can use the shared scenarios (similar to the `details_form_spec`):

`it_behaves_like 'a has-one-association form'`

app/controllers/steps/client/has_nino_controller.rb
willmcb: Thanks don't know why I didn't spot that! :)

zheileman (approved):

PR #32, opened 2022-07-28
CRIMAP-56 Custom string type to strip blanks

zheileman:
## Description of change
A common thing seen across web applications is copy&paste data, sometimes from a spreadsheet etc. where inadvertently they introduce some extraneous spacing.

These leading/trailing spaces are kept intact on saving to the database, which then can produce some validation errors in downstream services.

To reduce the number of instances where this happens, we can create a custom `:string` type that literally behave like the usual type but performs a strip if the value is a string.

Note: as part of these I moved the types we've defined so far to the `Type` namespace.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-56


app/attributes/type/stripped_string.rb
```diff
@@ -0,0 +1,10 @@
+module Type
+  class StrippedString < ActiveModel::Type::String
+    # Remove leading and trailing spaces from strings, as these
+    # are usually obtained from input text and users can copy&paste
+    # introducing extraneous spaces without noticing.
+    def cast(value)
+      value.is_a?(String) ? super(value.strip) : super
```
willmcb, line 7: 💥

spec/attributes/type/stripped_string_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    describe 'intermediate spaces' do
+      let(:value) { ' foo bar ' }
+      it { expect(coerced_value).to eq('foo bar') }
+    end
+
+    describe 'more than one intermediate spaces' do
+      let(:value) { ' foo   bar ' }
+      it { expect(coerced_value).to eq('foo   bar') }
+    end
```
willmcb, line 54: Small thing but I guess we could test 'foobar   ' and '   foobar' 🤔

spec/attributes/type/stripped_string_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  describe 'when value is numeric' do
+    let(:value) { 123 }
+    it { expect(coerced_value).to eq('123') }
+  end
+
+  describe 'when value is true' do
+    let(:value) { true }
+    it { expect(coerced_value).to eq('t') }
```
willmcb, line 64: Why coerce it to 't' / 'f' rather than 'true' / 'false'?

spec/attributes/type/yes_no_type_spec.rb
```diff
@@ -1,10 +1,22 @@
 require 'rails_helper'
 
-RSpec.describe YesNoType do
+RSpec.describe Type::YesNoType do
   subject { described_class.new }
 
   let(:coerced_value) { subject.cast(value) }
 
+  describe 'registry' do
```
willmcb, line 8: 👍

willmcb (approved):
LGTM 👍 

Few minor non-blocking comments :)

spec/attributes/type/stripped_string_spec.rb
zheileman: Is how the original `ActiveModel::Type::String` behaves, as I don't really want to replace it, just slightly modify it to strip blanks from strings, I would prefer to keep it working as intended (thus why I `super` instead of just returning `value`)

spec/attributes/type/stripped_string_spec.rb
zheileman: Yep, as I know internally it uses `#strip` method, it will do lead and trail, but you're right an external observer might not know the specific implementation so a explicit expectation for only lead blank / trail blanks will not hurt, will add it 👍

PR #33, opened 2022-07-28
CRIMAP-10 Add developer tools to footer

zheileman:
## Description of change
Just some basic operations for now. It is behind a feature flag, enabled in development/staging envs, disabled on production (although we don't have yet prod).

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-10

## Notes for reviewer
As we develop more of the app, we might have to change some of the links/actions but that is OK, this is supposed to help us get quicker to some steps / pre-fill something etc. For now we don't have a lot anyways.

I tried to use a request spec, instead of controller spec, and couldn't make it work to check the `reset_session` was being called as apparently request specs don't have access to the session. So instead of having one test (the redirect) in the request spec, and another test (the reset session) in the controller spec, I opted for having all in the controller spec...

Screenshot 2022-07-28 at 16 12 34

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,0 +1,42 @@
+class SessionsController < ApplicationController
+  def destroy
+    reset_session
+    redirect_to root_path
+  end
+
+  # :nocov:
+  def bypass_to_client_details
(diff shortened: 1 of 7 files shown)


willmcb (approved):
Nice work! :)

PR #35, opened 2022-08-01
CRIMAP-60 Add DB fields to hold contact details

zheileman:
## Description of change
We are working on the next step of the journey, to ask for home address, phone number and correspondence address. These can all be considered as "contact details" for sake of simplicity, as eventually this also will include a phone number, and stored in a separate table.

This table for now only includes the home address and the correspondence address. These are JSON attributes in the database but exposed as individual attribute getter/setter with `store_accessor`.

## Link to relevant ticket


zheileman:
Closing. A new PR for a different approach will be raised soon.

PR #36, opened 2022-08-02
Refactor people relationships and prepare contact details

zheileman:
## Description of change
To anticipate for when we have `partner`, it makes more sense to have a `Person` superclass with 2 (for now and probably not to change ever) subclasses `Applicant` and `Partner`.

Also to prepare for the contact details, it makes more sense for these to be a `has_one` relationship on each person. However trying to simplify second level relationships, we expose 2 `has_one through` relationships on the `crime_application` so it keeps behaving kind of like a 1st level relation and our form object are kept simple.

## Link to relevant ticket
This work is partially motivated by the following story:

Everything should continue working but any locally created applicant details will be purged from DB.

### DIFF
--- a/db/migrate/20220801144607_create_contact_details_table.rb
+++ b/db/migrate/20220801144607_create_contact_details_table.rb
@@ -1,0 +1,12 @@
+class CreateContactDetailsTable < ActiveRecord::Migration[7.0]
+  def change
+    create_table :contact_details, id: :uuid do |t|
+      t.timestamps
+
+      t.references :person, type: :uuid, foreign_key: true, null: false, index: { unique: true }
+
+      t.jsonb :home_address, default: {}
+      t.jsonb :correspondence_address, default: {}
+    end
+  end
+end
--- a/spec/forms/steps/client/has_nino_form_spec.rb
+++ b/spec/forms/steps/client/has_nino_form_spec.rb
@@ -11,12 +11,9 @@ RSpec.describe Steps::Client::HasNinoForm do
   } }
 
   let(:crime_application) { 
-    instance_double(CrimeApplication, 
-      applicant_details: applicant_details
-    )
+    instance_double(CrimeApplication)
   }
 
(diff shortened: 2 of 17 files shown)


db/migrate/20220801144607_create_contact_details_table.rb
```diff
@@ -0,0 +1,12 @@
+class CreateContactDetailsTable < ActiveRecord::Migration[7.0]
+  def change
+    create_table :contact_details, id: :uuid do |t|
+      t.timestamps
+
+      t.references :person, type: :uuid, foreign_key: true, null: false, index: { unique: true }
+
+      t.json :home_address, default: {}
```
willmcb, line 8: Have you thought about jsonb for this I think we should be able to run faster / more meaningful queries on it if need be?

https://blog.saeloun.com/2022/01/04/postgresql-data-types-in-rails.html#:~:text=Though%20both%20JSON%2FJSONB%20look,querying%20more%20efficient%20and%20faster.

spec/forms/steps/client/has_nino_form_spec.rb
```diff
@@ -55,20 +52,6 @@
       end
     end
 
-    context 'when `has_nino` is valid' do
```
zheileman, line 58: @willmcb I removed this, as in reality is the same test the shared examples perform below in the 
```it_behaves_like 'a has-one-association form'``` as it is the happy path.

willmcb (approved):
Nice :) One small, non blocking, comment on the JSON column type 👍

db/migrate/20220801144607_create_contact_details_table.rb
zheileman: I've used `jsonb` in the past for audit tables where queries are performed, but for this I didn't think we would need to query the data (as in doing selects, etc.), but as I don't think there is a negative impact to use `jsonb` even if we don't fully use the capabilities, I can change it 👍

PR #38, opened 2022-08-03
CRIMAP-60 Add home address manual entry

zheileman:
## Description of change
First part of the journey to enter the home address of the applicant (client's home address).

For now just the manual entry. Postcode lookup will be added later on as a separate one or more PRs.

The code is somehow coupled for now to the applicant as we don't have `partner` yet for the MVP, but in the future I anticipate we could make these controllers / form objects somehow work for partner too with not too much effort.

The database models already supports as of PR #36 to have different contact details for applicant and for partner.

## Link to relevant ticket

## Notes for reviewer
Manual entry only. After entering details it goes back to home (as we don't have yet following step).

Screenshot 2022-08-03 at 10 40 51

Screenshot 2022-08-03 at 10 41 15

### DIFF
--- a/app/controllers/steps/contact_step_controller.rb
+++ b/app/controllers/steps/contact_step_controller.rb
@@ -1,0 +1,9 @@
+module Steps
+  class ContactStepController < BaseStepController
+    private
+
+    def decision_tree_class
+      Decisions::ContactDecisionTree
+    end
+  end
+end
--- a/app/services/decisions/contact_decision_tree.rb
+++ b/app/services/decisions/contact_decision_tree.rb
@@ -1,0 +1,12 @@
+module Decisions
+  class ContactDecisionTree < BaseDecisionTree
+    def destination
+      case step_name
+      when :home_address
+        show('/home', action: :index)
+      else
+        raise InvalidStep, "Invalid step '#{step_name}'"
+      end
+    end
+  end
+end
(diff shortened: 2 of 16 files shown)


app/controllers/steps/contact_step_controller.rb
```diff
@@ -0,0 +1,9 @@
+module Steps
+  class ContactStepController < BaseStepController
+    private
+
+    def decision_tree_class
+      Decisions::ContactDecisionTree
```
willmcb, line 6: 👍

app/services/decisions/contact_decision_tree.rb
```diff
@@ -0,0 +1,12 @@
+module Decisions
+  class ContactDecisionTree < BaseDecisionTree
+    def destination
+      case step_name
+      when :home_details
```
willmcb: wondering if "home_address_details" would be less ambiguous? 🤔

willmcb (approved):
LGTM 👍

app/services/decisions/contact_decision_tree.rb
zheileman: Yep, I changed it to `home_address` as it also is the name of the attribute where we are storing the address details, in the `contact_details` table.
`home_details` was a leftover from a previous incarnation of the code 🤣

PR #40, opened 2022-08-03
Crimap 45 eforms redirect

willmcb:
## Description of change
Adds a partial for the eForms redirect if NINO is not given.

As we are no longer checking for partner, we agreed that this can be NINO specific for now. Hence the `NinoExitController` etc

## Notes for reviewer
- Please check copy looks okay
- Note: changes to NINO page where the question is replaced by single input with link for what is now "No" option will be added in different PR.
- Also "Back to you applications" just goes to '#' at the moment pending that page actually being created

[CRIMAP-45](https://dsdmoj.atlassian.net/jira/software/projects/CRIMAP/boards/897?selectedIssue=CRIMAP-45)

### After changes:
Screenshot 2022-08-03 at 16 05 23

Cycle through application until you reach NINO page > click no > click "Save and Continue"

This should redirect to the new eForms partial as per the prototype


app/assets/stylesheets/local/custom.scss
```diff
@@ -1,2 +1,69 @@
 // App-specific custom styles and overrides
 //
+
```
zheileman, line 3: Maybe we can move all this to their own file inside local/ and then reference the import the file in the application.scss, as it all relate to the same thing (flash card).

I've not tested it as I'm in the middle of something else but a bit worried there might be something else to consider here (apart from ideally not hardcoding colours, paddings, margins, etc. and use the GOV.UK values?).

The one we did for C100 (inspired by BEIS) can serve as an example (but we probably don't need everything in there as we don't have chevron buttons, etc.)

https://github.com/ministryofjustice/c100-application/blob/master/app/assets/stylesheets/local/intro_card.scss

app/views/steps/client/nino_exit/show.html.erb
```diff
@@ -0,0 +1,29 @@
...
+<% step_header %>
+
+<div class="flash-card">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+      <header>
+        <h1 class="govuk-heading-l"><%= t('.heading') %></h1>
+      </header>
+
+      <p>You cannot use this service if your client:</p>
```
zheileman, line 11: On thing I usually do with heavy-copy views (and in particular these kind of show views without any particular logic or complexity) is to use translated views (as in `show.en.html.erb`) and use hardcoded copy (no locales, apart from perhaps just the `page_title`).

It makes these views easier to create, to read and to modify. If we ever have another language, we make a copy of the view, with a new locale (i.e. show.cy.html.erb) and just translate the copy on it.

Just a suggestion! 😄

app/views/steps/client/nino_exit/show.html.erb
```diff
@@ -0,0 +1,29 @@
...
+
+      <ul class="govuk-list govuk-list--bullet">
+        <li>has a partner</li>
+        <li>does not have a National Insurance number </li>
+        <li>does not receive a passporting benefit based on DWP records, including Universal Credits</li>
+        <li>is employed, including self-employed, business partner, company director or shareholder</li>
+      </ul>
+
+      <div>
+        <%= link_to t('.eforms_apply'), t('.eforms_url'), class: 'govuk-button' %>
```
zheileman, line 21: There is a `#link_button` helper to makes links look like buttons that might be useful here.

spec/requests/eforms_redirect_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    it "returns the correct with success" do
+      expect(response).to have_http_status(:success)
+    end
+
+    it "returns the correct redirect link to eForms" do
+      expect(response.body).to include("https://portal.legalservices.gov.uk")
+    end
+
+    it "returns the correct copy" do
```
zheileman, line 21: Unsure about testing copy line for line 🤔

zheileman (approved):
LGTM 👍 

Disclaimer: I've not actually tested it as I can't at the moment, I added a few comments with suggestions!

PR #42, opened 2022-08-04
CRIMAP-61 Refactor addresses and add lookup step

zheileman:
## Description of change
This is part of the work required to introduce the postcode lookup but also involve a refactoring and reorganisation of what little we had up until now for addresses, to cope with the fact we will have 2 kind of addresses (home and correspondence) for each of the people in an application (for now only applicant but soon also partner, and maybe provider too).

I've not fully removed the old address work, or the contact_details DB table, as to not make this PR too large. I will create a separate PR after this is merged to clean up the old code.

In essence now we have a new `Addresses` table, and each person has_many addresses, identified by its type (HomeAddress or CorrespondenceAddress).

The Address record has the attributes as plain strings, no longer jsonb as that is not really necessary.
Also, for simplicity, all the addresses can be queried from the crime application model as in `crime_application.addresses`, which is used in the address controllers to ensure a given UUID truly belongs to the current application in progress.

Note: as the controllers behave slightly different to other controllers, as these require the UUID of the address in question, I've created new shared examples for these address controllers.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-61

## Notes for reviewer
Note: no "postcode" validation yet, we can add that later, for now only presence.
No API lookup yet, in a follow-up PR integrate with the Ordnance API.

Screenshot 2022-08-04 at 13 02 13
Screenshot 2022-08-04 at 13 02 48
Screenshot 2022-08-04 at 13 03 17


arthurashman (approved):
Thanks for talking me through this. I have had a play and all works as expected. Great refactor, LGTM

PR #43, opened 2022-08-04
Add eForms redirect when there is a partner

willmcb:
## Description of change
Adds eFroms redirect to `/has_partner` page if applicant has a partner (selects 'yes')
## Notes for reviewer
- I haven't added to the request spec for the eforms redirect as I think it is covered by the nino_exit path.
- Also removed some the temp `selected_yes/no` routes views for has_partner and has_nino as these now have places to go as extra steps have been added.

- start application > at has partner? select 'Yes' > submit
- should go to eForms redirect page.


zheileman (approved):
🥇 Great!

PR #47, opened 2022-08-05
CRIMAP-87 Implement address lookup service

zheileman:
## Description of change
Implement the backend side of the address lookup functionality. It is not yet wired to the frontend but can be manually tested by using the console.

This is heavily inspired by the work done on Civil Apply and C100 (Child arrangements) services, maybe others. I'm keeping a few improvements we made for the C100 service too.

Individual commits with more context.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-87

As there is no interface yet for this, it can be tested as follow (checkout the branch and run `bundle` first):

1. Ask a colleague for an Ordnance Survey API key (or decode the one in staging/secrets.yml)
2. Add the key to your `.env.development.local` file as: `ORDNANCE_SURVEY_API_KEY=foobar`
3. Open a rails console
4. Instantiate an address lookup service with a postcode and then retrieve the results as follows:

```ruby
service = Ordnance::AddressLookup.new('SW1A 1AA')
results = service.call
=> 
[#<struct Ordnance::AddressLookupResults::Address         
  address_line_one="BUCKINGHAM PALACE",                   
  address_line_two="",                                    
  city="LONDON",                                          
  country="UNITED KINGDOM",                               
  postcode="SW1A 1AA">]  
```


.rubocop.yml
```diff
@@ -44,6 +44,10 @@ Layout/HashAlignment:
 # Enabled but tweaked cops
 ##########################
 
+# To be tweaked until we find the right balance
+Metrics/MethodLength:
+  Max: 12
+
```
arthurashman, line 50: If the decision trees will continue to grow, and the Max here grow with it, MethodLength will become useless at keeping other methods in check for us - What do you think to excluding the decisions folder from MethodLength?

zheileman: Yes, the idea is to get to some sensible max, I don't know if it will be 12, or 14, but then anything above will have to be either refactored to make it smaller or when it doesn't make sense as it makes it harder to follow or read the code we add the `disable` annotation like probably in decision trees.

I would say, around 14 maybe 15 lines limit is still pretty manageable. Rubocop by default wants 10 which in my opinion is too low and tends to have the perverse side effect of creating too many small methods that harms readability 👍 
But happy to have something lower!

arthurashman: No no, that sounds sensible. 

I thought you meant to increment this number every time the decision tree destination method got longer. All good 👍

arthurashman (approved):

arthurashman:
LGTM, works on the console for me

PR #48, opened 2022-08-08
CRIMAP-80 update client nino page

willmcb:
## Description of change
Slight rework of nino page as per updated design

## Link to relevant ticket
[CRIMAP-80](https://dsdmoj.atlassian.net/browse/CRIMAP-80)

### Before changes:
Screenshot 2022-08-08 at 11 44 33

### After changes:
Screenshot 2022-08-08 at 11 44 03

Check that NINO validation works the same as before.

Check that link to eForms redirect works.

### DIFF
--- a/app/services/decisions/client_decision_tree.rb
+++ b/app/services/decisions/client_decision_tree.rb
@@ -8,7 +8,7 @@ module Decisions
       when :details
         edit(:has_nino)
       when :has_nino
-        after_has_nino
+        edit('/steps/contact/home_address')
       when :contact_details
         show('/home', action: :index)
       else
@@ -24,14 +24,6 @@ module Decisions
         show(:partner_exit)
       else
         edit(:details)
-      end
-    end
-
-    def after_has_nino
-      if form_object.has_nino.yes?
-        start_address_journey(HomeAddress, form_object.applicant)
-      else
-        show(:nino_exit)
       end
     end
 
--- a/app/views/steps/client/has_nino/edit.html.erb
+++ b/app/views/steps/client/has_nino/edit.html.erb
@@ -5,16 +5,18 @@
   <div class="govuk-grid-column-two-thirds">
     <%= govuk_error_summary(@form_object) %>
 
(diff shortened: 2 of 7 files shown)


config/locales/en/steps.yml
```diff
@@ -13,6 +13,8 @@ en:
       has_nino:
         edit:
           page_title: Enter your client’s NINO
+          heading: Do you have your client's National Insurance number?
```
arthurashman, line 16: Needs to be: 'Enter your client's National Insurance number'
https://www.figma.com/file/thkvnDBbQbHlSqzt1TE3lk/Crime-Apply?node-id=2769%3A25139

app/views/steps/client/has_nino/edit.html.erb
```diff
@@ -5,16 +5,18 @@
...
+
     <%= step_form @form_object do |f| %>
-      <%= f.govuk_radio_buttons_fieldset :has_nino do %>
-        <%= f.govuk_radio_button :has_nino, YesNoAnswer::YES, link_errors: true do %>
-          <%= f.govuk_text_field :nino, width: 'three-quarters' %>
-        <% end %>
 
-        <%= f.govuk_radio_button :has_nino, YesNoAnswer::NO %>
-      <% end %>
+      <%= f.govuk_text_field :nino, width: 'three-quarters', label: nil %>
```
zheileman, line 12: I think for accessibility reasons, the label must be always set, however hidden (`label: { hidden: true }`).
The label can be just "Enter client's national insurance number" (comes from locales), anything representative of what the input is for as it is read aloud on focus by screen readers.

Also minor thing but feels like `width: 'three-quarters'` is too big for NINO? maybe `one-third`?

app/services/decisions/client_decision_tree.rb
```diff
@@ -27,14 +27,6 @@ def after_has_partner
       end
     end
 
-    def after_has_nino
```
zheileman, line 30: I think this is a bad rebase 😬

PR #49, opened 2022-08-08
CRIMAP-80 Implement Nino form redesign

willmcb:
## Description of change
Slight rework of nino page as per updated design

## Link to relevant ticket
[CRIMAP-80](https://dsdmoj.atlassian.net/browse/CRIMAP-80)

### Before changes:
Screenshot 2022-08-08 at 11 44 33

### After changes:
Screenshot 2022-08-08 at 12 44 34

Check that NINO validation works the same as before.

Check that link to eForms redirect works.

### DIFF
--- a/config/locales/en/steps.yml
+++ b/config/locales/en/steps.yml
@@ -12,7 +12,9 @@ en:
           full_name_legend: Full name
       has_nino:
         edit:
+          heading: Enter your client's National Insurance number 
           page_title: Enter your client’s NINO
+          eforms_link: My client does not have a National Insurance number or can't find it
       nino_exit:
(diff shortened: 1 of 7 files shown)


zheileman (approved):
🥳 LGTM

config/locales/en/steps.yml
```diff
@@ -12,7 +12,9 @@ en:
           full_name_legend: Full name
       has_nino:
```
arthurashman, line 13: Just wondering if it's worth renaming the namespace and page 'nino' instead of 'has_nino' - not a biggie if neither for you think it's worth bothering with

PR #50, opened 2022-08-08
Improve nino validation

arthurashman:
## Description of change

Based on NINO allocation rules as per https://www.gov.uk/hmrc-internal-manuals/national-insurance-manual/nim39110 - improved validation to include raising validation errors for specific unused prefixes

Also strips spaces and upcases input prior to validation to allow for input variation (particularly for spaces which are shown as part of the hint text NINO example)

### DIFF
--- a/spec/forms/steps/client/has_nino_form_spec.rb
+++ b/spec/forms/steps/client/has_nino_form_spec.rb
@@ -28,11 +28,39 @@ RSpec.describe Steps::Client::HasNinoForm do
     end
 
     context 'when `nino` is invalid' do
-      let(:nino) { 'not a NINO' }
+      context 'with a random string' do
+        let(:nino) { 'not a NINO' }
 
-      it 'has a validation error on the field' do
-        expect(subject).to_not be_valid
-        expect(subject.errors.of_kind?(:nino, :invalid)).to eq(true)
+        it 'has a validation error on the field' do
+          expect(subject).to_not be_valid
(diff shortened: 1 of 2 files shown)


spec/forms/steps/client/has_nino_form_spec.rb
```diff
@@ -28,11 +28,39 @@
...
+      context 'with spaces between numbers' do
+        let(:nino) { 'AB 12 34 56 C' }
+        it 'passes validation' do
+          expect(subject).to be_valid
+          expect(subject.errors.of_kind?(:nino, :invalid)).to eq(false)
+        end
+      end
+      context 'with trailing spaces' do
+        let(:nino) { ' AB 1234 56C ' }
+        it 'passed validation' do
```
willmcb, line 60: Could be worth checking the value when you call `nino` as well? Does it return the original for or one that has its white space stripped out?

Something along the lines of:

```ruby
      context 'with spaces between numbers' do
        let(:nino) { 'AB 12 34 56 C' }
        it 'passes validation' do
          expect(subject.nino).to be('AB123456C')
        end
      end
```

willmcb (approved):
LGTM 👍 

- Left a comment on possibly adding another test

spec/forms/steps/client/has_nino_form_spec.rb
arthurashman: Good idea, have done

PR #53, opened 2022-08-09
Add correspondence address type + amend view + specs

willmcb:
## Description of change

Adds ability for a provider to choose a correspondence type for the address their applicant can be contacted on.

## Link to relevant ticket
[CRIMAP-58](https://dsdmoj.atlassian.net/browse/CRIMAP-58)

## Notes for reviewer
- this PR does not change the client decision tree. Changes to this will be added in a future PR.

### Before changes:
Screenshot 2022-08-09 at 16 30 13

### After changes:
Screenshot 2022-08-09 at 16 31 23

- Go through journey past the address entry pages to the contact details page
- Test validations by not selecting a correspondence address type > error should appear
- Add a phone number and select an address and the page should progress to the holding page, pending decision tree updates in future PR


app/forms/steps/client/contact_details_form.rb
```diff
@@ -8,14 +8,26 @@ class ContactDetailsForm < Steps::BaseFormObject
 
       attribute :telephone_number, :string
 
+      attribute :correspondence_address_type, :correspondence_address_type
```
zheileman, line 11: As mentioned, we can use just a `:string` and not create a custom type just for this. There is no real benefit as we are not doing transformations or anything like that 👍

app/views/steps/client/contact_details/edit.html.erb
```diff
@@ -10,6 +10,11 @@
     <%= step_form @form_object do |f| %>
       <%= f.govuk_phone_field :telephone_number, autocomplete: 'off', width: 'one-third', label: { size: 'm' } %>
 
+      
+      <%= f.govuk_collection_radio_buttons :correspondence_address_type, 
+                                            @form_object.choices,
+                                            :value %>
```
zheileman, line 16: I think we might want to customise the legend as it seems too big (and tag is h1):
`legend: { tag: 'span', size: 'm' }`

In reality, I think we have to take a look at the defaults set in `config/initializers/form_builder.rb` because it feels like those defaults are not working for us in this app. We can change them so the amount of times we have to tweak tags/size on some inputs is reduced.

zheileman (changes requested):
Cool 🥳 I added a few comments, but nothing important really!

app/forms/steps/client/contact_details_form.rb
```diff
@@ -8,14 +8,26 @@ class ContactDetailsForm < Steps::BaseFormObject
...
+
+      validates :correspondence_address_type,
+                inclusion: { in: :choices },
+                presence: true
 
       def telephone_number=(str)
         super(str.delete(' ')) if str
       end
 
+      def choices
```
zheileman, line 27: Oh I just recalled, the choices are dynamic, as in, if the provider didn't enter a home address, then the Home address choice will not show.
But that can be done in a separate small PR if you prefer 👍

app/views/steps/client/contact_details/edit.html.erb
willmcb: Shall I leave as is then and we can put up a separate PR with the correct defaults?

app/forms/steps/client/contact_details_form.rb
willmcb: I'll raise in a separate PR 👍

app/views/steps/client/contact_details/edit.html.erb
zheileman: Yep, lets do it separately as changing the defaults will mean we have to go through all the existing form pages and ensure they are good or make tweaks.

I think the defaults are perfect for when you have 1 question per page (usually radios) as it defaults the radios legend to be the h1 and heading of the page.
That's not the case the moment you have more than 1 element per page.

app/forms/steps/client/contact_details_form.rb
```diff
@@ -30,6 +30,10 @@ def choices
 
       private
 
+      def string_choices
+        choices.map { |choice| choice.value.to_s }
```
zheileman, line 34: You can do if you prefer `choices.map(&:to_s)` because `to_s` is delegated to `value` in the `ValueObject` superclass 😜 it is just a shortcut anyways.

zheileman (approved):

app/forms/steps/client/contact_details_form.rb
willmcb: Nice I'll do that - didn't think of the delegation!

PR #54, opened 2022-08-10
Add UK postcode validator

zheileman:
## Description of change
Improve the postcode lookup step by ensuring valid UK postcodes are used, as otherwise Ordnance Survey will fail or return no addresses anyways.

It validates against million of postcodes in England, Wales, Scotland, Northern Ireland, the Channel Islands, and the Isle of Man.

I've also checked and it pass for Guernsey and Jersey postcodes too.

Not really related to this validator but I've renamed the Struct method from `address_lines` to `compact_address` to better capture the meaning, and added the city to it because at first glance in the dropdown a provider might want to have the assurance they are looking at results from a city they might expect, particularly important if we are going to skip the address manual entry page after they select an address in the dropdown.

Screenshot 2022-08-10 at 10 38 06

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ gem 'faraday', '~> 2.4'
 gem 'pg', '~> 1.4'
 gem 'puma'
 gem 'rails', '~> 7.0.3'
+gem 'uk_postcode'
 
 gem 'govuk_design_system_formbuilder', '~> 3.1.0'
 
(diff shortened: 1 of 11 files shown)


willmcb (approved):
Nice, this gem seems really useful 👍

zheileman:
> Nice, this gem seems really useful 👍

Yep I've used it in the past for the C100 and works like a charm.

PR #55, opened 2022-08-10
deps: Bump to latest GOV.UK Frontend version

zheileman:
## Description of change
No breaking changes or new features really, but good to be always in the latest version as they fix things behind the scenes.

Changelog:
https://github.com/alphagov/govuk-frontend/releases/tag/v4.3.0

## Notes for reviewer
After checkout, just in case run stop the server, run `rails assets:clobber` and then run `bin/dev` again. Don't think is purely necessary but might avoid any issues.

### DIFF
--- a/config/importmap.rb
+++ b/config/importmap.rb
@@ -1,4 +1,4 @@
 # Pin npm packages by running ./bin/importmap
 
 pin "application", preload: true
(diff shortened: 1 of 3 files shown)


willmcb (approved):
:shipit:

PR #57, opened 2022-08-10
Sort out form builder defaults and fix markup

zheileman:
## Description of change
We had non-standard defaults set for the form builder. These are good defaults when an app is mainly done by many 1 question per page, as it saves having to set again and again the same tag, size, etc.

But in reality this app will not have that many individual questions and where we have them, we can customise accordingly.

Removed the custom defaults and tweaked the existing pages where needed to output proper markup, tags and sizes.

Also added `autocomplete: 'off'` if it was missing.

## Notes for reviewer
Somehow some of the previous PR didn't commit the last version of the schema.rb, I've added it here otherwise migrations can fail.

Visually there is not that much difference, and everything should look pretty much the same.

### DIFF
--- a/app/views/steps/address/lookup/edit.html.erb
+++ b/app/views/steps/address/lookup/edit.html.erb
@@ -5,10 +5,8 @@
   <div class="govuk-grid-column-two-thirds">
     <%= govuk_error_summary(@form_object) %>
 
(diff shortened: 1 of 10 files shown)


willmcb (approved):
LGTM 👍

PR #58, opened 2022-08-11
Implement inquiry mechanism for value objects

zheileman:
## Description of change
I've probably spent more time that I should on this little QoL thing, but in the end it is kind of nice :-)

Instead of implementing again and again inquiry methods for each of the values on a value-object (although in many cases it is not necessary if they are not inspected later on), now the  superclass `ValueObject` takes care of defining this "question mark" methods to ask a value object about its values.

It iterates through all defined values in the `VALUES` constant declaring methods like `yes?` or `no?` or `home_address?` etc.

Also moved to the superclass the class method `.values`. This makes for leaner value objects, to the point many will only have the `VALUES` constant and nothing else.

Any more complicated inquiry logic can be defined as separated methods in each value object like before.
Tests are also leaner as no need to test in every instance the inquiry methods, the "definition" of these methods is tested in the superclass.

In a console you can do the following (considering the existing value-object we already have):

```ruby
c = CorrespondenceTypeAnswer.new('home_address')
c.home_address?
=> true
c.other_address?
=> false
c.foobar?
(irb):6:in `<main>': undefined method `foobar?'
c.methods - Object.methods
=> [:value, :home_address?, :providers_office_address?, :other_address?, :to_sym]
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -61,6 +61,11 @@ Naming/MethodParameterName:
   AllowedNames:
     - as
 
+# Rubocop mistakenly consider `#values` method as being called on a Hash
+Style/HashEachMethods:
+  Exclude:
+    - app/value_objects/value_object.rb
+
 Naming/PredicateName:
   AllowedMethods:
     - has_one_association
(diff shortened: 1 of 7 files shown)


arthurashman (approved):
Very nice!

PR #59, opened 2022-08-11
CRIMAP-101 Make corrs selector dynamic

willmcb:
## Description of change

Dynamically changes the Correspondence Address selection on the contact details page

## Link to relevant ticket
[CRIMAP-101](https://dsdmoj.atlassian.net/browse/CRIMAP-101)

### Before changes:
Screenshot 2022-08-11 at 16 21 33

### After changes:

- when no home address:
Screenshot 2022-08-11 at 15 39 53

Go through the form to the client details page, you should see all three options available to select a Correspondence address:
- Same as home address
- Provider's address
- Somewhere else

Open rails console, run the following:
```
ha = HomeAddress.last
ha.address_line_one = nil
ha.save
```
Reload the page. You should see only the following options now:
- Provider's address
- Somewhere else

Submit the form - validations should work as before.

### DIFF
--- a/spec/forms/steps/client/contact_details_form_spec.rb
+++ b/spec/forms/steps/client/contact_details_form_spec.rb
@@ -11,18 +11,18 @@ RSpec.describe Steps::Client::ContactDetailsForm do
 
   let(:telephone_number) { nil }
   let(:correspondence_address_type) { nil }
+  let(:has_home_address) { true }
 
   subject { described_class.new(arguments) }
 
+  let(:applicant_double) { instance_double(Applicant) }
+
+  before do
+    allow(subject).to receive(:applicant).and_return(applicant_double)
+    allow(applicant_double).to receive(:home_address?).and_return(has_home_address)
+  end
+
   describe '#choices' do
-    # A bit more extensive testing of the chain of calls, so coverage passes
-    let(:applicant_double) { instance_double(Applicant) }
-
-    before do
-      allow(subject).to receive(:applicant).and_return(applicant_double)
-      allow(applicant_double).to receive(:home_address?).and_return(has_home_address)
-    end
-
     context 'when applicant has home address' do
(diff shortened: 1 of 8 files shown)


spec/forms/steps/client/contact_details_form_spec.rb
```diff
@@ -11,10 +11,39 @@
 
   let(:telephone_number) { nil }
   let(:correspondence_address_type) { nil }
+  let(:has_home_address) { true }
```
arthurashman, line 14: Should we set this to nil if the 'when applicant has home address' context explicitly sets it as true

spec/forms/steps/client/contact_details_form_spec.rb
```diff
@@ -36,38 +65,51 @@
...
         expect(subject).to_not be_valid
-        expect(subject.errors.of_kind?(:correspondence_address_type, :blank)).to eq(true)
+        expect(subject.errors.of_kind?(:correspondence_address_type, :inclusion)).to eq(true)
       end
     end
 
     context 'when correspondence_address_type contains an incorrect value' do
       let(:telephone_number) { '07000 000 000' }
-      let(:telephone_number) { 'university_address' }
+      let(:correspondence_address_type) { 'university_address' }
```
arthurashman, line 88: I can't quite tell what's going on, but these are evergreen when asserting against correspondence_address_type

spec/forms/steps/client/contact_details_form_spec.rb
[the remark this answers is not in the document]
arthurashman: Thank you, yep makes sense👌 I just realised that changing it to nil then broke tests later on that I commented on before realising my mistake

arthurashman:
LGTM - all works as expected

arthurashman (approved):

PR #60, opened 2022-08-12
Improve telephone number validation and update error messages

arthurashman:
## Description of change
Updated previous regex that would allow all symbols and numeric input to allow numeric and +()- input only

Updated error messages to align with [design approved messaging ](https://docs.google.com/spreadsheets/d/1F30TsGgtB94e2AHoqo1piKqyF4_1kqbyIdKRtAFJzYQ/edit#gid=0)

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-64

### DIFF
--- a/config/locales/en/errors.yml
+++ b/config/locales/en/errors.yml
@@ -29,43 +29,40 @@ en:
         steps/client/has_partner_form:
           attributes:
             client_has_partner:
-              inclusion: Select whether you have a partner or not
+              inclusion: Select yes if your client has a partner
         steps/client/details_form:
           attributes:
             first_name:
-              blank: First name can’t be blank
+              blank: Enter a first name
             last_name:
-              blank: Last name can’t be blank
+              blank: Enter a last name
             date_of_birth:
-              blank: Enter date of birth
-              invalid: Date of birth must be a real date
-              invalid_day: Date of birth day is not valid
-              invalid_month: Date of birth month is not valid
-              invalid_year: Date of birth year is not valid
+              blank: Enter a date of birth
+              invalid: Enter a valid date
+              invalid_day: Enter a valid day
+              invalid_month: Enter a valid month
+              invalid_year: Enter a valid year
               year_too_late: Date of birth is too far in the future
               year_too_early: Date of birth is too far in the past
               past_not_allowed: Date of birth must be in the future
               future_not_allowed: Date of birth must be in the past
         steps/client/has_nino_form:
           attributes:
-            has_nino:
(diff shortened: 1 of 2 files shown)


config/locales/en/errors.yml
```diff
@@ -29,43 +29,40 @@ en:
...
-              invalid: Please enter a valid National Insurance number
+            nino:
+              blank: Enter a National Insurance number
+              invalid: Enter a National Insurance number in the correct format
         steps/client/contact_details_form:
           attributes:
             telephone_number:
-              blank: Telephone number can’t be blank
-              invalid: Please enter a valid UK telephone number
+              blank: Enter a UK telephone number
```
zheileman, line 57: Just thinking, do we want to give the impression only UK numbers are allowed? I know this is more a question for Carolina, as then the input label might need also to change.

config/locales/en/errors.yml
```diff
@@ -29,43 +29,40 @@ en:
...
-              blank: Correspondence address can’t be blank
-              inclusion: Select a correspondence address option 
-
+              inclusion: Select a correspondence address
         steps/address/lookup_form:
           attributes:
             postcode:
               blank: Enter postcode
-              invalid: Enter a valid UK postcode
+              invalid: Enter a postcode in the correct format
```
zheileman, line 65: Here, instead, we are forcibly validating it has to be a UK postcode so I think the error message can continue referencing this.

zheileman (approved):
👍 A few comments re: error messages but no blockers

config/locales/en/errors.yml
arthurashman: So I believe that is what the field always was on eForms, and throughout prototyping
@carolinapizatto - this is what we need right? We don't want to ask for any non-UK numbers?

config/locales/en/errors.yml
arthurashman: 'correct format' almost implies UK to my mind - as the 'correct format' will be a country specific format @carolinapizatto

zheileman: Hmm, on second thought, I think the original error message `Enter a valid UK postcode` is more accurate due to:

1. in the postcode lookup (where this error happens) only UK postcodes are allowed
2. a postcode can be in a correct format in appearance and still not be a valid real postcode, i.e. `SE1 9CA`, the format looks correct to untrained eyes, but it is not really a valid postcode (user may have entered a typo in a letter).

But happy to go with whatever is decided 😄

PR #62, opened 2022-08-15
CRIMAP-67 Add correspondence address sub-journey

zheileman:
## Description of change
We had already the home address sub-journey. The correspondence address one is the same but it customise some of the copy and links in the views accordingly.

In order to do this more clean and to copy with future changes if any, a decorator is used, that enhance the form object and expose some helper methods for the i18n, etc.
Also some logic previously existing in the form object has been moved to this decorator (the addresses counter).

Updated the decision trees to be able to enter this sub-journey from the contact details step, if the selected radio is "other address".

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-67
https://dsdmoj.atlassian.net/browse/CRIMAP-68

Screenshot 2022-08-15 at 08 49 21
Screenshot 2022-08-15 at 08 49 46
Screenshot 2022-08-15 at 08 50 02

Select "other address" in the contact details step. After finishing the correspondence address sub-journey you will get back to the home page as we don't have yet the next step.

### DIFF
--- a/app/decorators/address_form_decorator.rb
+++ b/app/decorators/address_form_decorator.rb
@@ -1,0 +1,42 @@
+class AddressFormDecorator < BaseDecorator
+  def page_title
+    ".page_title.#{i18n_key}"
+  end
+
+  def heading
+    ".heading.#{i18n_key}"
+  end
+
+  def home_address?
+    record.is_a?(HomeAddress)
+  end
+
+  def addresses
+    @addresses ||= begin
+      results = super
+
+      # Add the number of results as the first element of the collection
+      # User has to select something, otherwise there is a validation error
+      results.unshift(address_count_item) if results.any?
+      results
+    end
+  end
+
+  private
+
+  def address_count_item
+    Struct.new(:results_size, :lookup_id) do
+      def compact_address
+        I18n.translate!(:results_count, count: results_size, scope: 'steps.address.results.edit')
+      end
+    end.new(__getobj__.addresses.size)
+  end
+
+  def i18n_key
+    variant.join('.')
+  end
+
+  def variant
+    [record.person.type.underscore, record.type.underscore]
+  end
+end
--- a/spec/decorators/address_form_decorator_spec.rb
+++ b/spec/decorators/address_form_decorator_spec.rb
@@ -1,0 +1,118 @@
+require 'rails_helper'
+
+RSpec.describe AddressFormDecorator do
+  subject { described_class.new(form_object) }
+
+  let(:form_object) { double('FormObject', record: address_record) }
+
+  let(:address_record) { instance_double(Address, type: address_type, person: person_record) }
+  let(:person_record)  { instance_double(Person, type: person_type) }
+
+  let(:address_type) { nil }
+  let(:person_type)  { nil }
+
+  # NOTE: we only really have for now `Applicant`, but as these decorator is built
+  # to cope with multiple people types, we are testing it also for `Partner`
+
+  describe '#page_title' do
+    context 'for an applicant' do
+      let(:person_type) { Applicant.to_s }
+
+      context 'for a home address' do
+        let(:address_type) { HomeAddress.to_s }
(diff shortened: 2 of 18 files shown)


app/decorators/address_form_decorator.rb
```diff
@@ -0,0 +1,42 @@
...
+      end
+    end.new(__getobj__.addresses.size)
+  end
+
+  def i18n_key
+    variant.join('.')
+  end
+
+  def variant
+    [record.person.type.underscore, record.type.underscore]
```
arthurashman, line 40: is the first `.underscore` required here?

spec/decorators/address_form_decorator_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+
+    context 'for a partner' do
+      let(:person_type) { Partner.to_s }
+
+      context 'for a home address' do
+        let(:address_type) { HomeAddress.to_s }
+        it { expect(subject.page_title).to eq('.page_title.partner.home_address') }
+      end
+
+      context 'for a correspondence address' do
```
arthurashman, line 40: I don't expect this scenario ever to occur - we will have to check when we build the partner flow if we need to offer a partner correspondence address, but afaik we won't

spec/decorators/address_form_decorator_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+
+    context 'for a partner' do
+      let(:person_type) { Partner.to_s }
+
+      context 'for a home address' do
+        let(:address_type) { HomeAddress.to_s }
+        it { expect(subject.heading).to eq('.heading.partner.home_address') }
+      end
+
+      context 'for a correspondence address' do
```
arthurashman, line 70: Likewise here. Obviously no issue to leave in, but could be misleading to anyone reading the specs in the future as to what the functionality should be

app/decorators/address_form_decorator.rb
zheileman: Yes because `type` is CamelCase, i.e. `Applicant` (or `Partner` when we have it), and we want it all lowercase. For address it can be `HomeAddress` or `CorrespondenceAddress`.

spec/decorators/address_form_decorator_spec.rb
zheileman: I know, but the code has to be tested to cope with any variant, any combination of applicant/partner/address. There is a comment at the beginning of these tests to make it clear:
```
# NOTE: we only really have for now `Applicant`, but as these decorator is built
# to cope with multiple people types, we are testing it also for `Partner`
```

spec/decorators/address_form_decorator_spec.rb
zheileman: Yep, same as comment above.

We really want to test all scenarios this decorator can cope with. The code has been made generic on purpose to cope with any combination, and the decorator doesn't have any business logic so if then we don't have correspondence address for partner that is all fine as that's a business decision (we just don't show that journey to the user) but internally the code has to work for any type, as we could even have addresses for the codefendants if we wanted too.

app/decorators/address_form_decorator.rb
arthurashman: Right, understood - just in the case of person it is currently always one word, so it is essentially just to downcase

spec/decorators/address_form_decorator_spec.rb
arthurashman: Ok, sounds good

arthurashman (approved):
All looks really good to me! Nice refactor and reuse of the forms

app/decorators/address_form_decorator.rb
zheileman: Sorry my comment was not very clear. I intended to say, `underscore` will take the CamelCase and, in addition to lowercase everything, will replace the CamelCase with the underscore version. So for `HomeAddress` it becomes `home_address`, but for just 1 word it still works as intended: `Applicant` becomes `applicant` 😄

PR #65, opened 2022-08-17
Basic index page

willmcb:
## Description of change
Adds the index page "Crime applications" so the user can view applications.

Note this is a "simple" table at the moment as it is following govuk rather than MoJ styles for now.

Also adds a `govuk_formatted_date` helper method and a `full_name` method to person (this one could be moved to a presenter if its cleaner.

## Link to relevant ticket
[CRIMAP-80](https://dsdmoj.atlassian.net/browse/CRIMAP-90)
## Notes for reviewer
I wanted to keep this small so all it does is render things at the moment, further PRs will add more functionality:

- "Make new applications" will actually start an application
- At some point delete will work
- Statuses will be reflect an application's status (currently hard coded)
- Sorting of page
- Pagination (later)

There are also patterns that could be introduced that aren't included in this PR, for instance:
- Using presenters for the progress tag (this already exists but will need a little refactoring to work in this context rather than the Task list)

### After changes:
Screenshot 2022-08-17 at 11 02 29

- Create an application
- got to `/crime_applications`
- You should see the table with the list of cases
- if you click a case it will got to the its task list.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -20,4 +20,8 @@ module ApplicationHelper
   def decorate(model, decorator_class = nil)
     (decorator_class || [model.class, :Decorator].join.demodulize.constantize).new(model)
   end
+
+  def govuk_formatted_date(date)
+    date.strftime('%d %B %Y')
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -20,8 +20,4 @@ module ApplicationHelper
   def decorate(model, decorator_class = nil)
     (decorator_class || [model.class, :Decorator].join.demodulize.constantize).new(model)
   end
-
-  def govuk_formatted_date(date)
-    date.strftime('%d %B %Y')
-  end
 end
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -7,4 +7,8 @@ class Person < ApplicationRecord
   def home_address?
     home_address.address_line_one.present?
   end
+
+  def full_name
+    "#{first_name} #{last_name}"
+  end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,3 +1,7 @@
 en:
   service:
     name: Apply for criminal legal aid
+  time:
+    formats:
+      default: '%-d %B %Y'  # 3 January 2019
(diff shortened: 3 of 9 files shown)


app/helpers/application_helper.rb
```diff
@@ -20,4 +20,8 @@ def fallback_title
   def decorate(model, decorator_class = nil)
     (decorator_class || [model.class, :Decorator].join.demodulize.constantize).new(model)
   end
+
+  def govuk_formatted_date(date)
```
zheileman, line 24: In the view `#l(date)` gives you localisation of dates for free. We just have to add the formats to the `config/locales/en.yml`

```yaml
en:
  [...]
  date:
    formats:
      default: '%-d %B %Y'  # 3 January 2019
      short: '%d/%m/%Y'     # 03/01/2019
```

We can add more formats but those 2 are usually the main ones.

config/locales/en/tasklist.yml
```diff
@@ -26,6 +26,11 @@ en:
       not_applicable: Not applicable
 
   crime_applications:
+    index:
+      page_title: Your applications
+      heading: Your applications
+      start_button: Make a new application
+      delete_buttons: Delete
```
zheileman, line 33: Not that it matters a lot but should this key be singular? 😄

app/models/person.rb
```diff
@@ -7,4 +7,8 @@ class Person < ApplicationRecord
   def home_address?
     home_address.address_line_one.present?
   end
+
+  def full_name
```
zheileman, line 11: Yep, probably a decorator a bit down the line, along with some other bits that will surely be required for the dashboard table, but for now is perfectly ok.

zheileman (approved):
👍 Added a few suggestions/comments, non blocking!

config/locales/en/tasklist.yml
willmcb: Ha ha Yep it should be will change!

config/locales/en.yml
```diff
@@ -1,3 +1,7 @@
 en:
   service:
     name: Apply for criminal legal aid
+  time:
```
zheileman, line 4: I think instead of time what we want here is `date`. As you can have different formats for time in addition to date, for example:

```yaml
  time:
    formats:
      short: '%d %b %H:%M:%S'
```

PR #66, opened 2022-08-17
CRIMAP-92 Add deletion page and shared flash banner

willmcb:
## Description of change
Adds the ability for user to check and then, delete an application, or cancel deleting an application.

Also adds a flash message a partial

## Link to relevant ticket
[CRIMAP-92](https://dsdmoj.atlassian.net/browse/CRIMAP-92)

## Notes for reviewer
- Given that applications don't have a state at the moment, and cannot be "finished" as the steps aren't there, the deletion does not check that they are "in progress"
- when we can check that applications are in progress this should be added to the view logic so that it doesn't render a delete button as per the designs, and similarly the delete request should error if the application is not on progress.
- We can add these checks as we are able to.

### After changes:

Confirmation page:
Screenshot 2022-08-18 at 15 00 16

Flash message on deletion:
Screenshot 2022-08-18 at 15 00 56

- create an application
- go to dashboard
- click delete
- confirm the deletion / cancel the deletion
- return to dashboard - if you chose deletion then the flash message should appear otherwise it will show that the application is still in the list.


config/routes.rb
```diff
@@ -37,6 +37,8 @@ def show_step(name)
     end
   end
 
+  get 'crime_applications/:id/delete', to: 'crime_applications#destroy', as: :confirm_application_deletion
```
zheileman, line 40: I think we can add this to the existing resources, it feels more clean and self-contained that way, like this:

```ruby
  resources :crime_applications, except: [:new, :update] do
    get :confirm_destroy, on: :member, action: :destroy
  end
```

On a related note, I've been thinking, perhaps is best to split the confirm and the actual destroy into two separate actions, to not overload the `#destroy` one? I know I mention doing this way but now I'm not so sure 😄 

It would be something like this:

```ruby
  resources :crime_applications, except: [:new, :update] do
    get :confirm_destroy, on: :member
  end
```

Note the removal of `action: :destroy`. That way we have 2 separate actions in the controller, the usual `destroy` and this new `confirm_destroy` and we don't have to check the request verb or anything like that.

Let me know what you think or we can jump into a call.

app/controllers/crime_applications_controller.rb
```diff
@@ -5,4 +5,23 @@ def index
...
+
+  def destroy
+    @application = CrimeApplication.find(delete_params[:id])
+    @full_name = @application.applicant.full_name
+
+    if request.get?
+      render :delete
+    elsif request.delete?
+      @application.destroy
+      flash[:alert] = "#{@full_name}'s application has been permenantly deleted"
```
zheileman, line 17: i18n? and a little typo: `permenantly` -> `permanently`

We can do the flash and redirect in one go with:

```ruby
redirect_to crime_applications_path, flash: { alert: 'message' }
```

app/views/crime_applications/index.html.erb
```diff
@@ -43,7 +45,7 @@
               <strong class="govuk-tag govuk-tag--blue">In progress</strong>
             </td>
             <td class="govuk-table__cell">
-              <%= button_to '#', method: :delete,
+              <%= button_to confirm_application_deletion_path(app.id), method: :get,
```
zheileman, line 48: I think `govuk-button` is adding quite a lot of margin-bottom, we have to add an override class:
`class: 'govuk-button govuk-button--secondary govuk-!-margin-bottom-0'`

app/views/shared/_flash_banner.html.erb
```diff
@@ -0,0 +1,16 @@
+<% flash.each do |type, msg| %>
+  <% if type == 'alert' %>
```
zheileman, line 2: Wondering if we should make this more clear that is a success? the markup is very coupled to a success banner, all green, etc. Alert sounds a bit like not so successful 🤣 
But is just an observation, no blocking.

spec/requests/dashboard_spec.rb
```diff
@@ -32,8 +32,53 @@
...
+    end
+
+
+    it 'can delete an application' do
+      applicant = Applicant.find_by(first_name: "Jane")
+      app = applicant.crime_application
+
+      apps_count = CrimeApplication.count
+
+      delete "/crime_applications/#{app.id}"
```
zheileman, line 77: You can also do this if you prefer:

```ruby
expect {
  delete "/crime_applications/#{app.id}"
}.to change { CrimeApplication.count }.by(-1)
```

spec/requests/dashboard_spec.rb
```diff
@@ -32,8 +32,53 @@
       assert_select 'tbody.govuk-table__body' do
         assert_select 'tr.govuk-table__row', 1 do
           assert_select 'a.govuk-link', count: 1, text: 'John Doe'
+          assert_select 'button.govuk-button', count: 1, text: 'Delete'
```
zheileman, line 35: 👍

app/views/crime_applications/delete.html.erb
```diff
@@ -0,0 +1,29 @@
+<% title t('.page_title') %>
+<% step_header(path: crime_applications_path) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-three-quarters">
+    <h1 class="govuk-heading-xl"><%=t('.heading', applicant_name: @full_name) %></h1>
+  </div> 
+</div>
+
+<div class: "govuk-body">
```
zheileman, line 10: Typo, `class:` -> `class=`

I think all the markup has to be inside grid containers, in this case probably the `<div class="govuk-grid-column-three-quarters">` container?

If not, then we have to open a new `govuk-grid-row`.

zheileman (changes requested):
Hey I added a few comments and suggestions. Most are non blocking tho! 🤣 

There are a few strings not localised. Not a problem with that but is best to always force us to localise as then later on it becomes a chore.

app/views/shared/_flash_banner.html.erb
willmcb: Yeah I was thinking something similar

config/locales/en/tasklist.yml
```diff
@@ -24,3 +24,21 @@ en:
       not_started: Not started
       unreachable: Cannot yet start
       not_applicable: Not applicable
+
+  crime_applications:
```
zheileman, line 28: I think the rebase failed with the locales. All this should go to `dashboard.yml` now. Also be careful as I think some of the `edit` action locales are botched.

zheileman (approved):
🎉

PR #68, opened 2022-08-18
Return applications having applicants with name

zheileman:
## Description of change
This improves a bit the dashboard `#index` (list) action so that it only returns applications that have advanced enough to at the very least have an applicant with a name, as otherwise the dashboard either blows app (if there is no applicant record yet) or shows a blank name, no link (if the applicant has no name).

Move the index controller test to a `request` test `spec/requests/dashboard_spec.rb` as it makes a lot of sense for this to be a request test, and much easier to test (although slower I reckon as we are not mocking and we are using real records).

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-90

Dashboard will not blow up if there are crime_application records without an applicant relationship. Also, applicants without a `first_name` will not show. Removed N+1 query.

### DIFF
--- a/app/controllers/crime_applications_controller.rb
+++ b/app/controllers/crime_applications_controller.rb
@@ -1,7 +1,7 @@
 class CrimeApplicationsController < ApplicationController
   def index
     # TODO: scope will change as we know more
-    @applications = CrimeApplication.all
+    @applications = CrimeApplication.joins(:people).includes(:applicant).merge(Applicant.with_name)
   end
 
   def edit; end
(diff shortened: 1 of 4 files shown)


app/controllers/crime_applications_controller.rb
```diff
@@ -1,7 +1,7 @@
 class CrimeApplicationsController < ApplicationController
   def index
     # TODO: scope will change as we know more
-    @applications = CrimeApplication.all
+    @applications = CrimeApplication.joins(:people).includes(:applicant).merge(Applicant.with_name)
```
willmcb, line 4: 👍

willmcb (approved):
LGTM :)

PR #69, opened 2022-08-18
Hide secondary button on some steps

zheileman:
## Description of change
We decided it doesn't make too much sense to show the "Save and come back later" button right from the beginning when the provider didn't enter anything yet.

As we are taking the provider back to the task list, at least we want to wait until past the client details step so they enter first name / last name and there is something to show if they get back to their list of applications.

Made it so we can disable the secondary action in specific steps by passing `secondary: false`. Default it to show secondary action.

## Notes for reviewer
I still think it makes more sense to take the provider to their application list (i.e. to the "dashboard") as it will be more likely they want to either continue with another application or start a new one. The task list feels like a weird choice, but at least is easy to change if we decide so down the line.

Screenshot 2022-08-18 at 14 52 11
Screenshot 2022-08-18 at 14 52 34

After checking out the code, you must restart the server as this code is part of an initialiser that only runs once. The has partner, and client details steps will not show secondary action, any other steps after that will show it.

### DIFF
--- a/app/helpers/form_builder_helper.rb
+++ b/app/helpers/form_builder_helper.rb
@@ -5,19 +5,11 @@
 module FormBuilderHelper
   def continue_button(primary: :save_and_continue, secondary: :save_and_come_back_later)
     submit_button(primary) do
-      submit_button(secondary, secondary: true, name: 'commit_draft') if show_secondary_button?
+      submit_button(secondary, secondary: true, name: 'commit_draft') if secondary
     end
   end
 
   private
-
-  # The `save and come back later` might not always be needed,
-  # this method can have logic for show/hide.
-  # :nocov:
-  def show_secondary_button?
-    true
-  end
-  # :nocov:
 
   def submit_button(i18n_key, opts = {}, &block)
     govuk_submit I18n.t("helpers.submit.#{i18n_key}"), **opts, &block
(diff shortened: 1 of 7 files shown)


app/helpers/form_builder_helper.rb
```diff
@@ -5,20 +5,12 @@
 module FormBuilderHelper
   def continue_button(primary: :save_and_continue, secondary: :save_and_come_back_later)
     submit_button(primary) do
-      submit_button(secondary, secondary: true, name: 'commit_draft') if show_secondary_button?
+      submit_button(secondary, secondary: true, name: 'commit_draft') if secondary
```
willmcb, line 8: 👍

willmcb (approved):
LGTM 👍

spec/helpers/form_builder_helper_spec.rb
```diff
@@ -45,15 +39,18 @@
     end
 
     context 'button text can be customised' do
-      let(:draftable) { false }
-      let(:expected_markup) {
-        '<button type="submit" formnovalidate="formnovalidate" class="govuk-button" data-module="govuk-button" data-prevent-double-click="true">Find address</button>'
-      }
+      before do
+        # Ensure we don't rely on specific locales, so we have predictable tests
+        allow(I18n).to receive(:t).with('helpers.submit.find_address').and_return('Find address')
```
willmcb, line 44: Nice

willmcb (approved):
💥 Nice one 👍

PR #70, opened 2022-08-19
CRIMAP-115 Scope the steps journey to the application uuid

zheileman:
## Description of change
The task list has surfaced a pesky issue with the fact we were assuming always an application is in the session and the journey affects just that application.

However with the task list, the provider can enter in whatever application they want, and then initiate the journey for that application, meaning we have to assign that application to the session so it becomes “active“, so to speak.

Instead of that approach which can be fragile and give problems if the provider click through different applications (as it will be re-assigning the session with each one), in this PR I've refactored some code to scope all steps journey to the specific application record, so all routes identify uniquely the application record they modify.

Example of scoped routes are:

```
/steps/:id/client/has_partner
/steps/:id/client/details
/steps/:id/address/lookup/:address_id
/steps/:id/address/details/:address_id
```

Where `:id` is the UUID of the `crime_application` record, and when we have nested resources like with addresses, any other IDs are named (like `:address_id`) to avoid clashing.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-115

## Notes for reviewer
As this involve logical changes to the routes, please do `CrimeApplication.destroy_all` to start fresh otherwise existing applications might have old values in their `navigation_stack`, making the back link fail.

As above, start fresh. The developer tools has changed to link directly to the dashboard. Start a new application from the dashboard, it will generate a new record, you can also check in the footer said UUID. From the task list enter the client details journey and check all routes are scoped to this record UUID. You can at any point "save and come back later" which will take you to the dashboard, and start a new application, etc.

NOTE: despite some heavy manual testing (in addition to the automated tests) there might be some place I've missed that fail, but hopefully is easy to fix.


app/controllers/application_controller.rb
```diff
@@ -4,16 +4,18 @@ class ApplicationController < ActionController::Base
 
   add_flash_types :success
 
+  # NOTE: once we have applications linked to each provider, this
+  # needs to be scoped to the currently signed in provider
   def current_crime_application
-    @current_crime_application ||= CrimeApplication.find_by(id: session[:crime_application_id])
+    @current_crime_application ||= CrimeApplication.find_by(id: params[:id])
```
willmcb, line 10: I guess as we build in the auth system we will want to scope this by the providers office? I imagine they individuals could swap work so it makes sense to see cases related to a given office, though will need to check first if this assumption is correct.

app/controllers/crime_applications_controller.rb
```diff
@@ -9,38 +10,31 @@ def index
...
   def edit
     @tasklist = TaskList::Collection.new(
-      view_context, crime_application: crime_application
+      view_context, crime_application: current_crime_application
     )
   end
 
   def destroy
-    @application.destroy
+    current_crime_application.destroy
```
willmcb, line 26: 👍

app/controllers/crime_applications_controller.rb
```diff
@@ -9,38 +10,31 @@ def index
...
                 flash: {
-                  success: t('.success_flash', full_name: @full_name)
+                  success: t('.success_flash', applicant_name: @applicant_name)
                 }
   end
 
-  def confirm_destroy
-    render :confirm_destroy
-  end
+  def confirm_destroy; end
```
willmcb, line 33: hehe I was called confirm delete to begin with - forgot to delete this line after changing the partial to match the action

app/presenters/tasks/client_details.rb
```diff
@@ -1,7 +1,7 @@
 module Tasks
   class ClientDetails < BaseTask
     def path
-      '/steps/client/details'
+      edit_steps_client_details_path(crime_application)
```
willmcb, line 4: 👍

app/views/crime_applications/index.html.erb
```diff
@@ -8,9 +8,9 @@
 </div>
 
 <%= render partial: 'shared/flash_banner' %>
-  
+
 <div class="govuk-button-group">
-  <%= button_to '#',
+  <%= button_to crime_applications_path,
     class: 'govuk-button govuk-button--start',
     data: { module: 'govuk-button' } do; t('.start_button'); end %>
```
willmcb, line 15: I think we still need to add the chevron onto the button here so it reads "Make a new application >" so that it matches the designs. 

Tiny thing though for another PR

config/routes.rb
```diff
@@ -41,7 +41,7 @@ def show_step(name)
     get :confirm_destroy, on: :member
   end
 
-  namespace :steps do
+  scope module: :steps, path: 'steps/:id', as: :steps do
```
willmcb, line 44: 💥

willmcb (approved):
👍 LGTM

app/controllers/application_controller.rb
zheileman: Yes, there is a comment right above this method precisely for that 😄

app/views/crime_applications/index.html.erb
zheileman: Yep I didn't want to do that as didn't know how far Arthur was with its PR, just needed the create request for this refactor to be completed 👍

PR #71, opened 2022-08-22
Rename routes `crime_applications` to `applications`

zheileman:
## Description of change
Small follow-up to PR #70.

Also the steps are now part of this same path, making it more logical when thinking in terms of resources and members:

```
/applications
/applications/:id/edit
/applications/:id/steps/client/has_partner
/applications/:id/steps/client/details
/applications/:id/steps/address/details/:address_id
(...)
```

In terms of functionality everything remains the same, and also named routes are kept the same. This is only an external facing URL rename.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-115

## Notes for reviewer
Check all generated routes with: `rails routes | grep applications`

Delete existing applications in the DB as they may have wrong navigation stack. everything else should work as before.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -37,24 +37,26 @@ Rails.application.routes.draw do
     end
   end
 
-  resources :crime_applications, except: [:new, :update] do
+  resources :crime_applications, except: [:new, :update], path: 'applications' do
     get :confirm_destroy, on: :member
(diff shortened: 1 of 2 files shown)


willmcb (approved):
Nice 👍

PR #72, opened 2022-08-22
deps: Bump dependencies

zheileman:
## Description of change
Bump some minor versions of the gems and latest GOV.UK Frontend version.

No breaking changes or new features really, but good to be always in the latest version as they fix things behind the scenes.

Only thing is new version of faraday fetches from the ENV an http_proxy, (defaults to nil), thus we have to consider this in the tests.

Changelog:
https://github.com/alphagov/govuk-frontend/releases/tag/v4.3.1

After checkout, just in case, stop the server, run `rails assets:clobber` and then run bin/dev again. Don't think is purely necessary but might avoid any issues with the new assets.


willmcb (approved):
LGTM 👍

PR #74, opened 2022-08-22
CRIMAP-117 Skip address manual entry pages

zheileman:
## Description of change
If the provider finds a suitable address when using the postcode lookup, after selecting this address and submitting the form, we “skip“ the manual entry page.

This applies to home address and correspondence address.

We've been showing these pages as a way to test/debug the address lookup but the original designs and user journey was to not show them.

Still providers can access these pages when entering addresses manually, and will also be accessible from the CYA when we have it.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-117

After selecting an address from the dropdown, and submitting the form, the next page is no longer the manual entry (with the address pre-filled) but instead the following step.
For home address that is contact details, and for correspondence address that is the home page for now as we don't have case details yet.

### DIFF
--- a/app/forms/steps/address/results_form.rb
+++ b/app/forms/steps/address/results_form.rb
@@ -2,7 +2,7 @@ module Steps
   module Address
     class ResultsForm < BaseFormObject
       attribute :lookup_id, :string
-      validates :lookup_id, inclusion: { in: :address_ids }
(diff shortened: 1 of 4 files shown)


willmcb (commented):
👍 looks good - manually tested and works well

willmcb (approved):

PR #75, opened 2022-08-22
Crimap 116 tidy application statuses

willmcb:
## Description of change
Adds application status to a crime application

## Notes for reviewer
I was getting clash errors with the key "initialised" so I settled on "newly_initialised"

Another problem was that the `deafult:` in the migration wasn't working so I decided to set it in the enum declaration instead.

## Link to relevant ticket

- Create a new application, post has_partner if you get the record from the DB it should have a status of `in_progress`
- from the console create a new record with `CrimeApplication.create` - it should have a default status of `newly_initialised`

### DIFF
--- a/app/forms/steps/client/has_partner_form.rb
+++ b/app/forms/steps/client/has_partner_form.rb
@@ -13,7 +13,7 @@ module Steps
 
       def persist!
         crime_application.update(
-          attributes
+          attributes.merge(status: ApplicationStatus::IN_PROGRESS)
         )
       end
     end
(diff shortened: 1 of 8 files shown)


zheileman:
@willmcb Cool. I've seen the default doesn't work in the migration it is true. Well, let's use the `_default` on the enum then as you've done 👍 

I've pushed a few tweaks to DRY the enum values. Also the form object didn't save for me with the value object, had to cast to string (`to_s`). Let me know if it works ok for you?

zheileman (approved):
🎉

PR #77, opened 2022-08-23
Remove hardcoded status from list and add status from application

willmcb:
## Description of change
Removes hard coded "IN PROGRESS" from the dashboard and instead gets the value from the application object.

## Link to relevant ticket

## Notes for reviewer
Added a pretty_status method to the CrimeApplication, happy to move to a presenter / decorator at a later date - think its fine here for now.

- create a case > in progress should be viewable on the list page
- in the rails console generate a case with an "initialised" status it should not be viewable on the crime_applications page

### DIFF
--- a/app/controllers/crime_applications_controller.rb
+++ b/app/controllers/crime_applications_controller.rb
@@ -5,7 +5,6 @@ class CrimeApplicationsController < ApplicationController
   def index
     # TODO: scope will change as we know more
     @applications = CrimeApplication
-                    .viewable
                     .joins(:people)
                     .includes(:applicant)
                     .merge(Applicant.with_name)
--- a/app/controllers/crime_applications_controller.rb
+++ b/app/controllers/crime_applications_controller.rb
@@ -5,6 +5,7 @@ class CrimeApplicationsController < ApplicationController
   def index
     # TODO: scope will change as we know more
     @applications = CrimeApplication
+                    .viewable
                     .joins(:people)
                     .includes(:applicant)
                     .merge(Applicant.with_name)
--- a/app/models/crime_application.rb
+++ b/app/models/crime_application.rb
@@ -5,12 +5,6 @@ class CrimeApplication < ApplicationRecord
   has_many :people, dependent: :destroy
   has_many :addresses, through: :people
 
-  scope :viewable, -> { where(status: ApplicationStatus.viewable_statuses) }
-
   enum status: ApplicationStatus.enum_values,
-       _default: ApplicationStatus.enum_values[:initialised]
-
-  def pretty_status
-    status.titleize.upcase
-  end
+       _default: ApplicationStatus.enum_values[:in_progress]
 end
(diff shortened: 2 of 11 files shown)


app/controllers/crime_applications_controller.rb
```diff
@@ -5,6 +5,7 @@ class CrimeApplicationsController < ApplicationController
   def index
     # TODO: scope will change as we know more
     @applications = CrimeApplication
+                    .viewable
```
zheileman, line 8: Hmm, I think this might not be necessary, or early optimisation. We are already scoping applications by having an applicant record and in addition, this record having at least first_name entered.

Can't think of a scenario (apart from faking it through the console) where we would return applications in an `initialised` status.

I think if we have to do all this, then I'm more inclined to remove the initialised status as it is clearly adding extra complexity and mental overhead.

app/models/crime_application.rb
```diff
@@ -5,6 +5,12 @@ class CrimeApplication < ApplicationRecord
   has_many :people, dependent: :destroy
   has_many :addresses, through: :people
 
+  scope :viewable, -> { where(status: ApplicationStatus.viewable_statuses) }
+
   enum status: ApplicationStatus.enum_values,
        _default: ApplicationStatus.enum_values[:initialised]
+
+  def pretty_status
+    status.titleize.upcase
```
zheileman, line 14: Never rely on `titleize` (or similars like `humanize`, etc.) for human presented copy or translations (in fact it is not translating), it can have unexpected side effects 😬 as it is coupling the internal identifier, that now just happens to be the same we present, but it might not be in the future. Better to always go through locales.

app/views/crime_applications/index.html.erb
```diff
@@ -39,7 +39,7 @@
               LAA-a1234b
             </td>
             <td class="govuk-table__cell">
-              <strong class="govuk-tag govuk-tag--blue">In progress</strong>
+              <strong class="govuk-tag govuk-tag--blue"><%= app.pretty_status %></strong>
```
zheileman, line 42: The color of the tag also changes if it is a submitted application. Not that we have submitted anyways but we could as well leave it prepared or we will forget.

zheileman (commented):
A few comments! We can discuss on slack! not sure about the scope. We might as well just get rid of that extra complexity.

app/controllers/crime_applications_controller.rb
willmcb: Yeah I think you're right on balance - i'll remove for now

app/models/crime_application.rb
willmcb: Yep makes sense 👍

zheileman (approved):
👍 Probably next step is to introduce the decorator, so we can move the applicant name, and the status tag to it, along with (although we don't have a use for it yet) the reference, and the localisation of the date. So it all sits together in one single decorator/presenter.

PR #78, opened 2022-08-23
Tweak dashboard table css and markup

zheileman:
## Description of change
The vertical alignment (top) in the table rows was driving me crazy LOL.
Now it is aligned in the middle and universe balance has been restored.

Also added a bit of extra markup to the table according to the design system (mainly the `scope` attribute), and i18n.

### Before changes:
Screenshot 2022-08-23 at 11 44 00

### After changes:
Screenshot 2022-08-23 at 11 44 11


willmcb (approved):
Nice 👍

PR #79, opened 2022-08-23
Refinement to facilitate decorate collections

zheileman:
## Description of change
A simple refinement, that extends `Array` and `ActiveRecord::Relation` to facilitate the decoration of collection.

For example, given a decorator class named `CrimeApplicationDecorator` exists, we can decorate the list of applications in the dashboard by doing:

```ruby
@applications.decorate_each(CrimeApplicationDecorator) do |app|
```

The method yields a decorated instance. We can add to the decorator methods like `#applicant_name` (removing this method from the Applicant model as I don't think we use it anywhere else currently), a `#status_tag`, etc.

I will be doing the same with Presenters so maybe this instead of a decorator fits better the definition of presenter?


willmcb (approved):
👍

PR #80, opened 2022-08-23
Refinement to facilitate present collections

zheileman:
## Description of change
Same as we did with decorators, but now also with presenters. It behaves and have the same interface.

Refer to PR #79 for more details.


willmcb (approved):
👍 Nice

PR #81, opened 2022-08-23
Simplify collection refinements

zheileman:
## Description of change
It turned out we don't really need these to be pure refinements as we want these methods to be in all arrays and all AR relations, and we are accomplishing that just by including the modules.

Previous code was working because of that, not really because of the `refine`.

I would keep the structure as in the future we might have pure refinements that are to be used in specific classes or modules in a more surgical way.


willmcb (approved):
Nice one less code its always best code 👍

PR #82, opened 2022-08-23
Add application presenter

willmcb:
## Description of change
Adds presenter and spec for Applications + amends status copy to sentence case

## Link to relevant ticket

load the `/applications` page it should look the same as it did previously

status should be in lower case

### DIFF
--- a/app/controllers/crime_applications_controller.rb
+++ b/app/controllers/crime_applications_controller.rb
@@ -35,8 +35,8 @@ class CrimeApplicationsController < ApplicationController
   private
 
   def set_applicant_name
-    @applicant_name = CrimeApplicationPresenter
-                      .new(current_crime_application)
+    @applicant_name = helpers
+                      .present(current_crime_application)
                       .applicant_name
   end
 end
--- a/app/controllers/crime_applications_controller.rb
+++ b/app/controllers/crime_applications_controller.rb
@@ -35,6 +35,8 @@ class CrimeApplicationsController < ApplicationController
   private
 
   def set_applicant_name
-    @applicant_name = current_crime_application.applicant.full_name
+    @applicant_name = CrimeApplicationPresenter
+                      .new(current_crime_application)
+                      .applicant_name
   end
 end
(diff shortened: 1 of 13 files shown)


config/locales/en/shared.yml
```diff
@@ -2,5 +2,5 @@ en:
   shared:
```
zheileman, line 2: Not sure we will be reusing this in other places. Might be more clean to keep it all together for now in the `dashboard.yml` index section? Perhaps `crime_applications.index.status.{xxx}`
We have also in there the table headings so kind of flows nice.

spec/presenters/application_presenter_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'rails_helper'
+
+RSpec.describe ApplicationPresenter do
+  include ApplicationHelper
```
zheileman, line 4: So we don't have to include this (as it is really not the concern of this presenter) we can always use this way to instantiate the presenter:

`described_class.new(crime_application)`

Maybe cleaner?

spec/presenters/application_presenter_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  let(:presented_application) {
```
zheileman, line 6: Would this be the subject under test?

subject { ... }

zheileman (commented):
I added a few suggestions!
Nothing really blocking 🚀

config/locales/en/shared.yml
willmcb: Makes sense I will amend 👍

zheileman (approved):
🥳

app/controllers/crime_applications_controller.rb
```diff
@@ -35,6 +35,8 @@ def confirm_destroy; end
   private
 
   def set_applicant_name
-    @applicant_name = current_crime_application.applicant.full_name
+    @applicant_name = CrimeApplicationPresenter
```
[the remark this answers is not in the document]
willmcb: Seems to work okay

PR #83, opened 2022-08-24
Add right hand column to task list, markup tidy up

zheileman:
## Description of change
Last batch (maybe) of refinements, tidy up and TLC.

Added the right hand column to the task list page, showing the LAA reference. We pick this from the presenter.

Made the reference in the presenter be a bit more dynamic so is more evident clicking through different applications in the dashboard takes you to different task lists (I think the task list should include also the client's name, I've added a comment in figma design).

As we now use the presenter in the task list, and in the destroy page, made the code in the controller a bit more generic.

Tidy up some html markup and extracted to locales some hardcoded strings.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-91

Screenshot 2022-08-24 at 12 32 10

Nothing special, but If the blue bar in the right hand doesn't show, purge local assets and recompile.

### DIFF
--- a/app/presenters/task_list/collection.rb
+++ b/app/presenters/task_list/collection.rb
@@ -1,5 +1,5 @@
 module TaskList
-  class Collection < ::BasePresenter
+  class Collection < SimpleDelegator
     attr_reader :view, :crime_application
 
     delegate :size, to: :all_tasks
--- a/spec/presenters/crime_application_presenter_spec.rb
+++ b/spec/presenters/crime_application_presenter_spec.rb
@@ -1,12 +1,14 @@
 require 'rails_helper'
 
 RSpec.describe CrimeApplicationPresenter do
-  subject { 
-    described_class.new(
-      CrimeApplication.new(
-        status: 'in_progress', 
-        created_at: DateTime.new(2022, 01, 12))
-    ) 
+  subject { described_class.new(crime_application) }
+
+  let(:crime_application) {
+    instance_double(CrimeApplication,
+                    id: 'a1234bcd-5dfb-4180-ae5e-91b0fbef468d',
+                    created_at: DateTime.new(2022, 01, 12),
+                    status: 'in_progress',
+                    applicant: applicant)
   }
 
   let(:applicant) { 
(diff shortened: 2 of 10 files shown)


app/presenters/task_list/collection.rb
```diff
@@ -1,5 +1,5 @@
 module TaskList
-  class Collection < ::BasePresenter
+  class Collection < SimpleDelegator
```
zheileman, line 2: I've reverted this to not subclass from `BasePresenter` and use just the pure `SimpleDelegator` like before, as this presenter is kind of a one-off and don't require some of the functionality pushed to the `BasePresenter` (like the `present` method, etc. and maybe more in the future).

spec/presenters/crime_application_presenter_spec.rb
```diff
@@ -1,12 +1,14 @@
...
-    described_class.new(
-      CrimeApplication.new(
-        status: 'in_progress', 
-        created_at: DateTime.new(2022, 01, 12))
-    ) 
+  subject { described_class.new(crime_application) }
+
+  let(:crime_application) {
+    instance_double(CrimeApplication,
+                    id: 'a1234bcd-5dfb-4180-ae5e-91b0fbef468d',
```
zheileman, line 8: By using this fake UUID, the expectation a few lines below continues to pass as we take the first 6 chars and prepend LAA, making it still `LAA-a1234b` 😄

spec/requests/dashboard_spec.rb
```diff
@@ -89,11 +88,11 @@
 
     it 'allows a user to check before deleting an application' do
       applicant = Applicant.find_by(first_name: "Jane")
-      app = CrimeApplicationPresenter.new(applicant.crime_application)
```
zheileman, line 92: I think is best not to couple the request tests to any implementation details in the controller, as we want to just test inputs and outputs. It is fine if the implementation changes some day and the output is no longer "Jane Doe" but, for instance, "Doe, Jane". The test will fail then and alert us something has changed 😛

willmcb: Yep better to not do this for sure 👍

willmcb (approved):
LGTM 👍

PR #85, opened 2022-08-25
Add new case model

willmcb:
+ route namespace
+ blank decision tree

## Description of change
- just some backend code to enable the future work with case details.

## Link to relevant ticket
Part of [CRIMAP-74](https://dsdmoj.atlassian.net/browse/CRIMAP-74)

## Notes for reviewer
- I haven't added any tests really as this is basically just placeholder stuff.
- is there any plumbing that we need that you think would make sense here?

`rails routes | grep urn` - gives you the new routes to check

in the rails console you can add cases to an application now:

```ruby
c = Case.new(urn: '12345')
ca = CrimeApplication.new(case: c)

# then you can do
ca.case
>> # returns related case

c.crime_application
>> # returns crime application
```


config/routes.rb
```diff
@@ -52,6 +52,10 @@ def show_step(name)
         edit_step :contact_details
       end
 
+      namespace :cases do
```
zheileman, line 55: It has no real impact but I find it useful to declare the routes in the same logical order that they appear in the journey.
So perhaps we can move this namespace below the `address` namespace?

db/migrate/20220825101437_create_cases_table.rb
```diff
@@ -0,0 +1,10 @@
+class CreateCasesTable < ActiveRecord::Migration[7.0]
+  def change
+    create_table :cases, id: :uuid do |t|
+      t.string :urn
+      t.references :crime_application, null: false, foreign_key: true, type: :uuid
```
zheileman, line 5: As a precaution let's add `index: { unique: true }` so in case of race conditions it is not possible to end up with 2 case records pointing to the same `crime_application`.

PR #87, opened 2022-08-26
CRIMAP-109 Initial co-defendants DB schema and model

zheileman:
## Description of change
Just the basic schema that we will require for the co-defendants (a case has many co-defendants, a co-defendant belongs to a case).

Added the routes as well. Rename the existing `cases` namespace to be singular `case` to follow the pattern we have (like `client` or `address`).

## Link to relevant ticket

## Notes for reviewer
I'm not subclassing `Codefedant` from `Person` because apart from first_name and last_name, all other attributes are not relevant to co-defendants, don't have addresses either, and the relationship is with the Case, not with the CrimeApplication (although we expose a shortcut using `through`).

### DIFF
--- a/app/models/codefendant.rb
+++ b/app/models/codefendant.rb
@@ -1,0 +1,6 @@
+class Codefendant < ApplicationRecord
+  belongs_to :case
+
+  # Using UUIDs as the record IDs. We can't trust sequential ordering by ID
+  default_scope { order(created_at: :asc) }
+end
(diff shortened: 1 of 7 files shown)


app/models/codefendant.rb
```diff
@@ -0,0 +1,6 @@
+class Codefendant < ApplicationRecord
+  belongs_to :case
+
+  # Using UUIDs as the record IDs. We can't trust sequential ordering by ID
```
willmcb, line 4: 👍

willmcb (approved):
LGTM 👍

zheileman:
@willmcb Thanks! so quick. Don't feel obliged to review/approve PRs. Don't know if I will raise another one today but can wait until next week for sure! Enjoy! 🌴

PR #88, opened 2022-08-30
Add URN step

arthurashman:
## Description of change
Adds URN input to user journey after client contact details input
## Link to relevant ticket
[CRIMAP-74](https://dsdmoj.atlassian.net/browse/CRIMAP-74)
Follow user journey to create an application. This step is added to follow after client contact details page. 
Designs as per [Figma screen](https://www.figma.com/file/thkvnDBbQbHlSqzt1TE3lk/Crime-Apply?node-id=2773%3A29320)

image

Follow user journey to create an application. This step is added to follow after client contact details page. 
Check against designs as per [Figma screen](https://www.figma.com/file/thkvnDBbQbHlSqzt1TE3lk/Crime-Apply?node-id=2773%3A29320)

### DIFF
--- a/app/services/decisions/client_decision_tree.rb
+++ b/app/services/decisions/client_decision_tree.rb
@@ -40,7 +40,7 @@ module Decisions
           form_object.applicant
         )
       else
-        show('/home', action: :index)
+        edit('/steps/case/urn')
       end
     end
 
--- a/app/controllers/steps/case/urn_controller.rb
+++ b/app/controllers/steps/case/urn_controller.rb
@@ -1,0 +1,15 @@
+module Steps
+  module Case
+    class UrnController < Steps::CaseStepController
+      def edit
+        @form_object = UrnForm.build(
+          current_crime_application
+        )
+      end
+
+      def update
+        update_and_advance(UrnForm, as: :urn)
+      end
+    end
+  end
+end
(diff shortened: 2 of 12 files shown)


app/services/decisions/client_decision_tree.rb
```diff
@@ -40,7 +40,7 @@ def after_contact_details
           form_object.applicant
         )
       else
-        show('/home', action: :index)
+        show('/steps/case/urn', action: :edit)
```
zheileman, line 43: As this is an `edit` step it is best to do the helper edit method `edit('/steps/case/urn')` although in this case is functionally equivalent.

zheileman (approved):
Cool 🎉 

I guess we can add it later but in the figma design seems to be a revealing details link about the URN purpose. I guess we have to check with Carolina if we have the final content for that.

arthurashman:
> I guess we can add it later but in the figma design seems to be a revealing details link about the URN purpose. I guess we have to check with Carolina if we have the final content for that.

Yeah I did see that, but it didn't have an action to do anything, or any copy - happy to add later - Carol's comment on Figma suggested it needed a discussion too.

app/services/decisions/client_decision_tree.rb
arthurashman: Oh of course, I wasn't thinking. Thank you

PR #89, opened 2022-08-31
CRIMAP-109 Enter codefendant names

zheileman:
## Description of change
Follow-up to previous preparation PR #87.

This implements the controller and form object(s) needed for the CRUD step to add/edit/delete co-defendants.

Note: in this PR only add and edit actions are implemented. In a future PR the delete will also be implemented.

Also only first_name and last_name attributes in this PR. Conflict of interest is a separate ticket and will be added later, otherwise the PR grows too much.

There are some ancillary changes in this PR that were needed as this was our first CRUD step really (addresses was kind of different).

The way this works is slightly different to other steps. Basically, we have a main form object, named `CodefendantsForm` which is in charge of rendering the view which is composed of a collection of `CodefendantFieldsetForm` which are like mini form objects representing each of the individual co-defendant records.

The validation is performed in these individual co-defendant form objects but propagated to the parent form `CodefendantsForm` so errors can be rendered in the summary as well, and are indexed pointing to the specific field with error (as there may be more than 1 co-defendant and all their attributes are named the same really, we have to index them).

## Link to relevant ticket

## Notes for reviewer
No delete action yet. No conflict of interest yet. Some of the code is temporary until we have previous step (the yes-no question).

Screenshot 2022-08-31 at 10 40 20
Screenshot 2022-08-31 at 10 41 01

Start an application and make sure it has an associated `case` (go to the URN step and enter something). Then go manually to the URL for codefendants (`/applications/{crime_application_id}/steps/case/codefendants`) and it should show a first co-defendant empty, ready to fill. You can enter first name and last name, or trigger errors, etc.

### DIFF
--- a/app/validators/codefendants_validator.rb
+++ b/app/validators/codefendants_validator.rb
@@ -1,0 +1,42 @@
+class CodefendantsValidator < ActiveModel::Validator
+  attr_reader :record
+
+  def validate(record)
+    @record = record
+
+    record.codefendants.each.with_index do |codefendant, index|
+      add_indexed_errors(codefendant, index) unless codefendant.valid?
+    end
+  end
+
+  private
+
+  def add_indexed_errors(codefendant, index)
+    codefendant.errors.each do |error|
+      attr_name = indexed_attribute(index, error.attribute)
+
+      record.errors.add(
+        attr_name,
+        error.type,
+        message: error_message(codefendant, error), index: index + 1
+      )
+
+      # We define the attribute getter as it doesn't really exist
+      record.define_singleton_method(attr_name) do
+        codefendant.public_send(error.attribute)
+      end
+    end
+  end
+
+  def indexed_attribute(index, attr)
+    "codefendants-attributes[#{index}].#{attr}"
+  end
+
+  # `activemodel.errors.models.steps/case/codefendant_fieldset_form.summary.x.y`
+  def error_message(obj, error)
+    I18n.t(
+      "#{obj.model_name.i18n_key}.summary.#{error.attribute}.#{error.type}",
+      scope: [:activemodel, :errors, :models]
+    )
+  end
+end
--- a/app/views/steps/case/codefendants/edit.html.erb
+++ b/app/views/steps/case/codefendants/edit.html.erb
@@ -1,0 +1,28 @@
+<% title t('.page_title') %>
+<% step_header %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_error_summary(@form_object) %>
+
+    <h1 class="govuk-heading-xl"><%=t '.heading' %></h1>
+
+    <%= step_form @form_object do |f| %>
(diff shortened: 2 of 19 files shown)


app/validators/codefendants_validator.rb
```diff
@@ -0,0 +1,42 @@
...
+        codefendant.public_send(error.attribute)
+      end
+    end
+  end
+
+  def indexed_attribute(index, attr)
+    "codefendants-attributes[#{index}].#{attr}"
+  end
+
+  # `activemodel.errors.models.steps/case/codefendant_fieldset_form.summary.x.y`
```
arthurashman, line 35: Is this meant to be here still?

zheileman: You mean the comment? Hmm, it is intended to facilitate understanding what the final computed i18n key is (the one passed to `I18n.t`), because is not so obvious. It helped me figuring out where in the locales I had to place the messages.

app/views/steps/case/codefendants/edit.html.erb
```diff
@@ -0,0 +1,28 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_error_summary(@form_object) %>
+
+    <h1 class="govuk-heading-xl"><%=t '.heading' %></h1>
+
+    <%= step_form @form_object do |f| %>
+      <%= f.fields_for :codefendants do |c| %>
+        <%= c.govuk_fieldset legend: { text: t('.codefendant_legend', index: c.index + 1) } do %>
```
arthurashman, line 12: Was displaying the codefendant with index discussed? It's not part of the designs

arthurashman: Font size for the legend in the design also looks to be the same size as the field label.  `size: 's'` gives us 1 rem which I think is the same

arthurashman (approved):
Some tiny comment but otherwise looks great!

app/validators/codefendants_validator.rb
arthurashman: Ok, all good

app/views/steps/case/codefendants/edit.html.erb
zheileman: I think it is necessary if you want to make this page perform well in terms of accessibility. Otherwise it would be a series of repetitions not meaning anything to a screen reader. Also, facilitates the errors in the summary, which are read aloud first thing if an error happens.

And finally, in the check your answers they are indexed, which as a user would expect to match the same indexing in the page where I've entered the names.

zheileman: Re: legend font size. I'm being consistent with same sizing we already had in the legends in client personal details (Full name, Other names, Date of birth).

I think as it indicates a degree of hierarchy it makes sense to be slightly bigger in size, and in fact we are using the default size provided by the form builder so provable best to stick the the defaults IMO.

PR #90, opened 2022-08-31
CRIMAP-96 enter case type

willmcb:
## Description of change
Adds the ability to select a case type for a case

## Link to relevant ticket
[CRIMAP-96](https://dsdmoj.atlassian.net/browse/CRIMAP-96)

## Notes for reviewer
- MAAT ids are only optional for Appeal to crown court types (the last two)
- Details on changes are mandatory for Appeals to CC with a change of financial curcumstances
- Please look at the wording of the errors are they clear?

### After changes:
Screenshot 2022-08-31 at 16 00 13

### Drop downs
Screenshot 2022-08-31 at 16 00 48

---

Screenshot 2022-08-31 at 16 01 20

- go through the regular journey until you reach the case type page
- submit blank --> error
- select either appeal type > add MAAT ID > select non appeal type -> error as MAAT ID present
- do the same for last appeal type with the extra details field
- leave MAAT ID blank on an appeals case --> should not error on submission
- leave extra details page blank on Appeal CC + change of financial circs --> should error as field is blank
- if valid should then redirect to codefendants step

### DIFF
--- a/app/forms/steps/case/case_type_form.rb
+++ b/app/forms/steps/case/case_type_form.rb
@@ -13,7 +13,7 @@ module Steps
 
       validates :cc_appeal_fin_change_details,
                 presence: true,
-                if: -> { case_type.present? && CaseType.new(case_type).cc_appeal_fin_change? }
+                if: -> { case_type.present? && case_type_val.cc_appeal_fin_change? }
 
       has_one_association :case
 
@@ -42,15 +42,19 @@ module Steps
       end
 
       def appeal_maat_id
-        CaseType.new(case_type).cc_appeal? ? cc_appeal_maat_id : nil
+        case_type_val.new(case_type).cc_appeal? ? cc_appeal_maat_id : nil
       end
 
       def appeal_fin_change_maat_id
-        CaseType.new(case_type).cc_appeal_fin_change? ? cc_appeal_fin_change_maat_id : nil
+        case_type_val.new(case_type).cc_appeal_fin_change? ? cc_appeal_fin_change_maat_id : nil
       end
 
       def appeal_fin_change_details
-        CaseType.new(case_type).cc_appeal_fin_change? ? cc_appeal_fin_change_details : nil
+        case_type_val.cc_appeal_fin_change? ? cc_appeal_fin_change_details : nil
+      end
+
+      def case_type_val
+        CaseType.new(case_type)
       end
     end
   end
--- a/app/forms/steps/case/case_type_form.rb
+++ b/app/forms/steps/case/case_type_form.rb
@@ -7,7 +7,21 @@ module Steps
       attribute :previous_maat_id
       attribute :cc_appeal_fin_change_details
 
-      validates :case_type, presence: true
+      validates :case_type,
+                presence: true, 
+                inclusion: { in: :string_choices }
+
+      validates :previous_maat_id, 
+                absence: true, 
+                unless: ->{ case_is_an_appeal_type? } 
+
+      validates :cc_appeal_fin_change_details, 
+                absence: true, 
+                unless: -> { case_is_appeal_with_financial_change? }
+
+      validates :cc_appeal_fin_change_details, 
+                presence: true, 
+                if: -> { case_is_appeal_with_financial_change? }
 
       has_one_association :case
       alias kase case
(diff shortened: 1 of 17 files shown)


zheileman:
@willmcb looking great! sorry I didn't have time for a review today (fighting with co-defendants LOL). Will do it first thing tomorrow morning!

zheileman (changes requested):
Hey, looking good!
I added some comments, we can discuss over a slack call 🚀

app/forms/steps/case/case_type_form.rb
```diff
@@ -0,0 +1,59 @@
+module Steps
+  module Case
+    class CaseTypeForm < Steps::BaseFormObject
+      include Steps::HasOneAssociation
+
+      attribute :case_type
```
zheileman, line 6: Just noticed the attributes are missing their types 🤣 
In this case not a big deal as I believe it defaults to being `:string` type but better to add them just in case hehe.

app/views/steps/case/case_type/edit.html.erb
```diff
@@ -0,0 +1,34 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_error_summary(@form_object) %>
+
+    <%= step_form @form_object do |f| %>
+      <%= f.govuk_radio_buttons_fieldset(:case_type, legend: {tag: 'h1', size: 'xl'}) do %>
+        <% @form_object.choices.each do |choice| %>
+          <% if choice.value == :cc_appeal %>
+
+            <%= f.govuk_radio_button :case_type, choice.value do %>
+              <%= f.govuk_text_field :cc_appeal_maat_id, width: 'one-third' %>
```
zheileman, line 14: Let's add the `autocomplete: 'off'` to the text_field before we forget 😛

app/views/steps/case/case_type/edit.html.erb
```diff
@@ -0,0 +1,34 @@
...
+          <% elsif choice.value == :cc_appeal_fin_change %>
+
+            <%= f.govuk_radio_button :case_type, choice.value do %>
+              <%= f.govuk_text_field :cc_appeal_fin_change_maat_id, width: 'one-third' %>
+              <%= f.govuk_text_area :cc_appeal_fin_change_details %>
+            <% end %>
+
+          <% else %>
+
+            <%= f.govuk_radio_button :case_type, choice.value, link_errors: true %>
```
zheileman, line 26: Unfortunately this still sets the `link_errors: true` in more than one radio.

I think a way to do it might be to iterate the collection yielding the index, and then only adding the link_errors to the index 0, this way we make sure only the first radio in the collection has this.

```ruby
@form_object.choices.each.with_index do |choice, index|
[...]
link_errors: index.zero?
```

The other alternative is to declare all the individual radios, and to not use a loop. In terms of lines of code it would be similar, maybe more readable as there will not be if-else etc. But is a balance of convenience as if the case types value object adds or removes case types then the view will need to be adapted too. Depends if we think the case types are going to change frequently 🤷‍♂️

config/locales/en/errors.yml
```diff
@@ -81,6 +81,16 @@ en:
           attributes:
             urn:
               invalid: Enter a valid URN
+        steps/case/case_type_form:
+          attributes:
+            case_type:
+              invalid: Enter a valid case type
```
zheileman, line 87: Can be remove as unused.

config/locales/en/errors.yml
```diff
@@ -81,6 +81,16 @@ en:
...
+              inclusion: Select a case type
+            previous_maat_id:
+              present: Previous MAAT IDs are optional with appeal case types only
```
zheileman, line 90: Can be remove as unused.

config/locales/en/helpers.yml
```diff
@@ -20,6 +20,8 @@ en:
         date_of_birth: Date of birth
       steps_client_contact_details_form:
         correspondence_address_type: Correspondence address
+      steps_case_case_type_form:
+        case_type: Enter the case type
```
zheileman, line 24: I think Carolina changed some of the copy in the figma design and this is a different heading now 😢

zheileman (approved):
A few comments, but they are non blockers really so I'm approving!
Let me know if you want to discuss further! 🥳

config/locales/en/helpers.yml
```diff
@@ -64,3 +66,14 @@ en:
       steps_case_codefendants_form:
         first_name: First name
         last_name: Last name
+      steps_case_case_type_form:
+        case_type_options:
+          summary_only: Summary only
+          either_way: Either way
+          indictable: Indictable
+          already_cc_trial: Trial already in crown court
```
zheileman, line 74: Nitpick but in some places we are titleizing `Crown Court` and in others we are using all lower case `crown court` 😅

app/views/steps/case/case_type/edit.html.erb
willmcb: I think i'll go with the index option 👍 see if that's okay. I like the idea of the loop as we _can_ iterate it if you know what I mean ha

PR #91, opened 2022-09-01
CRIMAP-109 Implement co-defendants deletion

zheileman:
## Description of change
Follow-up to PR #89.

Implement the delete functionality.
It works automatically thanks to `accepts_nested_attributes_for`.

The only complication is we don't want to run validations when removing a co-defendant, as for once, removing a blank one will give us errors, but also, if any previous co-defendant are edited (like blanking any of their text fields) it will also give us errors on those unrelated fields.

This was solved by bypassing validations if any of the records are marked for destruction (`_destroy` param).

Also, on removal, we circle back to the same page, similar to adding another, but in this case we don't add another blank record obviously.

Note: we only show delete buttons after there is more than 1 co-defendant, as deleting the only one in the page would leave the page empty. If the provider don't want any co-defendant they can always go back to the previous 'yes-no' step and select no. This will (when the step is finished) delete any existing codefendants, in case there are left overs.

## Link to relevant ticket

## Notes for reviewer
Conflict of interest will be implemented in a follow-up PR.

Screenshot 2022-09-01 at 12 12 27

### DIFF
--- a/spec/forms/steps/case/codefendants_form_spec.rb
+++ b/spec/forms/steps/case/codefendants_form_spec.rb
@@ -50,6 +50,44 @@ RSpec.describe Steps::Case::CodefendantsForm do
     end
   end
 
+  describe '#any_marked_for_destruction?' do
+    # NOTE: this scenario requires real DB records to exercise nested attributes
+    context 'there are records marked for destruction' do
+      let(:application) { CrimeApplication.create }
+      let(:case_record) { Case.create(crime_application: application) }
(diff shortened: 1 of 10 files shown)


spec/forms/steps/case/codefendants_form_spec.rb
```diff
@@ -50,6 +50,44 @@
     end
   end
 
+  describe '#any_marked_for_destruction?' do
+    # NOTE: this scenario requires real DB records to exercise nested attributes
```
willmcb, line 54: 👍

willmcb (approved):
💥 Nice LGTM

PR #92, opened 2022-09-01
CRIMAP-110 Add conflict of interest radio options

zheileman:
## Description of change
Another follow-up to all previous PRs related to co-defendants.

This I think is the last one. We add the conflict of interest yes-no radios.

Pretty simple PR as almost everything has been done already in previous PRs.

Only pain was configuring the `govuk_collection_radio_buttons`, because we are using nested attributes, the locales are not automagically fetched for us and we must pass them explicitly to the legend and to both options (the yes and the no).

I've [raised an issue](https://github.com/DFE-Digital/govuk-formbuilder/issues/385) in the `govuk-formbuilder` gem repository for a possible enhancement that would make these nested attribute locales work automagically too so we could simplify a lot the code in this view if that happens.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-110

Screenshot 2022-09-01 at 17 05 46
Screenshot 2022-09-01 at 17 06 52

### DIFF
--- a/app/forms/steps/case/codefendant_fieldset_form.rb
+++ b/app/forms/steps/case/codefendant_fieldset_form.rb
@@ -5,9 +5,16 @@ module Steps
       attribute :id, :string
       attribute :first_name, :string
       attribute :last_name, :string
+      attribute :conflict_of_interest, :yes_no
 
       validates_presence_of :first_name,
                             :last_name
+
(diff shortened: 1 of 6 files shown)


willmcb:
Good idea raising an issue on DfE Formbuilder this would be nice to have for sure 👍

willmcb (approved):
Nice 👍

PR #93, opened 2022-09-02
CRIMAP-111 Co-defendant acknowledgement

arthurashman:
## Description of change
Adds the ability to acknowledge if a client has any co-defendants in their case

## Link to relevant ticket
[CRIMAP-111](https://dsdmoj.atlassian.net/browse/CRIMAP-111)

### After changes:

image

- go through the regular journey up to new has codefendant step 
- validation error on not selecting an option
- selecting 'No' redirects to /home
- selecting 'Yes' initiates co-defendant details journey

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -46,7 +46,7 @@ Layout/HashAlignment:
 
 # To be tweaked until we find the right balance
 Metrics/MethodLength:
-  Max: 14
+  Max: 12
 
 # https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/HashSyntax
 Style/HashSyntax:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -46,7 +46,7 @@ Layout/HashAlignment:
 
 # To be tweaked until we find the right balance
 Metrics/MethodLength:
-  Max: 12
+  Max: 14
 
 # https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/HashSyntax
 Style/HashSyntax:
--- a/app/views/steps/case/has_codefendants/edit.html.erb
+++ b/app/views/steps/case/has_codefendants/edit.html.erb
@@ -1,5 +1,5 @@
 <% title t('.page_title') %>
-<% step_header(path: crime_applications_path) %>
+<% step_header %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
@@ -9,7 +9,7 @@
         <%= f.govuk_collection_radio_buttons :has_codefendants, @form_object.choices,
                                              :value, legend: { tag: 'h1', size: 'xl' } %>
 
-        <%= f.continue_button(secondary: false) %>
+        <%= f.continue_button %>
     <% end %>
   </div>
 </div>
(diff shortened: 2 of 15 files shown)


.rubocop.yml
```diff
@@ -46,7 +46,7 @@ Layout/HashAlignment:
 
 # To be tweaked until we find the right balance
 Metrics/MethodLength:
-  Max: 12
+  Max: 14
```
arthurashman, line 49: For CaseDecisionTree

zheileman: I think we can revert this change because in a previous PR by @willmcb he already opted out the case decision tree from the offending cop with `rubocop:disable Metrics/MethodLength`

app/views/steps/case/has_codefendants/edit.html.erb
```diff
@@ -0,0 +1,15 @@
+<% title t('.page_title') %>
+<% step_header(path: crime_applications_path) %>
```
zheileman, line 2: No need for `path` here as we want to go back to the previous step so the path will be calculated automatically. Just `step_header` is enough like in other (intermediate) steps.

app/views/steps/case/has_codefendants/edit.html.erb
```diff
@@ -0,0 +1,15 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_error_summary(@form_object) %>
+
+    <%= step_form @form_object do |f| %>
+        <%= f.govuk_collection_radio_buttons :has_codefendants, @form_object.choices,
+                                             :value, legend: { tag: 'h1', size: 'xl' } %>
+
+        <%= f.continue_button(secondary: false) %>
```
zheileman, line 12: We want the save and come back later button in this step so let's remove `secondary: false`

I think you based some of your code on the `has_partner` step which admittedly has some unique properties given is the very first step in the journey, but that's ok, I think this and the step_header are the only peculiar things.

zheileman (commented):
Looking great! A few minor comments!

.rubocop.yml
arthurashman: We had disabled for the destination method before, on an earlier PR, and the discussion was that we would tweak the method length rather than disabling it. Are we reverting that decision then?

zheileman: Yes, we want to keep a sensible max method length. But decision trees are going to always be lengthy by definition so it doesn't make sense to keep increasing this cop max every time a decision tree grows in size.

So for decision trees that go over the max (in this case 12 lines) like the case decision tree, we opt out that method (`#destination`) in the code itself.

It allow us to still keep a sensible max value in other places and other methods.

If we keep finding ourselves going over the limit (12 right now) in other methods and places that are not just decision trees, then it may make sense to increase the limit instead of making the methods more difficult to read just to save lines of code.

Hope it makes sense 😄 We can discuss on a slack call maybe that's easier.

arthurashman: No that's fine, that makes sense. I think I just misunderstood when we discussed last time.

zheileman (approved):
Approved! 🥳

PR #95, opened 2022-09-05
Enable case details in task list

zheileman:
## Description of change
Now that we have some of the case details steps, we can update this task so it links to the first step (URN) and also to know when it is in progress or not.

We will not be able to set the "completed" status tag until we have finalised all the case details step so that will need to change sometime in the future.

Screenshot 2022-09-05 at 09 55 29

Start an application and run through the client details up until you pass the NINO step. Then go to the task list and the "Case details" task should be enabled and be able to click the link to go to the URN step.
When you run through some of the case details steps, and go back to the task list, the case details task will be marked as "in progress".

### DIFF
--- a/app/presenters/tasks/case_details.rb
+++ b/app/presenters/tasks/case_details.rb
@@ -1,8 +1,7 @@
 module Tasks
   class CaseDetails < BaseTask
-    # TODO: update when we have the case details steps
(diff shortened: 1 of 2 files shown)


willmcb (approved):
👍

PR #96, opened 2022-09-05
Cleanup temporary code from co-defendants form object

zheileman:
## Description of change
As we now have the previous yes-no step, I'm removing some of the temporary code I added in the co-defendants form object (and cleaned up the specs).

The temporary code was introduced to simplify development and testing, by creating a blank co-defendant record to start with, just by visiting the co-defendants page.
The decision tree is now in charge of handling this logic as it is bad practise to create records on GET requests.

I've also taken care of resetting any previously entered co-defendants, if the provider go back to the `yes-no` question and selects `no`.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-111

Run through the journey up until the "Does your client have any co-defendants in this case?" question. Answer `yes` and a new record will be created, go back, answer `no` and any existing co-defendant record(s) will be deleted from the database.

### DIFF
--- a/app/forms/steps/case/codefendants_form.rb
+++ b/app/forms/steps/case/codefendants_form.rb
@@ -9,14 +9,10 @@ module Steps
       validates_with CodefendantsValidator, unless: :any_marked_for_destruction?
 
       def codefendants
-        @codefendants ||= begin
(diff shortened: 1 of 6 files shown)


willmcb (approved):
LGTM 👍

PR #97, opened 2022-09-05
Enable Rubocop and default linters in ERB

zheileman:
## Description of change
Enabling Rubocop and default linters in `erblinter`.

This will run Rubocop on any ruby code found in ERB templates, with the same configuration we have for the rest of the code base. This includes our custom cops (for now just the autocomplete, as per PR #94).

However some cops are highly annoying or return false positives (i.e. Style/BlockDelimiters), thus we can tweak the configuration to disable some of these cops on ERB code, while still using them for normal non-ERB code, to get us to a good balance.

It also enables all the default linters except one I've disabled explicitly:
https://github.com/Shopify/erb-lint#linters

I had to fix a few very minor offences. I did individual commits with each of the fixes for better context.

This linter is run automatically as part of `rake` or individually with `rake erblint`.
After introducing an offence, the rake task will raise an error and show the offending code.

### DIFF
--- a/.erb-lint.yml
+++ b/.erb-lint.yml
@@ -1,13 +1,15 @@
 ---
 # https://github.com/Shopify/erb-lint#configuration
 #
-EnableDefaultLinters: false
+EnableDefaultLinters: true
 exclude:
   - 'node_modules/**/*'
   - 'lib/generators/**/*'
   - 'vendor/**/*'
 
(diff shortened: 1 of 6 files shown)


willmcb (approved):
LGTM 👍

arthurashman (approved):

PR #98, opened 2022-09-06
Improve handling of value objects in form attributes

zheileman:
## Description of change
The `yes_no` custom attribute is a QoL or syntax sugar addition to facilitate automatic casting of the commonly used yes-no radios.

This makes for simpler form objects, with less boilerplate code and more granted for free. For example there is no need to use a `#string_choices` method to validate inclusion, etc.

With these changes, we are removing the `yes_no` attribute but adding a new `value_object` attribute. This new attribute is generic and can be passed the "source" class of the values. All the rest behaves the same.

For example, before for a `YesNoAnswer` value object attribute, we had this in the form object:

```ruby
attribute :conflict_of_interest, :yes_no
```

Now we replace that with the following, which is equivalent:

```ruby
attribute :conflict_of_interest, :value_object, source: YesNoAnswer
```

Up until here is all like before, but now we can also use any other value object as well, for example:

```ruby
attribute :case_type, :value_object, source: CaseType
attribute :correspondence_address_type, :value_object, source: CorrespondenceType
```

Simplifying the form object code (we don't need to do the casting ourselves, nor use a `#string_values` method, etc.)

This is a quality of life change.

Nothing really. All should behave and work as before.


app/attributes/type/value_object.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    def type
+      :value_object
+    end
+
+    def serialize(value)
+      value.to_s
+    end
+
+    def type_cast_for_schema(value)
```
willmcb, line 18: Looking at the docs this is a `private` method that has been overridden here, and the docs say that they were trying to remove it - wondering if there is an alternative that might have more longevity.

https://apidock.com/rails/ActiveRecord/Type/Value/type_cast_for_schema

willmcb (approved):
LGTM 👍 - tiny non-blocking comment

app/attributes/type/value_object.rb
zheileman: Yep I read that. But it is not private, the comment is misleading, if you take a look at the class source code the method is public 🤣 I was overriding out of excessive precaution mostly, as the default method from the `ActiveModel::Type::Value` superclass does an `inspect` on the value, but that's probably not what we want on a ValueObject.

I'm inclined to remove the method completely. I don't think is being used and also I'm not clear what exactly "cast a value for schema dumping" means. In all my tests and debugging sessions that method is never called. Maybe is called under certain scenarios that I can't foresee now but we can always get back to this an change behaviour 🤔

app/value_objects/case_type.rb
```diff
@@ -2,7 +2,7 @@ class CaseType < ValueObject
   VALUES = [
     SUMMARY_ONLY = new(:summary_only),
     EITHER_WAY = new(:either_way),
-    INTICTABLE = new(:indictable),
+    INDICTABLE = new(:indictable),
```
arthurashman, line 5: Good catch!

app/attributes/type/value_object.rb
arthurashman: Curious how you caught this - how did you know to think to overwrite the method?

willmcb: Yeah I'm wondering if it might be more confusing to leave it in. But you never know it might stop the app from blowing up for obscure reason so I have no strong opinion either way really

arthurashman (approved):

PR #99, opened 2022-09-06
Update Rubocop Style/HashSyntax cop

zheileman:
## Description of change
Changing from `either` to `consistent` so we avoid mixing styles in a single hash.

Also setting `EnforcedStyle` to `ruby19_no_mixed_keys` for similar reasons.

This raises a few offences previously ignored tho, that I've corrected.


.rubocop.yml
```diff
@@ -52,7 +52,8 @@ Metrics/MethodLength:
 
 # https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/HashSyntax
 Style/HashSyntax:
-  EnforcedShorthandSyntax: either
+  EnforcedStyle: ruby19_no_mixed_keys
+  EnforcedShorthandSyntax: consistent
```
willmcb, line 56: I'm wondering, unless this is something that will be more common going forward, would it not be clearer to use `EnforcedShorthandSyntax: never` - I think its clearer to be honest.

I'm not too fussed either way though.

zheileman: I guess it can be more common going forward as it is the new hash style introduced in Ruby 3.1 and it has very powerful features for pattern matching etc.

I found this post quite enlightening:
https://dev.to/baweaver/ruby-3-1-shorthand-hash-syntax-first-impressions-19op

Of course it is not mandatory to use it that is why I left it as `consistent`, meaning that we can have old format, or new format, but not both in the same hash or arguments. I think is a good compromise to give us time to accommodate and use it if we see fit 😄

arthurashman: I guess if we make a decision either way in the future, we can modify the rule and the linter can auto correct to make it consistent for us

arthurashman (approved):

.rubocop.yml
zheileman: I made a quick test and rubocop will complain of the following:

```ruby
some_method(a: a, b: b, c: c, d: d)
# Style/HashSyntax: Omit the hash value.
```

Because the args can all be shortened with the new style, so rubocop wants it as follows:

```ruby
some_method(a:, b:, c:, d:)
```

However the moment any of the arguments can't be shortened because their value has a different name, like this:

```ruby
some_method(a:, b: bar, c:, d:)
# Style/HashSyntax: Do not mix explicit and implicit hash values. Include the hash value.
```

Then rubocop wants it to be consistent, like this:

```ruby
some_method(a: a, b: bar, c: c, d: d)
```

I feel like it may work going forward, but if we find any issues, let's discuss how best to proceed. This syntax is partially new to me as well so it may take a bit of time to adapt 😅

PR #100, opened 2022-09-07
Join client details journey to case journey

zheileman:
## Description of change
There was a place where these 2 journeys were not joined together, when coming from a correspondence address sub-journey.


willmcb (approved):
💥

PR #102, opened 2022-09-07
Identify deployment environment in github workflow

zheileman:
## Description of change
By explicitly identifying the environment where the workflow performs the deploy, we can set rules and assign reviewers.

This is mostly used for production environments (to assign reviewers), which we don't have yet, but for staging environment it can also be useful and we can then enable the setting "Require deployments to succeed before merging", so a merge to `main` can't happen if there is a deploy to `staging` in progress.

Screenshot 2022-09-07 at 12 03 30

Screenshot 2022-09-07 at 12 03 45


willmcb (approved):
LGTM I think this is a great idea 👍

PR #104, opened 2022-09-12
CRIMAP-125 Initial commit of offences CSV

zheileman:
## Description of change
Store the CSV file containing the pre-filtered list of offences and their attributes, and a thin layer to read it and perform basic operations like find by code (or by name if that make sense), and retrieve all offences.

This is inspired by HMCTS common platform with a very similar approach: https://github.com/ministryofjustice/hmcts_common_platform

I've made some small tweaks and modifications to better fit our purposes (also implemented the equality methods), but overall is exactly the same.

We might not need the Offences DB table anymore but for now I'm not doing anything on that front until we've validated this other alternative (CSV file).

I've just reused the same `Offence` model but removed its ActiveRecord capabilities so it is just a simple PORO/entity for now.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-125

## Notes for reviewer
Old DB table and related code not yet removed.

After checkout, open a ruby console and run, for example:

```ruby
Offence.all

Offence.find('TH68010')
Offence.find_by(code: 'TH68010')
Offence.find_by(name: 'Theft from a shop')

o = Offence.find('TH68010')
o.code
=> "TH68010"
o.name
=> "Theft from a shop"
o.offence_class
=> "F/G/K"
o.offence_type
=> "CE Either Way"
```

### DIFF
--- a/app/models/concerns/csv_queryable.rb
+++ b/app/models/concerns/csv_queryable.rb
@@ -1,0 +1,56 @@
+require 'csv'
+
+module CsvQueryable
+  extend ActiveSupport::Concern
+
+  included do
+    attr_reader :row
+  end
+
+  def initialize(row:)
+    @row = row
+  end
+
+  def ==(other)
+    other.is_a?(self.class) && other.row == row
+  end
+  alias === ==
+  alias eql? ==
+
+  def hash
+    [self.class, row].hash
+  end
+
+  class_methods do
+    def csv_filepath(filepath)
+      self.filepath = filepath
+    end
+
+    def csv_attributes(*attrs)
+      attrs.each do |name|
+        define_method(name) { row[name.to_s].strip }
+      end
+    end
+
+    def find(code)
+      find_by(code:)
+    end
+
+    def find_by(**query)
+      attr, value = query.first
+      all.find { |obj| obj.public_send(attr) == value }
+    end
+
+    def all
+      @all ||= csv.map { |row| new(row:) }
+    end
+
+    private
+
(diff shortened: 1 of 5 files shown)


app/models/concerns/csv_queryable.rb
```diff
@@ -0,0 +1,56 @@
...
+    other.is_a?(self.class) && other.row == row
+  end
+  alias === ==
+  alias eql? ==
+
+  def hash
+    [self.class, row].hash
+  end
+
+  class_methods do
```
willmcb, line 24: Much nicer than manually extending the module to add class methods as in the HMCTS' code 👍

zheileman: Yeah, I believe they did it that way as that is the non-rails way of doing it, without using `ActiveSupport::Concern`. Usually is done in gems etc, that are not assuming you will use the gem only with Rails, you might use it with Sinatra too, for example.

But in our case, there is really no point in depriving us from using Rails modules and makes for simpler and sometimes more readable code too.

app/models/concerns/csv_queryable.rb
```diff
@@ -0,0 +1,56 @@
+require 'csv'
+
+module CsvQueryable
+  extend ActiveSupport::Concern
+
+  included do
```
zheileman, line 6: I will remove this `included {}`, is not strictly necessary for attributes.

willmcb (approved):
👍

PR #107, opened 2022-09-14
CRIMAP-127 List of added offences

zheileman:
## Description of change
Little follow-up PR to #106, so we have something to "see" if the offences page is adding offences and dates as expected.

It is still kind of a mock (there is a presenter with some mock data), and no remove functionality implemented yet, that will be next in another PR, but the "change" link will work.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-127

Screenshot 2022-09-15 at 09 42 04

### DIFF
--- a/app/presenters/charge_presenter.rb
+++ b/app/presenters/charge_presenter.rb
@@ -1,0 +1,15 @@
+class ChargePresenter < BasePresenter
+  # TODO: some methods produce mock data for now
+
+  def offence_name
+    super || id
+  end
+
+  def offence_class
+    'H'
+  end
+
+  def offence_dates
+    super.pluck(:date).compact
+  end
+end
--- a/app/forms/steps/case/charges_summary_form.rb
+++ b/app/forms/steps/case/charges_summary_form.rb
@@ -7,6 +7,8 @@ module Steps
       attribute :add_offence, :value_object, source: YesNoAnswer
       validates :add_offence, inclusion: { in: :choices }
 
+      delegate :charges, to: :kase
+
       def choices
         YesNoAnswer.values
       end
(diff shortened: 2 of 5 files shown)


app/presenters/charge_presenter.rb
```diff
@@ -0,0 +1,15 @@
+class ChargePresenter < BasePresenter
+  # TODO: some methods produce mock data for now
```
willmcb, line 2: 👍

app/presenters/charge_presenter.rb
```diff
@@ -0,0 +1,15 @@
...
+  def offence_name
+    super || id
+  end
+
+  def offence_class
+    'H'
+  end
+
+  def offence_dates
+    super.pluck(:date).compact
```
willmcb, line 13: I was wondering if we could remove `nil` dates on save rather than have to `compact` them like this, happy either way though really

willmcb (approved):
Nice 👍

app/presenters/charge_presenter.rb
zheileman: This really is just a precaution, as we initialise records (charge and associated charge_dates) with nil values for the provider to populate, there is an edge case where the provider might get to the offences page, do not enter nothing, and get back or use the browser history to go back to the "offences summary page" where this offence will contain nil values and blow up (thus this compact really).

Another scenario is in the offence page, without entering anything, the provider can use "Save and come back later" leaving the current offence with nil values.

These are not scenarios that will be common. So is just protecting from nil values. Instead of a compact we can do with a query, like `super.where.not(date: nil)` instead. I used compact as is easier on the test mocks 😄

zheileman: Actually it may be we completely filter out incomplete (as in not "valid" offences -- no name or no dates) from the summary page and that will solve the edge cases. I'm working on the `destroy` action so I can look into that for my next PR.

PR #108, opened 2022-09-14
CRIMAP-72 offence dates

willmcb:
## Description of change
First PR for adding offence dates to charges page - others will follow.

The PR adds the new fields, ability to add and ability to delete dates + some basic test. 

Further PRs will add validations, clean up the erb file + i18n entries and headings.

## Link to relevant ticket

Add offence dates / delete offence dates on the Charges page

### DIFF
--- a/app/services/decisions/case_decision_tree.rb
+++ b/app/services/decisions/case_decision_tree.rb
@@ -53,14 +53,12 @@ module Decisions
     end
 
     def after_add_offence_date
-      charge = form_object.record
-      charge.offence_dates << OffenceDate.new if blank_date_required?(charge)
-      edit(:charges, charge_id: charge)
+      current_charge.offence_dates << OffenceDate.new if blank_date_required?
+      edit(:charges, charge_id: current_charge)
     end
 
     def after_delete_offence_date
-      charge = form_object.record
-      edit(:charges, charge_id: charge)
+      edit(:charges, charge_id: current_charge)
     end
 
     def after_charges_summary
@@ -82,8 +80,12 @@ module Decisions
       @case_charges ||= form_object.case.charges
     end
 
-    def blank_date_required?(charge)
-      charge.offence_dates.map(&:date).exclude?(nil)
+    def current_charge
+      @current_charge ||= form_object.record
+    end
+
+    def blank_date_required?
+      current_charge.offence_dates.map(&:date).exclude?(nil)
     end
   end
 end
--- a/app/services/decisions/case_decision_tree.rb
+++ b/app/services/decisions/case_decision_tree.rb
@@ -19,6 +19,8 @@ module Decisions
         edit(:charges_summary)
       when :add_offence_date
         after_add_offence_date
+      when :delete_offence_date
+        after_delete_offence_date
       when :charges_summary
         after_charges_summary
       else
@@ -52,9 +54,12 @@ module Decisions
 
     def after_add_offence_date
       charge = form_object.record
-      if charge.offence_dates.map(&:date).exclude?(nil)
-        charge.offence_dates << OffenceDate.new
-      end
+      charge.offence_dates << OffenceDate.new if blank_date_required?(charge)
+      edit(:charges, charge_id: charge)
+    end
+
+    def after_delete_offence_date
+      charge = form_object.record
       edit(:charges, charge_id: charge)
     end
 
@@ -75,6 +80,10 @@ module Decisions
 
     def case_charges
       @case_charges ||= form_object.case.charges
+    end
+
+    def blank_date_required?(charge)
+      charge.offence_dates.map(&:date).exclude?(nil)
     end
   end
 end
--- a/app/services/decisions/case_decision_tree.rb
+++ b/app/services/decisions/case_decision_tree.rb
@@ -17,6 +17,8 @@ module Decisions
         after_codefendants
       when :charges
         edit(:charges_summary)
+      when :add_offence_date
+        after_add_offence_date
       when :charges_summary
         after_charges_summary
       else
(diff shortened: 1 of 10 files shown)


app/views/steps/case/charges/edit.html.erb
```diff
@@ -10,6 +10,16 @@
     <%= step_form @form_object do |f| %>
       <%= f.govuk_text_field :offence_name, autocomplete: 'off' %>
 
+      <%= f.fields_for :offence_dates do |od| %>
+        <%= od.govuk_date_field :date, maxlength_enabled: true, legend: { size: 'm' } %>
```
zheileman, line 14: We might have to wrap each of the date field in a `govuk_fieldset`, not sure if that was purely needed for the validation and thus proper linking or maybe was just for styling etc. We will figure out next with validations, no worries for now!

spec/controllers/steps/case/charges_controller_spec.rb
```diff
@@ -4,6 +4,63 @@
   let(:form_class) { Steps::Case::ChargesForm }
   let(:decision_tree_class) { Decisions::CaseDecisionTree }
 
+  let(:crime_application) { CrimeApplication.create }
```
zheileman, line 7: If possible let's wrap these contexts and their associated lets in a new describe block as the lets are not needed for the other scenarios (`#edit` and `#update`) and make tests more isolated and less prone to clashing.

spec/controllers/steps/case/charges_controller_spec.rb
```diff
@@ -4,6 +4,63 @@
...
+      expect(
+        subject
+      ).to receive(:update_and_advance).with(
+        form_class, 
+        record: charge_record, 
+        as: :add_offence_date
+      )
+
+      put :update, params: { 
+        id: crime_application.id,
```
zheileman, line 24: It has no real impact, but no need to call `#id` explicitly, as just passing the record will implicitly call `#to_param` on the record, which will return (usually) the ID (in this case UUID).
This is valid anywhere in the rails app where we build URLs, and also in named urls/paths.

spec/forms/steps/case/charges_form_spec.rb
```diff
@@ -28,4 +28,38 @@
       end
     end
   end
+
+  describe '#offence_dates' do
+    it 'returns a list of offence date fieldset forms' do
+      charge_record.offence_dates << [
```
zheileman, line 34: I think if we can build the dates from the params is better as that's how this form is created in reality, through a post/put request to the controller, and passing the params.
Take a look at the `codefendants_form_spec` and the `let(:codefendants_attributes)`

spec/forms/steps/case/charges_form_spec.rb
```diff
@@ -28,4 +28,38 @@
...
+        OffenceDate.new(date: "02, 02, 2000"),
+        OffenceDate.new(date: "03, 10, 1998"),
+        OffenceDate.new(date: "11, 12, 2021")
+      ]
+
+      subject.offence_dates.each do |form|
```
zheileman, line 40: Maybe shorter:
`expect(subject.offence_dates).to all(be_a(Steps::Case::OffenceDateFieldsetForm))`

spec/forms/steps/case/charges_form_spec.rb
```diff
@@ -28,4 +28,38 @@
...
+        expect(form).to be_a(Steps::Case::OffenceDateFieldsetForm)
+      end
+      expect(subject.offence_dates.size).to be(3)
+    end
+  end
+
+  describe '#show_destroy?' do
+    it 'returns false if there are fewer than 2 offence dates' do
```
zheileman, line 48: Tiny thing, this scenario and the following one has the same `fewer than 2 offence dates` in their descriptions 😛

spec/forms/steps/case/offence_date_fieldset_form_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
+
+RSpec.describe Steps::Case::OffenceDateFieldsetForm do
+  let(:arguments) { {
+    crime_application: crime_application,
+    record: OffenceDate.new(date: "02, 07, 2020")
```
zheileman, line 6: A suggestion, it can be simplified to just:

```ruby
id: record_id
```

and then in a let:

```ruby
let(:record_id) { '12345' }
```

Because in reality this ID comes from the params, not from the record instance. It also allows for simpler tests.

spec/forms/steps/case/offence_date_fieldset_form_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  subject { described_class.new(arguments) }
+
+  describe 'validations' do
+    # TODO: validations
+  end
+
+  describe '#persisted?' do
+    context 'when form has an id' do
+      it 'it returns true' do
+        arguments[:id] = "1232456"
```
zheileman, line 20: And then we don't need this as per comment above.

zheileman (changes requested):
🥳 That's great, we are getting there.
I added a few comments and suggestions, I don't think there are any blockers really so feel free to pick the ones you are happy with or agree with the changes, leave the others, some are not yet actionable really.

app/services/decisions/case_decision_tree.rb
```diff
@@ -48,6 +52,17 @@ def after_codefendants
       edit_new_charge
     end
 
+    def after_add_offence_date
+      charge = form_object.record
+      charge.offence_dates << OffenceDate.new if blank_date_required?(charge)
```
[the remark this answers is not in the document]
willmcb: Yep I was thinking the same, just wanted it to act realistically for the time being 👍

spec/forms/steps/case/charges_form_spec.rb
willmcb: Copypasta error 🤦

app/services/decisions/case_decision_tree.rb
```diff
@@ -66,5 +81,9 @@ def edit_new_charge
     def case_charges
       @case_charges ||= form_object.case.charges
     end
+
+    def blank_date_required?(charge)
+      charge.offence_dates.map(&:date).exclude?(nil)
```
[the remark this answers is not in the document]
willmcb: Yep for sure :)

zheileman (approved):

PR #109, opened 2022-09-15
CRIMAP-130 Implement offence destroy action

zheileman:
## Description of change
Another follow-up to previous offence summary page PRs.

This one implements the "Remove" link.
It follows the same pattern we already did for the applications dashboard, with a confirmation page, and shows a flash banner as well.

I tweaked a bit the routes `crud_step` helper to simplify this and to expose the `confirm_destroy` action when we have a `destroy`.

As agreed, if the offence being removed is the last one in the list, we redirects to a brand new offence, but if there are more offences left, we redirect back to the summary page.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-130

## Notes for reviewer
There are some conversations around the wording "delete" or "remove" and also around the dates format. I'm leaving that out of this PR for now, once decided, we can change whatever is needed.

Screenshot 2022-09-15 at 15 36 07
Screenshot 2022-09-15 at 15 36 14

When deleting the last offence in the list, it goes back to add an offence page:
Screenshot 2022-09-15 at 15 35 38

Have some offences and go to the summary page, you can then delete them.

### DIFF
--- a/.erb-lint.yml
+++ b/.erb-lint.yml
@@ -10,6 +10,8 @@ exclude:
 linters:
   SpaceAroundErbTag:
     enabled: false
+  PartialInstanceVariable:
+    enabled: true
   ErbSafety:
     enabled: true
     better_html_config: .better-html.yml
--- a/app/controllers/steps/case/charges_controller.rb
+++ b/app/controllers/steps/case/charges_controller.rb
@@ -13,6 +13,25 @@ module Steps
         )
       end
 
+      def destroy
+        charge_record.destroy
+
+        if case_charges.reload.any?
+          redirect_to edit_steps_case_charges_summary_path,
+                      success: t('.success_flash')
+        else
+          # If this was the last remaining record, redirect
+          # to the charges page with a new blank one
+          charge = case_charges.create!
+          redirect_to edit_steps_case_charges_path(charge_id: charge),
+                      success: t('.success_flash')
+        end
+      end
+
+      def confirm_destroy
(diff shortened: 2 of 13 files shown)


app/controllers/steps/case/charges_controller.rb
```diff
@@ -13,6 +13,27 @@ def update
...
+      def destroy
+        charge_record.destroy
+
+        if case_charges.reload.any?
+          redirect_to edit_steps_case_charges_summary_path,
+                      success: t('.success_flash')
+        else
+          # If this was the last remaining record, redirect
+          # to the charges page with a new blank one
+          charge = case_charges.create!(
```
zheileman, line 25: There is a bit of repetition here and in the decision tree. For now I've not addressed this as is very minor, but if we see we have to use this in more places, it may be we could expose a method on the relationship named `create_blank!` or something.

zheileman: I've already cleanup the duplication and extracted the creation of the blank associated date to `Case` has_many relationship in a recent commit 😄

.erb-lint.yml
```diff
@@ -10,6 +10,8 @@ exclude:
 linters:
   SpaceAroundErbTag:
     enabled: false
+  PartialInstanceVariable:
```
willmcb, line 13: 👍

willmcb (approved):
Nice one, works great! 👍

PR #111, opened 2022-09-20
CRIMAP-124 Add datalist for the non-JS offences dropdown

zheileman:
## Description of change
Individual commits for more context.

This datalist is populated from the CSV, and offences are assigned to the Charge as value-objects making use of `composed_of` where we find the most suitable offence by name and return it if found, otherwise we assume the offence is being "created" by the provider, in which case we will not have an `#offence` (returns nil) and thus we lack additional attributes like offence class and type.

Add basic presence validation on the `offence_name` attribute.

Update the `ChargePresenter` as almost everything is now retrieved from the offence instance (or returns `nil` for manual offences).

In follow-up PRs we will add some javascript to be able to style the dropdown and to perform better in terms of accessibility. This is the non-JS version.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-124
https://dsdmoj.atlassian.net/browse/CRIMAP-73

## Notes for reviewer
I've only added the presence validation on the `offence_name` attribute. There will be additional validation on the dates but that is part of another ticket.
Some copy is still TBD, for example the hint text in the input should be more explicit about the dropdown etc.

Screenshot 2022-09-20 at 11 56 42
Screenshot 2022-09-20 at 11 56 55

Go to the charges page, and enter a few letters on the offence name input, you should see a dropdown with offences that contain those letters. You can either select one, or keep typing and "create" your own offence.

If you select an offence from the list, it will have a designated class on the summary page. For manually entered offences, these will not have class.

### DIFF
--- a/app/forms/steps/case/charges_form.rb
+++ b/app/forms/steps/case/charges_form.rb
@@ -2,6 +2,7 @@ module Steps
   module Case
     class ChargesForm < Steps::BaseFormObject
       attribute :offence_name, :string
+      validates :offence_name, presence: true
 
       delegate :offence_dates_attributes=, to: :record
 
--- a/app/models/charge.rb
+++ b/app/models/charge.rb
@@ -1,12 +1,12 @@
 class Charge < ApplicationRecord
   belongs_to :case
 
-  # This will potentially change to retrieve the offence from the CSV
-  # belongs_to :offence, optional: true
-
   has_many :offence_dates, dependent: :destroy
   accepts_nested_attributes_for :offence_dates, allow_destroy: true
 
   # Using UUIDs as the record IDs. We can't trust sequential ordering by ID
   default_scope { order(created_at: :asc) }
+
+  composed_of :offence, mapping: %i[offence_name name],
+              constructor: :find_by_name, allow_nil: true
 end
--- a/app/models/concerns/csv_queryable.rb
+++ b/app/models/concerns/csv_queryable.rb
@@ -26,7 +26,7 @@ module CsvQueryable
 
     def csv_attributes(*attrs)
       attrs.each do |name|
-        define_method(name) { row[name.to_s].strip }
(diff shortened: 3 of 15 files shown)


zheileman:
NOTE: in my tests, I've seen some inconsistencies across browsers. It seems most of them like to show 4 options before having to scroll, except Safari that for some reason reduces it to 3 options shown.
There are also some differences in styling depending on the browser, being **Edge** the better looking of all, IMO.

However, there is a breaking thing with the `Class XYZ` we add to the datalist.
It seems to work in all desktop browsers I've tried **except in Firefox**. More testing is required in mobile devices.

So far: Edge, Brave, Chrome, Safari and Opera all show the "Class" usually in a different text colour and below the offence name which is very nice. This class however is only a visual hint and is not submitted in the request to the server nor used in the input text (only the option value is used in the input text, at least with non-JS).

**Firefox only shows the class**, behaving completely differently to the rest of browsers. For this reason if we want to consider this a cross-browser solution we will need to either remove the class in the non-JS version (this will not be a problem for the JS-styled version tho), or remove it only for those browsers that don't show it properly. Some investigation on how best to do this might be needed, but for now I've leave the class in this PR, and will get back to it probably after the javascript implementation.

### Example on Edge:
Screenshot 2022-09-20 at 12 17 54

### Example on Opera:
Screenshot 2022-09-20 at 12 24 31

### Example on Chrome:
Screenshot 2022-09-20 at 12 21 43

### Example on Brave:
Screenshot 2022-09-20 at 12 22 56

### Example on Safari:
Screenshot 2022-09-20 at 12 22 04

### Example on Firefox
How it looks on Firefox (it only shows the Class, and only searches in the Class):
Screenshot 2022-09-20 at 12 15 16

willmcb (approved):
💥 👍 Looks great :)

PR #112, opened 2022-09-21
CRIMAP-72 Offence Date Validations

willmcb:
## Description of change
Adds validations and fixes the labels and hint text of the offence dates on the charges page 

## Link to relevant ticket

## Notes for reviewer
- you views on accessibility would be appreciated
- note: I haven't removed the blank date initialisation in the cases decision tree in the PR. Can raise another one when we decide the best way to do this.

### Before changes:
Screenshot 2022-09-21 at 10 09 59

### After changes:
Screenshot 2022-09-21 at 10 10 32

- try to submit dates in the future 
- try to submit blank dates

### DIFF
--- a/app/forms/steps/case/charges_form.rb
+++ b/app/forms/steps/case/charges_form.rb
@@ -5,6 +5,8 @@ module Steps
 
       delegate :offence_dates_attributes=, to: :record
 
+      validates_with ChargesValidator 
+
       def offence_dates
         @offence_dates ||= record.offence_dates.map do |offence_date|
           OffenceDateFieldsetForm.build(
--- a/app/forms/steps/case/charges_form.rb
+++ b/app/forms/steps/case/charges_form.rb
@@ -5,7 +5,7 @@ module Steps
 
       delegate :offence_dates_attributes=, to: :record
 
-      validates_with ChargesValidator 
+      validates_with ChargesValidator
 
       def offence_dates
         @offence_dates ||= record.offence_dates.map do |offence_date|
--- a/app/views/steps/case/charges/edit.html.erb
+++ b/app/views/steps/case/charges/edit.html.erb
@@ -14,13 +14,30 @@
       <%= f.govuk_text_field :offence_name, autocomplete: 'off' %>
 
       <%= f.fields_for :offence_dates do |od| %>
-        <%= od.govuk_fieldset legend: { text: t('.offence_dates_legend', index: od.index + 1) } do %>
-          <%= od.govuk_date_field :date, maxlength_enabled: true, legend: { size: 'm' } %>
-          <%= od.button t('.remove_button'), type: :submit, value: '1', name: od.field_name(:_destroy, index: od.id),
-                       class: %w[govuk-button govuk-button--warning],
-                       data: { module: 'govuk-button' } if @form_object.show_destroy? %>
-          <% end %>
+        <% if od.index.zero? %>
+          <%= od.govuk_fieldset legend: { text: t('.offence_dates_legend', index: od.index + 1), described_by: '#add-or-remove-date' } do %>
+            <div class='govuk-hint'>
+              <%= t('.offence_dates_hint') %>
+            </div>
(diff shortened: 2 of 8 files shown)


config/locales/en/errors.yml
```diff
@@ -111,3 +111,11 @@ en:
           attributes:
             add_offence:
               inclusion: Select yes if you want to add another offence
+        steps/case/offence_date_fieldset_form:
+          attributes:
+            date:
+              future_not_allowed: Offence date cannot be in the future
```
zheileman, line 117: We need to add here also the `blank` otherwise on error it looks like this on the field:

Screenshot 2022-09-21 at 11 17 12

In fact, I would suggest copy and paste all date errors from `en.activemodel.errors.models.steps/client/details_form.attributes.date_of_birth` (tweaking the copy and/or removing the ones that can't trigger) and we are good to go.

It may seem like a lot of repetition but as we want unique and explicit error messaging we can't generalise really.

app/views/steps/case/charges/edit.html.erb
```diff
@@ -10,14 +10,34 @@
...
     <%= step_form @form_object do |f| %>
+
       <%= f.govuk_text_field :offence_name, autocomplete: 'off' %>
 
       <%= f.fields_for :offence_dates do |od| %>
-        <%= od.govuk_date_field :date, maxlength_enabled: true, legend: { size: 'm' } %>
-        <%= od.button t('.remove_button'), type: :submit, value: '1', name: od.field_name(:_destroy, index: od.id),
-                     class: %w[govuk-button govuk-button--warning],
-                     data: { module: 'govuk-button' } if @form_object.show_destroy? %>
+        <% if od.index.zero? %>
```
zheileman, line 17: I'm looking at doing this in a simpler and cleaner way, will share the solution 👍

app/forms/steps/case/charges_form.rb
```diff
@@ -5,6 +5,8 @@ class ChargesForm < Steps::BaseFormObject
 
       delegate :offence_dates_attributes=, to: :record
 
+      validates_with ChargesValidator
```
zheileman, line 8: We will want to also implement `unless: :any_marked_for_destruction?` otherwise you can't delete blank dates if the user mistakenly add another date and then they try to remove it while being blank/invalid.
Same I did for codefendants.

However this can be done in a separate PR, not a problem.

config/locales/en/errors.yml
```diff
@@ -111,3 +111,11 @@ en:
...
+          summary:
+            date:
+              blank: Offence dates cannot be blank
```
zheileman, line 120: I believe it will still worth to have indexed summary errors otherwise if more than one date fails validation there will be repetition of messages in the summary and is something we want to avoid for screen readers.
Even if we don't show in the page "Offence date 1", "Offence date 2"... we will have that as visually hidden in each of the dates and thus it makes sense, but we can discuss along with the simplification of the mark-up 👍

zheileman (commented):
Yep, let's take a look at the markup. I think it can be simplified.

config/locales/en/errors.yml
willmcb: Yep course - I didn't know that it had individual errors for constituent parts of the date 👍

zheileman:
@willmcb I've pushed an alternative simplified markup. It also adds indexed hidden legends for each date so screen readers gets the context if they are in the first date, second date, etc...
More info in the commit message. We can discuss later 👍

config/locales/en/errors.yml
zheileman: If we do that it would be just like this: `blank: Enter offence date %{index}` because the `index` is already exposed.
I suppose is a matter of carefully crafting proper error messages if we do that. Probably one for Carolina.

It can be done as a separate PR anyways.

zheileman (approved):
I'm approving to get this moving 😄 
The refinements can be done in a separate PR once we also merge the offences dropdown as there is an extra validation we require, in the offence name, that we need to sort out once both pieces of work are glued together.

PR #113, opened 2022-09-22
CRIMAP-72 Add .any_marked_for_destruction? to ChargesForm + test

willmcb:
## Description of change
Allows blank offence dates to be deleted from the Charges page as it makes validations dependant on there not being "_destroy" params in the form attributes sent to the server.

## Link to relevant ticket

- on Charges page add another date, so there are is a blank date input on the form. Then try to delete this without adding any date into it. The page should reload without the blank date, and not error.


zheileman (approved):
Wonderful! 💯

PR #114, opened 2022-09-22
Present the offence class(es), comma separated

zheileman:
## Description of change
This is a tweak to adapt the way we show the offence class in the views, to the requirements of the design.

First, we unify the presentation of the class in just one single place (`OffencePresenter`). This presenter works with `Offence` instances thus always will have class attributes.

Additionally modify the existing `ChargePresenter` use the new `OffencePresenter` also to present information related to the offence, including the class.

Last, a new `OffenceCollectionDecorator` takes care of the offences dropdown, also presenting them with the new `OffencePresenter` so class is presented comma-separated as well.

Screenshot 2022-09-22 at 17 09 48
Screenshot 2022-09-22 at 17 12 57
Screenshot 2022-09-22 at 17 13 13

See in the dropdown the different classes, and add some with one or more classes to see them in the basket too.

### DIFF
--- a/app/presenters/charge_presenter.rb
+++ b/app/presenters/charge_presenter.rb
@@ -1,13 +1,9 @@
 class ChargePresenter < BasePresenter
-  # TODO: some methods produce mock data for now
+  delegate :code,
+           to: :offence, allow_nil: true, prefix: true
 
-  def offence_name
-    super || id
-  end
-
-  def offence_class
-    'H'
-  end
+  delegate :offence_class, :offence_type,
+           to: :offence, allow_nil: true, prefix: false
 
   def offence_dates
     super.pluck(:date).compact
--- a/app/presenters/charge_presenter.rb
+++ b/app/presenters/charge_presenter.rb
@@ -1,9 +1,10 @@
 class ChargePresenter < BasePresenter
-  delegate :code,
-           to: :offence, allow_nil: true, prefix: true
-
   delegate :offence_class, :offence_type,
-           to: :offence, allow_nil: true, prefix: false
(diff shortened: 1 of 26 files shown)


app/presenters/charge_presenter.rb
```diff
@@ -1,9 +1,10 @@
 class ChargePresenter < BasePresenter
-  delegate :code,
```
zheileman, line 2: Removed this `code` delegation as we are not using it and is likely we will not need it after all.
It is of course possible to access it through the `offence` method, but from a presenter point of view it doesn't make sense to have it unless we are going to present it.

willmcb: That makes sense - confusing if we done use it for sure!

willmcb (approved):
👍

willmcb:
Only thing I was thinking is that we could have `map_with_decorator` / `map_with_presenter` and it might look more rubyish small thing though

zheileman:
@willmcb Yeah, I was unsure on the naming tbh. I went with those naming because didn't want to imply a yielding method.

For example, `#map_with_decorator` it feels to me like it will yield each element, like this:

```ruby
@collection.map_with_decorator(Decorator) do |elem|
[...]
end

# as in a traditional map:
@collection.map do |elem|
[...]
end
```

So I thought, maybe naming it in the "past", implies that the operation was performed, and thus it returns the result, not giving a sense of yielding each element? 🤔

willmcb:
Yeah I can see that, it does sort of mix meanings with the traditional `each_with_object` kind of way - I think what we have is fine tbh 👍

PR #115, opened 2022-09-23
CRIMAP-129 Enhancement of offence datalist with JS

zheileman:
## Description of change
Preliminary discovery work and first rough version of the javascript enhancement in the offences `datalist`.

This produces a "suggestions" dropdown, with consistent look and feel across all browsers, when javascript is enabled.

When javascript is disabled the normal "datalist" is shown instead so functionality is not lost.

Fair disclaimers:

- We are piggybacking on a draft PR for an unfinished work, not part of the GOV.UK Frontend yet. And it might not be part of it for a long time, if ever.

- It just doesn't work out of the box. I had to made a few modifications to the javascript.

- Additionally, to show the offence class, I also had to extend the functionality with a `data-caption` feature. This has an impact on the render of the datalist markup too.

- I've marked all this modifications with `[change: xyz]` so it is easy to identify them.

Finally, for the sake of simplification and this initial version, I've removed the "Class Z, Y, Z" from the non-javascript datalist. Keeping it required additional changes to the javascript and I wanted to keep these modifications as light and limited as possible.

Some modifications I'm thinking contributing back to the draft PR.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-129

## Notes for reviewer
I've added disclaimers.

Screenshot 2022-09-23 at 16 03 26

Checkout. Recompile assets, if there is any issue clobber and recompile. Go to the offences page and enter 2 or more chars.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -3,5 +3,9 @@
 @import "local/info_card";
 @import "local/task_list";
 
+// NOTE: suggestions input component not yet part of GOV.UK frontend
+// https://github.com/alphagov/govuk-frontend/pull/2453
+@import "local/suggestions";
+
 // App-specific custom styles and overrides
 @import "local/custom";
--- a/app/assets/stylesheets/local/suggestions.scss
+++ b/app/assets/stylesheets/local/suggestions.scss
@@ -1,0 +1,43 @@
+.govuk-input__suggestions-header {
+  margin-bottom: 0;
+  padding: 5px;
+  @include govuk-font($size: 19, $weight: bold);
+  border-width: $govuk-border-width-form-element $govuk-border-width-form-element 0 $govuk-border-width-form-element;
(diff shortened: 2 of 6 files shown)


zheileman:
@willmcb found the bug 🤣 take a look at the last commit. Speak Monday!

willmcb (approved):
Looks good 👍 

I see as well the commit that fixed the intermittent bug, good job.

Do you think there is any need to add a test for this? Perhaps a system test might be the easiest to add as I we can test JS with them.

Maybe it can wait for us to add in an end-to-end test though.

zheileman:
@willmcb do request specs execute javascript? If so, we could have, as we did for the summary page, a request spec with a few scenarios to cover adding offences with dates. Even if it does not run javascript, the fallback datalist might be tested.
But I wouldn't add those tests to this PR so we keep this focused just on what was needed to make the suggestions work.

PR #117, opened 2022-09-27
Adds basic "date stamp" page without any business logic

willmcb:
## Description of change

PR adds a basic, essentially static, "date stamp" page. Logic etc. to follow in a future PR.

## Link to relevant ticket
[CRIMAP-49](https://dsdmoj.atlassian.net/browse/CRIMAP-49)

### Before changes:
n/a - this is a new page

### After changes:
Screenshot 2022-09-27 at 16 25 48

- currently this is "outside" the step flow. you need to hit the URL directly. Also note the buttons don't really do anything at the moment - this will come in a future PR.
- open an application and go to `/case/date_stamp`

### DIFF
--- a/app/views/steps/case/date_stamp/show.html.erb
+++ b/app/views/steps/case/date_stamp/show.html.erb
@@ -3,7 +3,7 @@
 
 <div class="govuk-grid-row app-inverted--card app-inverted-text">
   <div class="govuk-grid-column-full">
-    <h1 class="govuk-heading-l app-inverted-text">Your application has been date stamped</h1>
+    <h1 class="govuk-heading-l app-inverted-text"><%= t('.heading')%></h1>
 
     <div class="govuk-button-group">
       <%= link_button t('helpers.submit.save_and_continue'), '#', class: 'app-button--m app-button--inverted' %>
--- a/app/views/steps/case/date_stamp/show.html.erb
+++ b/app/views/steps/case/date_stamp/show.html.erb
@@ -6,12 +6,8 @@
     <h1 class="govuk-heading-l app-inverted-text">Your application has been date stamped</h1>
 
     <div class="govuk-button-group">
-      <%= link_to '#',
-                  class: 'govuk-button app-button--m app-button--inverted',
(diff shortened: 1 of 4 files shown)


app/views/steps/case/date_stamp/show.html.erb
```diff
@@ -0,0 +1,18 @@
+<% title t('.page_title') %>
+<% step_header %>
+
+<div class="govuk-grid-row app-inverted--card app-inverted-text">
+  <div class="govuk-grid-column-full">
+    <h1 class="govuk-heading-l app-inverted-text">Your application has been date stamped</h1>
```
zheileman, line 6: We can either convert this view to a localised view (adding extension `.en.html.erb`) or extract this string to locales.
I think as this page is so minimal regarding content, probably best just to extract to locales as all the other copy in the page is already in locales.

app/views/steps/case/date_stamp/show.html.erb
```diff
@@ -0,0 +1,18 @@
...
+
+    <div class="govuk-button-group">
+      <%= link_to '#',
```
zheileman, line 9: You can use `#link_button` helper, it adds some additional attributes for you. You can still provide any additional classes like `app-button--inverted`.

For the locales maybe we can use the existing `helpers.submit.save_and_continue` and `helpers.submit.save_and_come_back_later` to keep it in one place.

config/locales/en/steps.yml
```diff
@@ -78,6 +78,11 @@ en:
         edit:
           page_title: Enter the case type
           heading:  Enter the case type
+      date_stamp:
+        show:
+          page_title: Your application has been date stampled
```
zheileman, line 83: Little typo 😄

zheileman (approved):
Approving with some minor, non blocking, suggestions! 👍

app/views/steps/case/date_stamp/show.html.erb
willmcb: I think i'll just add it to the locals

PR #118, opened 2022-09-28
CRIMAP-75 Implement basic court hearing step

zheileman:
## Description of change
This includes the route, controller, view and form object as well as new required DB attributes on the `cases` table.

Court name input text, does not implement the dropdown or typeahead / suggestions yet. That will be part of a separate task/PR as it requires further research around where this data is coming from or how to "store" it in our application (if we do).

Co-Authored-By: timpeat <tim@timpeat.com>
Co-Authored-By: Chris Gannon <cjgannon@gmail.com>

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-75

## Notes for reviewer
No courts suggestions/typeahead yet. It will be part of a follow-up PR.

Screenshot 2022-09-28 at 13 13 38
Screenshot 2022-09-28 at 13 13 45

This step comes after the offences "basket".

### DIFF
--- a/app/forms/steps/case/hearing_details_form.rb
+++ b/app/forms/steps/case/hearing_details_form.rb
@@ -1,0 +1,23 @@
+module Steps
+  module Case
+    class HearingDetailsForm < Steps::BaseFormObject
+      include Steps::HasOneAssociation
+      has_one_association :case
+
+      attribute :hearing_court_name, :string
+      attribute :hearing_date, :multiparam_date
+
+      validates :hearing_court_name, presence: true
+      validates :hearing_date, presence: true,
+                multiparam_date: { allow_past: false, allow_future: true }
+
+      private
+
+      def persist!
+        kase.update(
+          attributes
+        )
+      end
+    end
+  end
+end
--- a/app/services/decisions/case_decision_tree.rb
+++ b/app/services/decisions/case_decision_tree.rb
@@ -1,6 +1,6 @@
 module Decisions
   class CaseDecisionTree < BaseDecisionTree
-    # rubocop:disable Metrics/MethodLength, Metrics/CyclomaticComplexity
+    # rubocop:disable Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/AbcSize
     def destination
       case step_name
       when :urn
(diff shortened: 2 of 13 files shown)


app/forms/steps/case/hearing_details_form.rb
```diff
@@ -0,0 +1,23 @@
...
+    class HearingDetailsForm < Steps::BaseFormObject
+      include Steps::HasOneAssociation
+      has_one_association :case
+
+      attribute :hearing_court_name, :string
+      attribute :hearing_date, :multiparam_date
+
+      validates :hearing_court_name, presence: true
+      validates :hearing_date, presence: true,
+                multiparam_date: { allow_past: false, allow_future: true }
```
willmcb, line 12: Wondering if there is any limit on the future? I guess more than a few years in the future would be too many? 🤔

app/services/decisions/case_decision_tree.rb
```diff
@@ -23,11 +23,14 @@ def destination
         after_delete_offence_date
       when :charges_summary
         after_charges_summary
+      when :hearing_details
+        # TODO: update when we have next step
+        show('/home', action: :index)
```
willmcb, line 28: 👍

willmcb (approved):
LGTM :shipit:

app/forms/steps/case/hearing_details_form.rb
zheileman: For now I think let's go with the defaults (2050 is the default). Don't think we will get a definite answer. The validation is mostly for fat finger typos.

timpeat: This is currently set to 2050 here: https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/blob/ed9c766a3fbe9c35eb1ad10738fb8a42b6016d6d/app/validators/multiparam_date_validator.rb#L4-L9

Perhaps we could also consider using age rather than fixed dates for this validation. I.e. 
```suggestion
             latest_years_from_now: 120
```

timpeat (approved):

app/forms/steps/case/hearing_details_form.rb
arthurashman: Agree with @timpeat here - something we should note down to improve in a future iteration

PR #120, opened 2022-09-29
Crimap 138 date stamp logic

willmcb:
## Description of change

Adds date stamp to the CrimeApplication after the CaseType step, and redirects user to see the "date stamp" page if one is added (its dependant on the case type), or moves them on to the "has codefendants" page if not.

## Link to relevant ticket
[CRIMAP-138](https://dsdmoj.atlassian.net/browse/CRIMAP-138)

## Notes for reviewer

- what are your thoughts on the `DateStamper`? I wanted to encapsulate any logic here in a re-usable component. So we thought it was better as a service than as a method in the `CaseDecisionTree`

### After changes (added date stamp to task list):
Screenshot 2022-09-29 at 11 28 34

- create a new application, take it through to case type, if you select: "Summary", "Either way", "committal", "Crown Court Appeal" you should see the date stamp page - if you select any other type it should go on to "has co-defendants" page.

### DIFF
--- a/app/forms/steps/case/case_type_form.rb
+++ b/app/forms/steps/case/case_type_form.rb
@@ -16,6 +16,7 @@ module Steps
                 presence: true,
                 if: -> { case_type&.cc_appeal_fin_change? }
 
+
       def choices
         CaseType.values
       end
@@ -23,6 +24,7 @@ module Steps
       private
 
       def persist!
+        DateStamper.new(crime_application, case_type).call
         kase.update(
           attributes.merge(attributes_to_reset)
         )
--- a/app/forms/steps/case/case_type_form.rb
+++ b/app/forms/steps/case/case_type_form.rb
@@ -16,7 +16,6 @@ module Steps
                 presence: true,
                 if: -> { case_type&.cc_appeal_fin_change? }
 
-
       def choices
         CaseType.values
       end
--- a/app/services/date_stamper.rb
+++ b/app/services/date_stamper.rb
@@ -1,17 +1,26 @@
 class DateStamper
+  DATE_STAMPABLE = [
+    :summary_only,
+    :either_way,
+    :committal,
+    :cc_appeal
+  ].freeze
+
   def initialize(crime_app, case_type)
     @crime_app = crime_app
     @case_type = case_type
   end
 
   def call
-    if @case_type.date_stampable? && @crime_app.date_stamp.nil?
+    # TODO: what happens when we already have a date?
+    if date_stampable?
       @crime_app.update(date_stamp: DateTime.now)
-    elsif !@case_type.date_stampable? && @crime_app.date_stamp.present?
-      @crime_app.update(date_stamp: nil)
-      false
-    else
-      false
     end
+  end
+
+  private
+
+  def date_stampable?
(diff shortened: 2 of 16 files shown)


app/services/decisions/case_decision_tree.rb
```diff
@@ -6,7 +6,7 @@ def destination
       when :urn
         edit(:case_type)
       when :case_type
-        edit(:has_codefendants)
+        after_case_type
```
timpeat, line 9: Could this not be handled by #update ?

app/value_objects/case_type.rb
```diff
@@ -8,4 +8,14 @@ class CaseType < ValueObject
     CC_APPEAL = new(:cc_appeal),
     CC_APPEAL_FIN_CHANGE = new(:cc_appeal_fin_change)
   ].freeze
+
+  def date_stampable?
+    date_stampable = [
```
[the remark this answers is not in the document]
willmcb: Yep that's a good point will add this in.

app/services/date_stamper.rb
```diff
@@ -0,0 +1,17 @@
+class DateStamper
+  def initialize(crime_app, case_type)
+    @crime_app = crime_app
+    @case_type = case_type
+  end
+
+  def call
```
timpeat, line 7: Could we also have an active record validation on date_stamp presence when case type is date_stampable? ?

app/forms/steps/case/case_type_form.rb
```diff
@@ -23,6 +23,7 @@ def choices
       private
 
       def persist!
+        DateStamper.new(crime_application, case_type).call
```
timpeat, line 26: @willmcb I forgot about making date stamping conditional on the update.

app/views/crime_applications/edit.html.erb
```diff
@@ -48,6 +48,15 @@
         <p class="govuk-body">
           <%= @crime_application.applicant_dob %>
         </p>
+
+        <% if @crime_application.date_stamp.present? %>
```
timpeat, line 52: @willmcb and we also need to implement new logic relating to not showing date stamp if set but case type not stampable .

willmcb: Ah of course! 🤦

app/forms/steps/case/case_type_form.rb
timpeat: Agreed to look at this in a separate commit.

kennyevil (approved):

spec/presenters/crime_application_presenter_spec.rb
```diff
@@ -42,6 +43,10 @@
       expect(subject.applicant_dob).to eq('1 Feb 1990')
     end
 
+    it 'can output the applicant date stamp in the correct format' do
+      expect(subject.pretty_date_stamp).to eq('1 Feb 2022')
```
kennyevil, line 47: Think this needs to change

kennyevil (changes requested):
Just some specs that need to be fixed.

spec/presenters/crime_application_presenter_spec.rb
willmcb: Thanks have fixed this now :)

timpeat (approved):

kennyevil (approved):

PR #122, opened 2022-10-03
Use rubocop for specs

timpeat:
Add rubocop-rspec and use it to check specs.

Prior to this change rubocop was not being used on the specs dir. 

This change: 
1. apply rubocop to the spec directory
2. fix all issues raised by rubocop
3. add rubocop-rspec
4. fix several issues raised by rubocop-rspec
5. disable or modify remaining failing rubocop-rspec cops

Are we happy to merge this as is or would we rather fix outstanding (disabled) rubcop-rspec cops?

### DIFF
--- a/spec/controllers/application_controller_spec.rb
+++ b/spec/controllers/application_controller_spec.rb
@@ -2,14 +2,14 @@ require 'rails_helper'
 
 RSpec.describe ApplicationController do
   controller do
-    def invalid_session; raise Errors::InvalidSession; end
-    def application_not_found; raise Errors::ApplicationNotFound; end
-    def another_exception; raise Exception; end
+    def invalid_session = raise(Errors::InvalidSession)
+    def application_not_found = raise(Errors::ApplicationNotFound)
+    def another_exception = raise(StandardError)
   end
 
   context 'Exceptions handling' do
     context 'Errors::InvalidSession' do
-      it 'should not report the exception, and redirect to the error page' do
+      it 'does not report the exception, and redirect to the error page' do
         routes.draw { get 'invalid_session' => 'anonymous#invalid_session' }
 
         expect(Sentry).not_to receive(:capture_exception)
@@ -20,7 +20,7 @@ RSpec.describe ApplicationController do
     end
 
     context 'Errors::ApplicationNotFound' do
-      it 'should not report the exception, and redirect to the error page' do
+      it 'does not report the exception, and redirect to the error page' do
         routes.draw { get 'application_not_found' => 'anonymous#application_not_found' }
 
         expect(Sentry).not_to receive(:capture_exception)
@@ -35,7 +35,7 @@ RSpec.describe ApplicationController do
         allow(Rails.application.config).to receive(:consider_all_requests_local).and_return(false)
       end
 
-      it 'should report the exception, and redirect to the error page' do
+      it 'reports the exception, and redirect to the error page' do
         routes.draw { get 'another_exception' => 'anonymous#another_exception' }
 
         expect(Sentry).to receive(:capture_exception)
--- a/spec/controllers/steps/case/charges_controller_spec.rb
+++ b/spec/controllers/steps/case/charges_controller_spec.rb
@@ -8,55 +8,60 @@ RSpec.describe Steps::Case::ChargesController, type: :controller do
 
   describe 'offence date actions' do
     let(:form_class_params_name) { form_class.name.underscore }
-    let(:existing_case) { Case.create(crime_application: crime_application)}
-    let(:charge_record) { Charge.create(case: existing_case)}
-    context 'when adding a date' do
-      it 'has the expected step name' do
-        expect(
-          subject
-        ).to receive(:update_and_advance).with(
-          form_class, 
-          record: charge_record, 
-          as: :add_offence_date
-        )
+    let(:existing_case) { Case.create(crime_application:) }
+    let(:charge_record) { Charge.create(case: existing_case) }
 
-        put :update, params: { 
-          id: crime_application.id, 
-          charge_id: charge_record.id, 
-          add_offence_date: '' 
+    context 'when adding a date' do
+      after do
+        put :update, params: {
+          id: crime_application.id,
+          charge_id: charge_record.id,
+          add_offence_date: ''
         }
       end
-    end
-
-    context 'when deleting a date' do
-      let(:offence_dates_attributes) { 
-        {
-          offence_dates_attributes: {
-            '0'=>{ 
-              'date(3i)'=>'10', 
-              'date(2i)'=>'10', 
-              'date(1i)'=>'2000', 
-              '_destroy'=>'1', 
-              'id'=> '123' 
-            }
-          }
-        }
-      }
 
       it 'has the expected step name' do
         expect(
           subject
         ).to receive(:update_and_advance).with(
           form_class,
-          record: charge_record, 
+          record: charge_record,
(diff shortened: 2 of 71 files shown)


spec/controllers/application_controller_spec.rb
```diff
@@ -2,14 +2,14 @@
 
 RSpec.describe ApplicationController do
   controller do
-    def invalid_session; raise Errors::InvalidSession; end
-    def application_not_found; raise Errors::ApplicationNotFound; end
-    def another_exception; raise Exception; end
+    def invalid_session = raise(Errors::InvalidSession)
```
kennyevil, line 5: learning something new every day. I didn't realise this was valid syntax.

spec/controllers/steps/case/charges_controller_spec.rb
```diff
@@ -8,55 +8,60 @@
 
   describe 'offence date actions' do
     let(:form_class_params_name) { form_class.name.underscore }
-    let(:existing_case) { Case.create(crime_application: crime_application)}
-    let(:charge_record) { Charge.create(case: existing_case)}
+    let(:existing_case) { Case.create(crime_application:) }
```
kennyevil, line 11: And this will definitely be useful too

spec/forms/steps/case/codefendants_form_spec.rb
```diff
@@ -1,22 +1,24 @@
...
 
   let(:crime_application) { instance_double(CrimeApplication, case: case_record) }
   let(:case_record) { Case.new }
 
-  let(:codefendants_attributes) {
+  let(:codefendants_attributes) do
     {
-      "0"=>{"first_name"=>"John", "last_name"=>"Doe", "conflict_of_interest"=>"no"},
-      "1"=>{"first_name"=>"Jane", "last_name"=>"Doe", "conflict_of_interest"=>"yes"},
+      '0' => { 'first_name' => 'John', 'last_name' => 'Doe', 'conflict_of_interest' => 'no' },
```
kennyevil, line 18: I think this change might have been caused by the cop that prefers single quotes for strings without interpolation on the basis that the single quoted string is more performant. I think it's been shown that there might not be a difference (eg this article https://www.viget.com/articles/just-use-double-quoted-ruby-strings/) and, if there is, it's pretty negligible, so I think we could safely turn that cop off.

kennyevil (approved):
This all looks great, just had one comment about one of the cops that might be superfluous, but it's more of a weakly held opinion than anything else, so shouldn't be a blocker. I've already learned a couple of useful things from the changes that have gone in.

spec/forms/steps/case/charges_form_spec.rb
```diff
@@ -53,39 +60,42 @@
...
         let(:case_record) { Case.create(crime_application: application) }
 
-        let(:offence_dates) { [
-          OffenceDate.new(date: Date.new(2000,11,03)),
-          OffenceDate.new(date: Date.new(2009,05,01))
-        ] }
+        let(:offence_dates) do
+          [
+            OffenceDate.new(date: Date.new(2000, 11, 0o3)),
+            OffenceDate.new(date: Date.new(2009, 0o5, 0o1))
```
arthurashman, line 66: Not sure I understand what these "o"s are about?

spec/forms/steps/case/codefendants_form_spec.rb
arthurashman: I'm personally happy to leave the cop on so that we end up with consistent quotes throughout

spec/forms/steps/client/contact_details_form_spec.rb
```diff
@@ -1,22 +1,23 @@
 require 'rails_helper'
 
 RSpec.describe Steps::Client::ContactDetailsForm do
-  let(:arguments) { {
-    crime_application: crime_application,
-    telephone_number: telephone_number,
-    correspondence_address_type: correspondence_address_type
-  } }
+  subject { described_class.new(arguments) }
```
arthurashman, line 4: I thought the cop was on to want named subjects? Does this want to be 'form'?

spec/presenters/crime_application_presenter_spec.rb
```diff
@@ -3,47 +3,50 @@
...
-  let(:applicant) { 
+  let(:applicant) do
     double(
-      Applicant, 
-      first_name: 'John', 
+      Applicant,
+      first_name: 'John',
       last_name: 'Doe',
-      date_of_birth: Date.new(1990, 02, 01),
+      date_of_birth: Date.new(1990, 0o2, 0o1)
```
arthurashman, line 21: Curious again why we're doing this?

spec/requests/charges_summary_spec.rb
```diff
@@ -8,7 +8,7 @@
 
     kase.charges.create!(
       offence_name: 'Robbery',
-      offence_dates_attributes: { id: nil, date: Date.new(1990, 02, 01) }
+      offence_dates_attributes: { id: nil, date: Date.new(1990, 0o2, 0o1) }
```
arthurashman, line 11: and here

spec/requests/dashboard_spec.rb
```diff
@@ -17,19 +17,22 @@
       # sets up a test record
       app = CrimeApplication.create
 
-      Applicant.create(crime_application: app, first_name: 'Jane', last_name: 'Doe', date_of_birth: Date.new(1990, 02, 01))
+      Applicant.create(crime_application: app, first_name: 'Jane', last_name: 'Doe',
+                       date_of_birth: Date.new(1990, 0o2, 0o1))
```
arthurashman, line 21: here

spec/services/date_stamper_spec.rb
```diff
@@ -1,17 +1,17 @@
 require 'rails_helper'
 
 RSpec.describe DateStamper do
+  subject { described_class.new(crime_app, case_type) }
+
   let(:case_type_sym) { CaseType::DATE_STAMPABLE.sample }
-  let(:date) { DateTime.new(2022, 02, 02) }
+  let(:date) { DateTime.new(2022, 0o2, 0o2) }
```
arthurashman, line 7: here

spec/support/matchers/validation_helpers.rb
```diff
@@ -2,11 +2,11 @@ module ValidationHelpers
   private
 
   def check_errors(object, attribute, error)
-    !object.valid?(validation_context)
+    object.valid?(validation_context)
```
arthurashman, line 5: Has this not changed what this line does altogether?

arthurashman (commented):
Thank you for trawling through all of that! 

I left a couple of questions. Confused by the inclusion of 'o's in dates, and the removal of a !

Also do you know why the naming a subject cop was enforced for some specs but not others?

spec/forms/steps/case/charges_form_spec.rb
timpeat: The leading zero seems confuse rubocop. I'll remove the leading zero for now rather than converting to base 8!

spec/support/matchers/validation_helpers.rb
timpeat: This method does not care what is returned by object.valid?.  object.valid? is being called because it runs the validations and adds any errors to the object.

timpeat:
> Also do you know why the naming a subject cop was enforced for some specs but not others?

Simply because I ran out of time part way through manually naming the subject.

I'd thought it better to get this code into main and then deal with the more time consuming cops as time permitted.

PR #125, opened 2022-10-10
Reorder steps to match MVP user flow

arthurashman:
## Description of change
Reordered steps in user flow to match user flow in Miro. 

Previously we were navigating from case_type to has_codefendant instead of the correct case_type to charges journey.
See [here](https://miro.com/app/board/uXjVOh963Us=/?moveToWidget=3458764530591549436&cot=10) for start of case journey to validate.

Also corrects date stamp to only show if user doesn't already have a date stamp.  

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-147

## Notes for reviewer
Check flow in local matches [Miro user flow](https://miro.com/app/board/uXjVOh963Us=/?moveToWidget=3458764530591549436&cot=10)

### Before changes:
image

### After changes:
image


app/controllers/steps/case/date_stamp_controller.rb
```diff
@@ -1,7 +1,9 @@
 module Steps
   module Case
     class DateStampController < Steps::CaseStepController
-      def show; end
+      def show
+        @charge = current_crime_application.case.charges.create!
```
kennyevil, line 5: Might be worth adding a comment about why you're putting this here, so we can decide later on about what to do with it later.

kennyevil (approved):

PR #126, opened 2022-10-11
Add benefit check service and mock benefit check service

kennyevil:
## Description of change

Adds a mock benefit service to check on the benefits status of an applicant. Also adds the BenefitCheckService, but this currently only calls the mock.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-107
https://dsdmoj.atlassian.net/browse/CRIMAP-133

## Screenshot

0 0 0 0_3000_applications_edfb48aa-54c9-4a5f-a1d3-bd8edf217625_steps_client_benefit_check_result

After the has_nino step, you should see a simple page with the result of the benefit check

### DIFF
--- a/app/services/mock_benefit_check_service.rb
+++ b/app/services/mock_benefit_check_service.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+class MockBenefitCheckService
+  KNOWN = {
+    'SMITH' => { nino: 'ZZ123459A', dob: '11-Jan-99' },
+    'JONES' => { nino: 'ZZ123458A', dob: '1-Jun-80' },
+    'BLOGGS' => { nino: 'ZZ123457A', dob: '4-Jan-90' },
+    'WRINKLE' => { nino: 'ZZ010150A', dob: '01-Jan-50' },
+    'WALKER' => { nino: 'JA293483A', dob: '10-Jan-80' }, # Used in cucumber tests and specs
+  }.freeze
+
+  def self.call(*args)
+    new(*args).call
+  end
+
+  attr_reader :crime_application
+
+  delegate :applicant, to: :crime_application, allow_nil: true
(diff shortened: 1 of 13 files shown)


app/services/mock_benefit_check_service.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
```
timpeat, line 1: We're using ruby 3 so this is not required.

timpeat (approved):
Looks good to me 👍  (As an aside, does the DWP service have a means to distinguish between not found and found but not on benefits.)

kennyevil:
@timpeat I don't believe it does, judging by the response in this cassette for the benefit checker service on Civil apply. 

https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/fb039be3be419dc8b4b46656ead8791dfd675855/spec/cassettes/benefit_check_service/successful_call.yml#L47

app/services/mock_benefit_check_service.rb
```diff
@@ -0,0 +1,47 @@
+class MockBenefitCheckService
+  KNOWN = {
+    'SMITH' => { nino: 'ZZ123459A', dob: '11-Jan-99' },
+    'JONES' => { nino: 'ZZ123458A', dob: '1-Jun-80' },
+    'BLOGGS' => { nino: 'ZZ123457A', dob: '4-Jan-90' },
+    'WRINKLE' => { nino: 'ZZ010150A', dob: '01-Jan-50' },
+    'WALKER' => { nino: 'JA293483A', dob: '10-Jan-80' }, # Used in cucumber tests and specs
```
arthurashman, line 7: ZZ is an invalid prefix, so won't pass our NINO validation. Can we change them to AB or something?

PR #127, opened 2022-10-11
update mock NINOs to use a valid format

kennyevil:
## Description of change

Updates the mock NINOs so they no longer use the invalid ZZ prefix.


timpeat (approved):
👍

PR #129, opened 2022-10-12
Replace hardcoded case type strings/symbols with constants

zheileman:
## Description of change
Ensuring we always use, as much as possible, the value object constants to avoid the situation where we rename or add/remove constants later on and we forget to update the tests or worse, some tests keep passing mysteriously.

## Notes for reviewer
There are no functional changes, this is a small code refactor, mainly specs.


kennyevil (approved):
LGTM

PR #130, opened 2022-10-12
Move DateStamper call to decision tree, only show date_stamp page once

arthurashman:
## Description of change

Only shows the user the date_stamp page on the first time a date_stampable case_type is chose and the form is submitted.
Modified the logic in the case_decision_tree to allow for this - called the DateStamper from the decision tree so it returns false if the date_stamp is not updated by the DateStamper on subsequent re submissions of case_type step.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-148

Go through user flow up to Case Type step
Select a date stampable Case Type (e.g. Summary Only)
Submit form and be redirected to Case Stamp page
Hit the back button to return to Case Type step 

Resubmit form/ change case type and resubmit and get redirected to charges journey

### DIFF
--- a/app/forms/steps/case/case_type_form.rb
+++ b/app/forms/steps/case/case_type_form.rb
@@ -23,14 +23,7 @@ module Steps
       private
 
       def persist!
-        # TODO: discuss wrapping in a transaction to prevent Case#update
-        # if DateStamper fails
-
-        return false unless kase.update(attributes.merge(attributes_to_reset))
-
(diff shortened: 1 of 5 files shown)


kennyevil (approved):
LGTM

zheileman (changes requested):
Just a comment for some code I believe was removed by mistake 😄

zheileman (approved):

PR #131, opened 2022-10-12
CRIMAP-152 Change date stamp step from show to edit

zheileman:
## Description of change
This looks like a lot of changes but in reality it is only changing the date stamp page from a `#show` to an `#edit` step, and thus introducing a form object that does not perform any particular operation on persist, but allow us to control the flow on the decision tree and thus initialise a charge to redirect to it, or show the charges summary page if we have already charges.

There is still an edge case where the provider could end up with an initialised charge, without name, if they hit the save draft button. I think we can sort this our in a separate PR.

As part of this, I made the `#continue_button` form builder helper to receive additional arguments to be able to customise the attributes of the buttons, in this case we needed to pass custom classes.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-152

The date stamp page will not create any charge record on load/reload. It is now a form that must be submitted and charges are created on the decision tree.


kennyevil (approved):
LGTM. Should work well with @arthurashman's work in #130

PR #132, opened 2022-10-13
CRIMAP-154 Reset correspondence address when not applicable

zheileman:
## Description of change
As we do in other places, if provider changes the correspondence type selection from `other` to anything else (like home or provider's office) we reset the previously entered correspondence address (if any).

As this is a `has_one` relationship, nilling the attribute will remove the record from the `Addresses` table.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-154

Go to contact details and select "other address", enter a correspondence address, then go back and change selection, the correspondence address record will be deleted.

### DIFF
--- a/app/forms/steps/client/contact_details_form.rb
+++ b/app/forms/steps/client/contact_details_form.rb
@@ -32,8 +32,16 @@ module Steps
 
       def persist!
         applicant.update(
-          attributes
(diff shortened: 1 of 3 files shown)


kennyevil (approved):
LGTM

PR #134, opened 2022-10-14
CRIMAP-163 Better handling of incomplete offences

zheileman:
## Description of change
Due to the way the offences page work (a page with nested attributes for details of the offence like offence name and one or more dates), each time the provider adds a new offence through the summary page, a blank record is created in the database, for the provider to fill in the details (offence name and dates).

This usually happens in one go, but if at that point, without finalising all the details, the provider clicks the “save and come back later“ button, for example, or click the “back“ link in the page, they will leave the offence in an incomplete state, maybe missing the name, or the dates, or both.

Removing these empty records is not a nice solution really, as some might contain partially entered details, like the offence name (but no dates) or vice versa, if the provider clicks the save draft button, in order to come back later on to these offences.

Also it breaks the navigation as clicking the "back" link in the summary page might go to a just deleted charge record.

A better solution is to present these incomplete charges for the provider to either remove them or edit and complete them.

Also some additional validation is added to not allow a provider to proceed with incomplete charges in the "basket" page.

Discussed the copy and design with service designer and we decided to go with the MOJ badges.

MOJ library is added to the package.json but only the badge component is used. In a separate PR I will also refactor the task list to use the MOJ import instead of maintaining our own copy of the component.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-163

### Before changes:
Screenshot 2022-10-14 at 09 29 18

### After changes:
Screenshot 2022-10-14 at 09 29 42

Screenshot 2022-10-14 at 09 30 09

Go to the offences page, enter an offence, proceed to the basket, then select add another offence YES, but don't enter details, click the back link, and see how it shows as incomplete.
Select add another offence NO and you will get a validation error as well.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,5 +1,7 @@
 // Components or full features
 @import "local/govuk";
+@import "local/moj";
+
 @import "local/info_card";
 @import "local/task_list";
 
--- a/app/assets/stylesheets/local/moj.scss
+++ b/app/assets/stylesheets/local/moj.scss
@@ -1,0 +1,9 @@
+// MOJ Design System
+// https://design-patterns.service.justice.gov.uk
+//
+// NOTE: using hand-picked components instead of importing
+// the whole library, as we only need a few ones.
+
+// Badge
+// https://design-patterns.service.justice.gov.uk/components/badge/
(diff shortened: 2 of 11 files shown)


arthurashman (approved):
This is really nice solution, nice one Jesus!

PR #135, opened 2022-10-14
CRIMAP-108 Add review page (CYA) skeleton

zheileman:
## Description of change
Add the necessary scaffolding for the review page (aka. check your answers). For now this is just an empty page, and in follow-up PRs the different sections will be added.

I've made this an edit step instead of a show step, even tho it does not have any form, in case we end up having a similar situation to the date stamp where we might need to run some logic or update some attribute on "continue" button. It is not a big deal, just an empty form object for now.

Also, this is namespaced as `submission`, with this step named `review`, as there will be some others like the `declaration`, `confirmation` and perhaps the `print` etc. that can be grouped together as a mini-journey to make it cleaner.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-108

## Notes for reviewer
This page is not yet integrated in the wider journey. Can be accessed manually knowing the URL.

Screenshot 2022-10-14 at 10 51 36

As this page is not yet part of the wider journey (will be added after we have IoJ step), you can access it manually by going to the `/applications/{uuid}/steps/submission/review` URL.


arthurashman (approved):
Perfect, LGTM

PR #136, opened 2022-10-14
CRIMAP-112 Add Interest of Justice page

arthurashman:
## Description of change
Adds ioj model and table - associated to case
Adds ioj controller and view

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-112
Designs: https://www.figma.com/file/thkvnDBbQbHlSqzt1TE3lk/Crime-Apply?node-id=2711%3A6495

image

Go through full flow
Check validation error by submitting without selecting an ioj
Reload page, check validation error for submitting with selected ioj but no description
Reload page, select ioj, enter text in text area, submit, check record is correct in console


config/locales/en/errors.yml
```diff
@@ -147,3 +147,27 @@ en:
               invalid_year: Enter a valid year
               year_too_late: Date of next hearing is too far in the future
               past_not_allowed: Date of next hearing must be in the future
+        steps/case/ioj_form:
+          attributes:
+            types:
+              invalid: Select a criteria
+            loss_of_liberty_justification:
+              blank: Enter more detail
```
zheileman, line 155: I know is a pain but I think this may fail stringent accessibility audits, because if there are more than one text area erroring, all will have the same error message read aloud to a screen reader in the summary.

Perhaps, "Enter justification for [xyz]"? i.e. "Enter justification for loss of liberty justification".

Maybe worth checking with @carolinapizatto

spec/value_objects/ioj_reason_type_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rails_helper'
+
+RSpec.describe IojReasonType do
```
zheileman, line 3: Super tiny thing but might be worth having a little test for the `#justification_field_name` method, to cover all our bases 😉

zheileman (approved):
Great!!! 🥳 
I'm approving. I've left some minor comments or suggestions, no blockers really.

config/locales/en/errors.yml
arthurashman: Good catch, that sounds sensible

app/models/crime_application.rb
```diff
@@ -3,6 +3,7 @@ class CrimeApplication < ApplicationRecord
 
   has_one :applicant, dependent: :destroy
   has_one :partner, dependent: :destroy
+  has_one :ioj, through: :case, dependent: :destroy
```
zheileman, line 6: I think we can get rid of this, so if we ever feel inclined to use this as a shorthand instead of going through `case`, it fails and we can consider our options.

PR #138, opened 2022-10-17
Import MOJ task-list component

zheileman:
## Description of change
As a follow-up to PR #134 where we introduced the MOJ design system in order to use the badges, now we can also replace the task-list styles with the MOJ component.

This is desirable so we get updates and fixes whenever they update the component and we pull a new version.

A few tweaks were needed however in order to render the tags properly but I've kept these as minimal and with separate class names to not clash with the MOJ ones.

After pulling the new code do not forget to compile assets. And also, if you didn't before, run `yarn` to pull the MOJ design system npm module.


timpeat (approved):
LGTM

PR #141, opened 2022-10-18
CRIMAP-156 Review case details

zheileman:
## Description of change
Add the case details section to the check your answers page. This only has 2 rows, one of them (URN) might be blank but we show them regardless so the provider can get back and add a reference if they need to.

There is some overlapping code between this PR and a previous one still open. Will fix conflicts before merging.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-156

Screenshot 2022-10-18 at 12 51 25


timpeat (approved):
Looks Great

PR #143, opened 2022-10-19
CRIMAP-47 Redirect non passported

kennyevil:
## Description of change

Adds DWP passporting check journey

## Link to relevant ticket

https://dsdmoj.atlassian.net/browse/CRIMAP-47
https://dsdmoj.atlassian.net/browse/CRIMAP-153

0 0 0 0_3000_applications_47e125c4-9a1c-4542-9298-f9cad2769807_steps_client_benefit_check_result

0 0 0 0_3000_applications_6cbe1bc6-381e-465c-b458-5f46d7fe0649_steps_dwp_confirm_result

0 0 0 0_3000_applications_47e125c4-9a1c-4542-9298-f9cad2769807_steps_client_confirm_details

### Before changes:

The DWP check just provided a yes or no

### After changes:

The journey now has the ability to agree or disagree with the result, and to confirm if client details are correct

After the National Insurance number page, you should be presented with the results of the DWP check and the onward journey. For NINOs that provide a confirmation of passporting benefits, see here https://dsdmoj.atlassian.net/wiki/spaces/CRIMAP/pages/4186800190/List+of+known+NINOs+for+the+MockBenefitCheckService

Otherwise, enter any valid NINO and follow the other path


app/forms/steps/client/benefit_check_result_form.rb
```diff
@@ -2,17 +2,31 @@ module Steps
   module Client
     class BenefitCheckResultForm < Steps::BaseFormObject
       include Steps::HasOneAssociation
+      attribute :passporting_benefit, :value_object, source: YesNoAnswer
```
zheileman, line 5: I think the validation is missing in this form, meaning a provider can continue without selecting any radio option.

app/views/steps/client/benefit_check_result/edit.html.erb
```diff
@@ -1,12 +1,18 @@
...
-  <div class="govuk-grid-column-two-thirds">
-    <p>Benefit check result: <%= @form_object.benefit_check_result[:benefit_checker_status] %></p>
+<div class="govuk-grid-row app-inverted--card app-inverted-text">
+  <div class="govuk-grid-column app-inverted-text">
+
     <%= govuk_error_summary(@form_object) %>
     <%= step_form @form_object do |f| %>
-      <%= f.continue_button %>
+      <% if @form_object.benefit_check_result[:benefit_checker_status] == 'Yes' %>
+        <%= render 'passported' %>
```
zheileman, line 10: I find the use of partials here a bit confusing. It hinders the readability IMO.

I presume it is so the different results (passported / non-passported) can have different titles, and one is coming from the `govuk_radio_buttons_fieldset` whereas the other (when no radio fieldset) is coming from an explicit `h1`?

I'm thinking perhaps it may be worth exploring, even if more verbose, branching out into 2 separate routes/controllers/formobjects, one for passported, another for non-passported?
And we decide which one to go to in the decision tree?

app/controllers/steps/client/benefit_check_result_controller.rb
```diff
@@ -10,6 +10,10 @@ def edit
       def update
         update_and_advance(BenefitCheckResultForm, as: :benefit_check_result)
       end
+
+      def additional_permitted_params
+        %i[confirm_benefit_check_result]
```
zheileman, line 15: I'm confused about this. Should not be necessary if the attributes are declared in the form object.

app/forms/steps/client/benefit_check_result_form.rb
zheileman, line 5: Wouldn't this be the other way around? the yes-no attribute is the one used in the radio options shown in the view.
The other attribute is something internal is it? Maybe I'm missing something here.

zheileman (changes requested):
Hey looking great!
I've added a few comments, but have to leave now. We can continue tomorrow, I think it may be best to jump into a call at some point to clarify some things 😄

app/views/steps/client/benefit_check_result/edit.html.erb
kennyevil: I initially thought that the differences were going to be more substantial than they were. I can move the content from the partials onto the edit page and do away with them. I don't think the difference between the two is substantial enough to warrant two separate routes/controller/formobjects though.

app/forms/steps/client/benefit_check_result_form.rb
kennyevil: You're not missing anything. My initial instinct was to declare the passporting_benefit as a boolean, but I noticed the ```YesNoAnswer``` was being used elsewhere and thought I should use that, especially as the benefit check result is either a 'yes' or 'no'

zheileman: Perhaps if this is to store the result of the benefit checker, a boolean might be better? This is assuming the result would be an absolute pass/fail. If we think we could have different nuanced results, like a success, a fail, and a "maybe" (just an example) then a string is best so we can store the result returned from the benefit checker in our DB and be able to propagate that to the case worker if they need it, or to perform different logic?

kennyevil: I'll change it to a boolean for now, and if it turns to be more nuanced, we can change it later.

app/views/steps/dwp/confirm_result/edit.html.erb
```diff
@@ -0,0 +1,23 @@
...
+
+<div class="govuk-grid-row app-inverted--card app-inverted-text">
+  <div class="govuk-grid-column app-inverted-text">
+    <h1 class="govuk-heading-l app-inverted-text"><%= t '.heading' %></h1>
+  </div>
+</div>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= step_form @form_object do |f| %>
+      <%= f.govuk_radio_buttons_fieldset(:confirm_result, legend: { tag: 'h1', size: 'l' }) do %>
```
zheileman, line 14: Let's use `govuk_collection_radio_buttons` (like in the other view `dwp/confirm_details/edit.html.erb`) so we don't have to do a loop on the choices.

Additionally this page has two `h1`. If you use the same code as in the other view this will also be fixed I believe.

app/views/steps/dwp/confirm_result/edit.html.erb
```diff
@@ -0,0 +1,23 @@
+<% title t('.page_title') %>
+<% step_header %>
+
+  <%= govuk_error_summary(@form_object) %>
```
zheileman, line 4: We will need a wrapper for the error summary:

```erb
<div class="govuk-grid-row">
   </div>
</div>
```

config/locales/en/steps.yml
```diff
@@ -18,7 +18,9 @@ en:
           page_title: Enter your client’s NINO
           eforms_link: My client does not have a National Insurance number or cannot find it
       benefit_check_result:
+        title: DWP records show that your client does not receive a passporting benefit - is this correct?
```
zheileman, line 21: I think this locale is not used anywhere now?

config/routes.rb
```diff
@@ -51,11 +51,19 @@ def show_step(name)
         edit_step :details
         edit_step :has_nino
         edit_step :benefit_check_result
+        edit_step :confirm_details
```
zheileman, line 54: Do we use these 2 routes?

spec/forms/steps/dwp/confirm_result_form_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  subject { described_class.new(arguments) }
+
+  let(:arguments) do
+    {
+      crime_application:,
+    }
+  end
+
+  let(:last_name) { 'Smith' }
```
zheileman, line 15: Maybe all these applicant details are not needed for this form object?

zheileman (changes requested):
I added a few comments and suggestions!
Looking great, functionally is working as expected, as I tested it locally 👍

Give me a shout if you need any help or clarification!

app/forms/steps/dwp/confirm_result_form.rb
```diff
@@ -0,0 +1,24 @@
...
+      def choices
+        YesNoAnswer.values
+      end
+
+      private
+
+      def persist!
+        return true unless confirm_result.no?
+
+        applicant.update(passporting_benefit: nil)
```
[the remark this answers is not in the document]
kennyevil: Good point. I think I was trying to preserve the true, false, nil behaviour which no longer applies.

app/views/steps/dwp/confirm_details/edit.html.erb
```diff
@@ -0,0 +1,65 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full govuk-!-margin-bottom-5">
+    <h1 class="govuk-heading-xl"><%=t '.heading' %></h1>
+  </div>
+</div>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <table class="govuk-table app-dashboard-table">
```
zheileman, line 18: Sorry I just realised this is a table.

Tables are not very responsive in small devices, and for presenting pairs of key-value like this, we use instead definition lists ([summary list](https://design-system.service.gov.uk/components/summary-list/)). In this case, without the change links. There is an example in that page that is almost what we want here.

Let me know if you need help with this!

app/views/steps/dwp/confirm_details/edit.html.erb
```diff
@@ -0,0 +1,65 @@
...
+            <%= @form_object.applicant.nino %>
+          </td>
+        </tr>
+      </tbody>
+    </table>
+  </div>
+</div>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
```
zheileman, line 58:
```suggestion
  <div class="govuk-grid-column-two-thirds">
```

app/views/steps/dwp/confirm_details/edit.html.erb
```diff
@@ -0,0 +1,65 @@
...
+<% step_header %>
+
+<div class="govuk-grid-row">
+   <div class="govuk-grid-column-two-thirds">
+      <%= govuk_error_summary(@form_object) %>
+   </div>
+</div>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full govuk-!-margin-bottom-5">
```
zheileman, line 11:
```suggestion
  <div class="govuk-grid-column-two-thirds govuk-!-margin-bottom-5">
```

spec/forms/steps/client/has_nino_form_spec.rb
```diff
@@ -13,12 +13,14 @@
...
   end
 
-  let(:crime_application) do
-    instance_double(CrimeApplication)
-  end
+  let(:crime_application) { instance_double(CrimeApplication, applicant: 'applicant') }
 
   let(:nino) { nil }
 
+  before do
```
zheileman, line 20: We can remove this now, as we don't have this call on the NINO form object anymore 😄

zheileman (approved):

PR #151, opened 2022-10-21
CRIMAP-161 Review justification for legal

arthurashman:
## Description of change
Add the "justification for legal aid" section to the check your answers page.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-161

### After changes:

image


app/presenters/summary/sections/justification_for_legal_aid.rb
```diff
@@ -0,0 +1,36 @@
+module Summary
+  module Sections
+    class JustificationForLegalAid < Sections::BaseSection
+      def name
+        :justification_for_legal_aid
+      end
+
+      def show?
+        kase.present? && super
```
zheileman, line 9: We will also need to protect against a potential nil `ioj`, just in case.

app/presenters/summary/sections/justification_for_legal_aid.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+
+      def answers
+        [
+          ioj.types.map do |ioj_type|
+            justification = "#{ioj_type}_justification"
```
zheileman, line 15: It is best if we keep using the `#justification_field_name` method. We can instantiate the value-object with `IojReasonType.new(ioj_type)` and then call the method.

app/presenters/summary/sections/justification_for_legal_aid.rb
```diff
@@ -0,0 +1,36 @@
...
+            Components::FreeTextAnswer.new(
+              :ioj_justification, ioj[justification],
```
zheileman, line 17: Here instead of `:ioj_justification` it would just be the type. Or if using the value-object, the `#value` (which is a symbol). Then in the locales we have to declare a key for each of the IoJ types.
Because `ioj_type.humanize` is not really doing localisation.

app/presenters/summary/sections/justification_for_legal_aid.rb
```diff
@@ -0,0 +1,36 @@
...
+              change_path: edit_steps_case_ioj_path(anchor: justification),
```
zheileman, line 18: I don't think this anchor is working. We would need to add the IDs to the view.

config/locales/en/summary.yml
```diff
@@ -89,3 +90,8 @@ en:
       hearing_date:
         question: Date of next hearing
       # END next court hearing section
+
+      # BEGIN justification for legal aid section
+      ioj_justification:
```
zheileman, line 95: So, here, in order to do proper localisation we will need to place all the IoJ types in the following format:

```yaml
loss_of_liberty:
  question: Loss of liberty
suspended_sentence:
  question: Suspended sentence
[...]
```

Is a bit verbose but it is the most flexible.

spec/presenters/summary/sections/justification_for_legal_aid_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  describe '#answers' do
+    let(:answers) { subject.answers }
+
+    it 'has the correct rows' do
+      expect(answers.count).to eq(2)
+
+      expect(answers[0]).to be_an_instance_of(Summary::Components::FreeTextAnswer)
+      expect(answers[0].question).to eq(:ioj_justification)
+      expect(answers[0].change_path).to match('applications/12345/steps/case/ioj')
```
zheileman, line 62: Paths are missing the anchors.

spec/presenters/summary/sections/justification_for_legal_aid_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  let(:ioj) do
+    Ioj.new(
+      types: %w[reputation other],
+      reputation_justification: 'A justification',
+      other_justification: 'Another justification'
+    )
+  end
+
+  before do
+    allow(ioj).to receive('reputation_justification').and_return('A justification')
```
zheileman, line 30: I'm not sure you need these allows? Can you try without them?

zheileman (changes requested):
Very nice. I added a few comments and suggestions!
Ping me if you have any issue!

app/presenters/summary/sections/justification_for_legal_aid.rb
arthurashman: Why is that needed here but not in https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/blob/main/app/presenters/summary/sections/next_court_hearing.rb

zheileman: Because on the NextCourtHearing we only access "kase" (so we protect against it being `nil`).

But in the IoJ we have a 2-level relationship where to reach `ioj` you need to go through `case`. Case could be present but `ioj` be nil and then in the loop the code will blow up 😝

arthurashman: Gotcha, ok, thank you!

app/presenters/summary/sections/justification_for_legal_aid.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+
+      def answers
+        [
+          ioj.types.map do |ioj_type|
+            type = IojReasonType.new(ioj_type)
+            Components::FreeTextAnswer.new(
+              type, ioj[type.justification_field_name],
+              change_path: edit_steps_case_ioj_path(anchor: type),
+              i18n_opts: { ioj_type: ioj_type.humanize }
```
zheileman, line 19: Tiny thing, no need for the `i18n_opts` anymore.

spec/presenters/summary/html_presenter_spec.rb
```diff
@@ -4,11 +4,16 @@
...
+    instance_double(CrimeApplication, applicant: double, case: kase)
   end
 
+  let(:kase) { instance_double(Case, ioj:) }
+
+  let(:ioj) { instance_double(Ioj) }
+
   describe '#sections' do
     before do
+      allow(kase).to receive(:ioj).and_return(ioj)
```
zheileman, line 16: Another tiny thing, no need for this allow.

spec/presenters/summary/sections/justification_for_legal_aid_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      other_justification: 'Another justification'
+    )
+  end
+
+  describe '#name' do
+    it { expect(subject.name).to eq(:justification_for_legal_aid) }
+  end
+
+  describe '#show?' do
+    context 'when there is a case' do
```
zheileman, line 34: No blocker, but would be good to, to add inside this context (when there is a case), a couple expectations to test when there is ioj and where there isn't.

app/presenters/summary/sections/justification_for_legal_aid.rb
arthurashman: 👍

spec/presenters/summary/html_presenter_spec.rb
arthurashman: 👍

PR #157, opened 2022-10-25
CRIMAP-170 View completed application

zheileman:
## Description of change
This is what in the prototype is called the application certificate and is a read-only view of the "review / check your answers" page.

The print button works but is not yet applying the print stylesheet, that will come as part of a separate ticket.

Also, for now I've had to make a lot of assumptions to get to this point, mainly, the "submission" which for not is no more than a mock service `ApplicationSubmission` that sets the `submitted` status in the record.

I've also added a new `submitted_at` DB field because it will be needed for some date stamp logic soon.

In separate PRs, we will introduce checks to make sure a "submitted" application can't be edited, or for what is worth, can't be re-submitted (which at the moment is possible as there is no such validation/check yet).

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-170
https://www.figma.com/file/thkvnDBbQbHlSqzt1TE3lk/Crime-Apply?node-id=2711%3A6326

## Notes for reviewer
There are many little details to sort out in separate PRs. I think is best to wait for the redesign of the dashboard to have sub-navigation to implement some of these details (like no delete buttons on submitted applications, or accessing this certificate page from the dashboard list).

Screenshot 2022-10-25 at 13 06 00

Go all the way down to the declaration, tick the checkbox and submit the application. Then in the confirmation page click the button "View completed application".

### DIFF
--- a/app/controllers/crime_applications_controller.rb
+++ b/app/controllers/crime_applications_controller.rb
@@ -11,6 +11,12 @@ class CrimeApplicationsController < ApplicationController
                     .merge(CrimeApplication.order(created_at: :desc))
   end
 
+  def show
+    @presenter = Summary::HtmlPresenter.new(
+      crime_application: current_crime_application
+    )
+  end
+
   def create
     initialize_crime_application do |crime_application|
       redirect_to edit_steps_client_has_partner_path(crime_application)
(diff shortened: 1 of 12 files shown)


kennyevil (approved):
LGTM

PR #158, opened 2022-10-25
CRIMAP-168 Add about controller and privacy page

arthurashman:
## Description of change
Add privacy policy page with content from civil apply (to be reviewed in https://dsdmoj.atlassian.net/browse/CRIMAP-150) 

Minor changes to copy where looked to be incorrect in civil. Adding full stop, removing colon from a h2 heading to make consistent. Adding missing into line to an address that is in the civil yml but unused. 

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-168

## Notes for reviewer
To match https://staging.apply-for-legal-aid.service.justice.gov.uk/privacy_policy?locale=en

### DIFF
--- a/app/views/about/privacy.en.html.erb
+++ b/app/views/about/privacy.en.html.erb
@@ -1,150 +1,155 @@
 <% title t('.page_title') %>
-  <h1 class="govuk-heading-xl">Privacy policy</h2>
-  <h2 class="govuk-heading-l">Purpose</h2>
-  <p class="govuk-body">This privacy notice sets out:</p>
-  <div class="govuk-list">
-    <ul class="govuk-list govuk-list--bullet">
-      <li>the standards that you can expect from the Legal Aid Agency (LAA) when we request or hold personal data about you</li>
-      <li>how you can get access to a copy of your personal data</li>
-      <li>what you can do if you think the standards are not being met</li>
-    </ul>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">Privacy policy</h2>
+    <h2 class="govuk-heading-l">Purpose</h2>
+    <p class="govuk-body">This privacy notice sets out:</p>
+    <div class="govuk-list">
+      <ul class="govuk-list govuk-list--bullet">
+        <li>the standards that you can expect from the Legal Aid Agency (LAA) when we request or hold personal data about you</li>
+        <li>how you can get access to a copy of your personal data</li>
+        <li>what you can do if you think the standards are not being met</li>
+      </ul>
+    </div>
+    <p class="govuk-body">The LAA is an Executive Agency of the Ministry of Justice (MoJ). The MoJ is the data controller for the personal information we hold.</p>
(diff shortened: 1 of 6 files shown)


app/views/about/privacy.en.html.erb
```diff
@@ -0,0 +1,150 @@
+<% title t('.page_title') %>
```
zheileman, line 1: If you use the hardcoded 'Privacy policy' string (as the whole page is hardcoded), you don't even need the locales file. But is fine either way 👍

app/views/about/privacy.en.html.erb
```diff
@@ -0,0 +1,150 @@
...
+  <h2 class="govuk-heading-l">Purpose</h2>
+  <p class="govuk-body">This privacy notice sets out:</p>
+  <div class="govuk-list">
+    <ul class="govuk-list govuk-list--bullet">
+      <li>the standards that you can expect from the Legal Aid Agency (LAA) when we request or hold personal data about you</li>
+      <li>how you can get access to a copy of your personal data</li>
+      <li>what you can do if you think the standards are not being met</li>
+    </ul>
+  </div>
+  <p class="govuk-body">The LAA is an Executive Agency of the Ministry of Justice (MoJ). The MoJ is the data controller for the personal information we hold.</p>
```
zheileman, line 12: I think there is an extra 2 or 3 leading spaces in all lines?

app/views/about/privacy.en.html.erb
```diff
@@ -0,0 +1,150 @@
+<% title t('.page_title') %>
+  <h1 class="govuk-heading-xl">Privacy policy</h2>
```
zheileman, line 2: It needs the wrapper.

```html
<div class="govuk-grid-row">
     [content here]
  </div>
</div>
```

config/routes.rb
```diff
@@ -40,6 +40,10 @@ def show_step(name)
     end
   end
 
+  resource :about, only: [:show] do
```
zheileman, line 43: This exposes an additional route `abouts#show`.
Also named routes become `privacy_about_path`.

I suggest the simpler form, because about controller is not resourceful:

`get 'about/privacy'`

This gets you a nicer named route `about_privacy_path` (and the same when we have other about pages).

zheileman (commented):
Cool. I added a few comments, nothing major!

Do you mind also linking to the privacy policy in the declaration page, here? At the moment is just a placeholder.

`app/views/steps/submission/declaration/edit.en.html.erb`

app/views/about/privacy.en.html.erb
arthurashman: I've left this as is, but primarily because I couldn't get it to work -very happy to change it if appropriate

zheileman: I think this would work:

```erb
<% title 'My page title' %>
```

But it is fine either way.

zheileman (approved):

app/views/about/privacy.en.html.erb
arthurashman: I'll switch it. Better to have less code!

PR #161, opened 2022-10-27
CRIMAP-174 Refactor to enable post-submission

zheileman:
## Description of change
This is a bit of a refactor to more clearly separate the boundaries between "local" (and thus `in_progress`) applications, and "remote" (or submitted) applications, those that eventually will reside in the document store.

The first challenge is identifying where in the service we are using a "current_crime_application" record, because at the moment all come from the local DB.

This will not be the case in the future when we have the document store.

Some pages can't continue referring to this method expecting an application is in the local DB. For example the submission confirmation is a funny one because right on submission we will have marked as "submitted" the application and eventually purged it from database, so this page can't rely on a record on the DB.

Fortunately this page renders very little, just a reference, and the UUID (used in one button) remains on the route URL, so passing the reference as a query param, allows us to not rely on a record, and to also make this page bookmarkable as recommend by the design guidelines: https://design-system.service.gov.uk/patterns/confirmation-pages/

> Some users will bookmark the confirmation page as a form of receipt. You should allow them to return to the page, whenever possible.

The other post-submission page we have currently is the "view certificate". This has been moved to its own controller where maybe more actions in the future will be added. The `current_crime_application` then scopes the DB query to `submitted` only, but once we introduce the document store, all these ActiveRecord queries need to be transformed into an API client querying the document store.

Finally, if the application in our local DB is marked as `submitted` (in the future, once these are purged, they will not even exist in local DB) and the provider tries to access any route making use of the record (any steps page for example, or task list, etc.) they will be given an error, as per discussion in the ticket.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-174

## Notes for reviewer
Many assumptions still, until we really have a document store. We keep working with local DB records, even for submitted applications, when that will not be the case if we purge them.

Screenshot 2022-10-27 at 12 38 27

Screenshot 2022-10-27 at 12 39 49

NOTE: until the dashboard is redesigned to have tabs (there is a separate ticket for that), any submitted application you access through the dashboard will give error, but if you copy and edit the URL so it is in the format: `/completed/applications/{uuid}`, it will work.
Once the dashboard is redesigned, there will be a tab for submitted applications and their links will be correct.

Try to access any journey URL of a submitted application, will give you error.
Submit all the way to the end an application, the confirmation page works, is bookmarkable and do not rely on the record (even if technically still exists in the local DB, but you can destroy it and still will work).


app/controllers/submitted_applications_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    # `submitted` or `returned` to render one or the other list.
+    #
+    # Example of routes:
+    # - /submitted/applications (renders `submitted` status tab)
+    # - /submitted/applications?status=submitted (same as above)
+    # - /submitted/applications?status=returned (renders `returned` tab)
+    #
+    # We don't have yet the `returned` status, and all this is really
+    # a bit pointless as this controller will ONLY talk to the
+    # document store through an API, but we don't have any of that yet.
```
timpeat, line 21: We have the api/applications and api/applications/{uuid} endpoints if you'd like to use this for the time being.

There is a very basic api client on review at the moment: https://github.com/ministryofjustice/laa-review-criminal-legal-aid/blob/main/app/lib/api_client.rb

timpeat (approved):
👍

app/controllers/submitted_applications_controller.rb
zheileman: Yes but wouldn't that API client be reading the applications back from our own database? 😄 
But yes, at some point, we will have to start decoupling from ActiveRecord and using something that feels more like "we don't have this locally anymore".

kennyevil (approved):
LGTM

PR #162, opened 2022-10-27
Update dev api to only list submitted applications

timpeat:
## Description of change

Update dev api to a version that only lists submitted applications


zheileman (approved):

willmcb (approved):
👍

PR #164, opened 2022-10-28
Configurable summary presenters, reuse in DWP journey

zheileman:
## Description of change
This PR is a bit of tidy up, DRY and preparing for continuing with CRIMAP-97 (amend applications).
It looks like lots of changes but it is very minor overall. I've split this into 2 separate commits.

First thing, decouple the views from having to use the `current_crime_application` to know if the change links should show or not. Instead, make it part of the configuration of each of the sections (`editable`), and propagate this to the views as `locals`.

Same if we want to hide the header of the section. Default is false, but we can pass `headless: true` to hide the header.

This makes for more flexible "section" components so we can reuse them in other places (like in the DWP confirm applicant details page).

Also tests are simpler and do not need to have the knowledge or concept of a "crime application" record.

With these tweaks done, we can now reuse the `Summary::Sections::ClientDetails`, the same we use in the check your answers, for the DWP passporting journey too.

Screenshot 2022-10-28 at 15 34 14
Screenshot 2022-10-28 at 15 34 02

Everything should keep working as before, there are no functional changes.

### DIFF
--- a/app/controllers/steps/dwp/confirm_details_controller.rb
+++ b/app/controllers/steps/dwp/confirm_details_controller.rb
@@ -1,6 +1,8 @@
 module Steps
   module Dwp
     class ConfirmDetailsController < Steps::DwpStepController
+      before_action :set_presenter
+
       def edit
         @form_object = ConfirmDetailsForm.new(
           crime_application: current_crime_application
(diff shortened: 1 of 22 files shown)


arthurashman (approved):
Excellent work, this is really great - all working for me in manual test too

timpeat (approved):
👍

PR #165, opened 2022-10-28
Add footer links, routes and placeholder html

arthurashman:
## Description of change
Adds links to the footer for:
Cookies
Contact
Feedback
Privacy policy
Accessibility statement 
Terms and conditions 

This follows design of Crime Apply Prototype 
Links to the same external pages for cookies and terms and conditions as Civil Apply Prototype (cookies is a generic gov.uk cookies page - their staging env goes to a service specific one)

Adds routes and html files for contact, feedback, accessibility - content needs creating and updating. 

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-8

### Before changes:
image

### After changes:
image


app/views/about/accessibility.en.html.erb
```diff
@@ -0,0 +1,157 @@
+<%# CONTNENT NEEDS CREATING %>
+
+<% title 'Privacy policy' %>
```
zheileman, line 3: Are all pages privacy policy content, on purpose? 🤣 

If it is a placeholder, wouldn't be better to have at least their title correspond with what the page is suppose to be, and then as body something like "Placeholder. Copy to be decided." or similar?

Having all them the same is confusing.

app/views/layouts/_footer_links.html.erb
```diff
@@ -2,6 +2,35 @@
...
+  </li>
+
+  <li class="govuk-footer__inline-list-item">
+    <%# Content needs creating %>
+
+    <%= link_to 'Feedback', 'about/feedback', class: 'govuk-footer__link' %>
+  </li>
+
+  <li class="govuk-footer__inline-list-item">
+    <%= link_to 'Privacy policy', 'about/privacy', class: 'govuk-footer__link' %>
```
zheileman, line 23: We can use the route helpers for the links, i.e. `about_privacy_path`

zheileman (approved):
👍

app/views/layouts/_footer_links.html.erb
arthurashman: 👍 Thank you

app/views/about/accessibility.en.html.erb
arthurashman: 👍 Good idea. Done!

kennyevil (approved):

PR #166, opened 2022-10-31
CRIMAP-97 Amend returned application

zheileman:
## Description of change
Follow-up to previous preparation PRs, this one implements a barebones "amend" action, with a button in the certificate page so the provider can put back an application into "editable" mode (`in_progress`) and make changes in order to submit again the application.

Note: there are so many implementation details we are still missing that is not worth it taking this further until we have perhaps the document store.

For now is a bit smoke and mirrors but it serve the purpose of getting an idea how the user experience will look like and we can test this with providers.

I'm awaiting confirmation regarding some of the time stamps on the application, if we have to keep the originals or reset on re-submission. I think we might have to introduce a `parent_id` at some point, it will keep track of previously submitted applications to create kind of a history of all amendments, but didn't want to add that DB attribute until we have more clarification.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-97

Screenshot 2022-10-31 at 11 43 30

To "fake" a returned application, access a submitted application certificate (URL is `/completed/applications/{uuid}`) and then in the developer tools in the footer click the button `Mark as returned`. This will take you back to the dashboard and you will see a returned application.

Access again its certificate page (same URL as before) and there should be a blue "Update application" button. Clicking this button will set the application back into `in_progress` status and take you to the Check your answers where you can change anything and progress all the way to the end to sign the declaration and submit again the application.

### DIFF
--- a/app/assets/stylesheets/local/custom.scss
+++ b/app/assets/stylesheets/local/custom.scss
@@ -49,8 +49,17 @@
   }
 }
 
+$app-blue-button-colour: govuk-colour("blue");
+.app-button--blue {
+  background-color: $app-blue-button-colour;
+  box-shadow: 0 $button-shadow-size 0 govuk-shade($app-blue-button-colour, 60%);
+
+  &:hover {
+    background-color: govuk-shade($app-blue-button-colour, 20%);
+  }
+}
+
 // style accessible autocomplete
 .autocomplete__wrapper * {
   @include govuk-typography-common();
 }
-
--- a/app/controllers/completed_applications_controller.rb
+++ b/app/controllers/completed_applications_controller.rb
@@ -27,6 +27,13 @@ class CompletedApplicationsController < DashboardController
     )
   end
 
+  def amend
+    ApplicationAmendment.new(current_crime_application).call
+
+    # Redirect to check your answers (review) page
+    redirect_to edit_steps_submission_review_path(current_crime_application)
+  end
+
   private
 
   # TODO: this will go to the document store when we have it.
--- a/app/services/application_amendment.rb
+++ b/app/services/application_amendment.rb
@@ -1,0 +1,25 @@
+class ApplicationAmendment
+  attr_reader :crime_application
+
+  def initialize(crime_application)
+    @crime_application = crime_application
+  end
+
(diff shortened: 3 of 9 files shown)


arthurashman (approved):
Looks good to me, and works well. 
Would it be possible to make the applicant's name link through again once marked as returned? 
Also, we need to review the copy of the contact us inset box once returned, as it basically suggests you need to call to amend, but then there is a button that contradicts that and say Amend

> If you need to withdraw or update this application, contact the criminal applications helpline.

@carolinapizatto Any thoughts?

zheileman:
@arthurashman

> Would it be possible to make the applicant's name link through again once marked as returned?

Not sure I know what you mean by that?

Is it the links for submitted/returned in the dashboard? At the moment these are broken, the PR Chris is working on will resolve them. I'm reluctant to change anything in that page, as there will be conflicts with that other work.

Regarding the copy in the certificate page, you're right, I added a comment in the design, will get any update to this copy if needed 👍

app/assets/stylesheets/local/custom.scss
```diff
@@ -49,8 +49,17 @@
   }
 }
 
+$app-blue-button-colour: govuk-colour("blue");
+.app-govuk-button--blue {
```
kennyevil, line 53: I think there's already a class for this ```app-button--inverted```

kennyevil (approved):
Looks good, just the css class to address.

kennyevil (changes requested):
Just the css class issue to address

arthurashman:
> 

Yeah, just the link from the dashboard 404s. No problem if it isn't possible to do cleanly yet

app/assets/stylesheets/local/custom.scss
zheileman: I'm afraid it is not the same 😞 
The inverted will set the background colour to white. We need it blue.
It works well on the blue containers, but outside a blue container is very different:

Screenshot 2022-10-31 at 15 07 18

Screenshot 2022-10-31 at 15 07 54

However to follow the same naming pattern I think I will rename `app-govuk-button--blue` to `app-button--blue` 👍

kennyevil (approved):

PR #168, opened 2022-10-31
Add settings mechanism, similar to `FeatureFlags`

zheileman:
## Description of change
We had `FeatureFlags` in the application already, and back at the time we talked about having a similar approach for general settings, those not requiring to change between environments.

These settings are added sometimes to the `application.rb` file and then retrieved with `Rails.configuration.x.whatever` but I think it is cleaner to have them in a separate file not relying on Rails.

The file is the same used for the feature flags too, so all is kept in the same place for easy tweaking.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,19 @@ FeatureFlags.your_new_feature.disabled?
 ```
 
 
+## Settings
+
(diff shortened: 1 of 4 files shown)


kennyevil (approved):

arthurashman (approved):
LGTM

PR #169, opened 2022-11-01
CRIMAP-173 Use sub-navigation in dashboard

kennyevil:
## Description of change

Adds subnavigation to the applications page

## Link to relevant ticket

https://dsdmoj.atlassian.net/browse/CRIMAP-173

0 0 0 0_3000_completed_applications

Complete and submit an application and then you should see it under submitted in 'you applications'. When you click the link for the application it should take you to the show page for that application. In developer tools, there will be an option to mark the application as 'returned'

### DIFF
--- a/app/controllers/completed_applications_controller.rb
+++ b/app/controllers/completed_applications_controller.rb
@@ -12,13 +12,13 @@ class CompletedApplicationsController < DashboardController
                         .returned
                         .joins(:people)
                         .includes(:applicant)
-                        .merge(CrimeApplication.order(created_at: :desc))
+                        .merge(CrimeApplication.order(submitted_at: :desc))
                     else
                       CrimeApplication
                         .submitted
                         .joins(:people)
                         .includes(:applicant)
-                        .merge(CrimeApplication.order(created_at: :desc))
+                        .merge(CrimeApplication.order(submitted_at: :desc))
                     end
 
     @in_progress_applications_count = CrimeApplication.in_progress
--- a/app/controllers/completed_applications_controller.rb
+++ b/app/controllers/completed_applications_controller.rb
@@ -37,6 +37,13 @@ class CompletedApplicationsController < DashboardController
     )
   end
 
+  def amend
+    ApplicationAmendment.new(current_crime_application).call
+
+    # Redirect to check your answers (review) page
+    redirect_to edit_steps_submission_review_path(current_crime_application)
+  end
+
   private
 
   # TODO: this will go to the document store when we have it.
@@ -44,6 +51,6 @@ class CompletedApplicationsController < DashboardController
   # not purging applications on submission yet.
   #
   def current_crime_application
-    @current_crime_application ||= CrimeApplication.submitted.find_by(id: params[:id])
+    @current_crime_application ||= CrimeApplication.not_in_progress.find_by(id: params[:id])
   end
 end
--- a/app/controllers/completed_applications_controller.rb
+++ b/app/controllers/completed_applications_controller.rb
@@ -2,24 +2,30 @@ class CompletedApplicationsController < DashboardController
   before_action :check_crime_application_presence,
                 :present_crime_application, only: [:show]
 
+  # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+  # TODO: Applications will change when we have the document store
+  # working so then we can use a nicer query that won't require
+  # disabling one of the cops.
   def index
-    # TODO: to be implemented.
-    #
-    # My suggestion is we use this action to render the `submitted`
-    # and `returned` listings. The default being `submitted` (we
-    # scope the collection `CrimeApplication.submitted`), and optional
-    # if a "status" query param is provided, we can filter by
-    # `submitted` or `returned` to render one or the other list.
-    #
-    # Example of routes:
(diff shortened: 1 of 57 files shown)


config/locales/en/dashboard.yml
```diff
@@ -16,6 +16,18 @@ en:
         in_progress: In progress
         submitted: Submitted
         returned: Returned
+      sub_navigation:
+        link_1: In progress %{count}
+        link_2: Submitted %{count}
+        link_3: Returned %{count}
+    show:
```
zheileman, line 23: I think this `show` action is no longer in this controller but instead in the `completed_applications` one so we can remove these locales. Perhaps were reintroduced in a rebase, sorry about that!

app/views/completed_applications/index.html.erb
```diff
@@ -1,3 +1,95 @@
...
+
+    <h1 class="govuk-heading-xl"><%=t '.heading' %></h1>
+
+    <%= button_to crime_applications_path,
+                  class: 'govuk-button govuk-button--start',
+                  data: { module: 'govuk-button' } do; t('.start_button'); end %>
+  </div>
+</div>
+
+<nav class="moj-sub-navigation" aria-label="Sub navigation">
```
zheileman, line 15: I think we may be able to extract the sub navigation into its own partial and just pass as `locals` the different counters?
I've added another comment to make the code simpler too, leveraging the locales, with no need for if-else.

app/controllers/crime_applications_controller.rb
```diff
@@ -1,14 +1,18 @@
 class CrimeApplicationsController < DashboardController
   before_action :check_crime_application_presence,
-                :present_crime_application, except: [:index, :create]
+                :present_crime_application, only: [:show, :edit, :destroy, :confirm_destroy]
```
zheileman, line 3: We moved the `#show` action into the other controller.

app/controllers/completed_applications_controller.rb
```diff
@@ -2,24 +2,34 @@ class CompletedApplicationsController < DashboardController
...
+                        .includes(:applicant)
+                        .merge(CrimeApplication.order(created_at: :desc))
+                    end
+
+    @in_progress_applications_count = CrimeApplication.in_progress
+                                                      .joins(:people)
+                                                      .includes(:applicant)
+                                                      .merge(Applicant.with_name)
+                                                      .count
+    @submitted_applications_count = CrimeApplication.submitted.count
```
zheileman, line 29: Tiny thing but might be a big win on the long run and simplify a bit the code, we don't need counter for submitted applications as per design (as this counter may run into the thousands anyways and is not very helpful to providers).

app/controllers/completed_applications_controller.rb
```diff
@@ -2,24 +2,34 @@ class CompletedApplicationsController < DashboardController
...
-    #
-    # We don't have yet the `returned` status, and all this is really
-    # a bit pointless as this controller will ONLY talk to the
-    # document store through an API, but we don't have any of that yet.
+    @applications = if params[:q] == 'returned'
+                      CrimeApplication
+                        .returned
+                        .joins(:people)
+                        .includes(:applicant)
+                        .merge(CrimeApplication.order(created_at: :desc))
```
zheileman, line 15: I think for submitted or returned we want this sorted by `submitted_at: :desc`

Agree with not over optimising these queries right now, as it all will change with the document store 👍

zheileman (changes requested):
Looking very nice! I added a few comments, nothing major but can have a big impact on simplifying the code 👍

arthurashman (approved):
Awesome! Looks and works as expected. 

Agree with Jesus' suggestions

config/locales/en/dashboard.yml
```diff
@@ -55,3 +87,14 @@ en:
...
+    subnavigation:
+        in_progress:
+          zero: In progress
+          one: In progress (1)
+          other: In progress (%{count})
+        submitted: Submitted
+        returned:
+          zero: Returned
+          one: Returned (1)
+          other: Returned %{count}
```
zheileman, line 100: Missing parenthesis in the counter 😛

config/locales/en/dashboard.yml
```diff
@@ -42,6 +52,28 @@ en:
...
+      table_headings:
+        applicant_name: Name
+        submitted_at: Submitted
+        reference: LAA reference
+        status: Status
+        action: Action
+      status:
+        in_progress: In progress
+        submitted: Submitted
+      sub_navigation:
```
zheileman, line 67: Do we need these locales if we are using the partial?

config/locales/en/dashboard.yml
```diff
@@ -16,6 +16,16 @@ en:
         in_progress: In progress
         submitted: Submitted
         returned: Returned
+      sub_navigation:
```
zheileman, line 19: Same with this partials.

zheileman (approved):
👍 Cool. Just a tiny thing with the locales, but I'm approving anyways.
As part of another PR we can clean up the status column and tags as they are no longer used with this new design.

config/locales/en/dashboard.yml
kennyevil: *swearwords*

PR #172, opened 2022-11-02
remove status tags from applications

kennyevil:
## Description of change

Removes status tags from the applications page as they are no longer required.


zheileman (approved):
🥳 Removing code is always good!

PR #180, opened 2022-11-15
Add passported legal aid section to review page

arthurashman:
## Description of change
Add Section for passported on IOJ on review your answers page 

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-208
## Notes for reviewer
To match [designs](https://www.figma.com/file/thkvnDBbQbHlSqzt1TE3lk/Crime-Apply?node-id=2711%3A6322&t=4Tjt7BFzGSE2r0uE-1)
Go through the flow with user:
surname: 'POTTER'
dob: '01-Jan-2007'
nino: 'NC010155A'

After hearing_details step, you will be taken to the passported interruption card
Save and continue takes you to the review page.
At the bottom of the review page will be a Justification for Legal Aid section with 'Passported': 'Details provided do not require further justification for legal aid'

### DIFF
--- a/spec/presenters/summary/html_presenter_spec.rb
+++ b/spec/presenters/summary/html_presenter_spec.rb
@@ -7,9 +7,10 @@ describe Summary::HtmlPresenter do
     instance_double(CrimeApplication, applicant: double, case: kase)
   end
 
-  let(:kase) { instance_double(Case, ioj:) }
+  let(:kase) { instance_double(Case, ioj:, ioj_passport:) }
 
   let(:ioj) { instance_double(Ioj) }
+  let(:ioj_passport) { ['foo'] }
 
   describe '#sections' do
     before do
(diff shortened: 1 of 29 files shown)


zheileman (approved):
I pushed a few minor tweaks! Approving.

arthurashman:
@zheileman I added another commit to make the section conditional on if there is no ioj reason present. I spoke with Carol re future sampling and she said that this passported section should only show if there was no sampling and therefore an ioj reason input.

zheileman:
@arthurashman should we worry about the sampling later? I feel like the code is being made complex based on something we don't even have yet. We can worry about how best to threat the sampling when time comes.

There are a few things about this sampling, the passporting, the IoJ that are moving targets at the moment and is difficult to get to a nice solution 😄

spec/presenters/summary/html_presenter_spec.rb
```diff
@@ -18,20 +18,42 @@
       ).to receive(:show?).and_return(true)
     end
 
-    it 'has the right sections in the right order' do
```
zheileman, line 21: HtmlPresenter doesn't care about conditionality. It is a smoke test so to speak, to ensure all the sections appear in the right order. We allow `#show?` on purpose to always return true on the BaseSection to accomplish this.

Conditionality is tested at the section level, in each of the specific sections.

kennyevil (approved):

PR #189, opened 2022-11-22
CRIMAP-211 Add JSON adapter and serialisers

kennyevil:
## Description of change

Adds JSON adapter that conforms to expected schema for the datastore

## Link to relevant ticket

https://dsdmoj.atlassian.net/jira/software/projects/CRIMAP/boards/897?selectedIssue=CRIMAP-211


app/serializers/submission_serializer/definitions/address.rb
```diff
@@ -0,0 +1,15 @@
+module SubmissionSerializer
+  module Definitions
+    class Address < Definitions::BaseDefinition
+      def to_builder
+        Jbuilder.new do |json|
```
zheileman, line 5: We need to also add the `lookup_id` attribute as it will be needed (eventually) for re-hydration.

app/serializers/submission_serializer/sections/application_details.rb
```diff
@@ -0,0 +1,20 @@
...
+        Jbuilder.new do |json|
+          json.id crime_application.id
+          json.schema_version 1.0
+          json.application_type ''
+          json.reference crime_application.reference
+          json.created_at crime_application.created_at
+          json.submitted_at crime_application.submitted_at
+          json.date_stamp crime_application.date_stamp
+          json.status crime_application.status
+        end
```
zheileman, line 15: Also here we will need the new `ioj_passport` attribute introduced recently.

app/serializers/submission_serializer/definitions/person.rb
```diff
@@ -0,0 +1,15 @@
+module SubmissionSerializer
+  module Definitions
+    class Person < Definitions::BaseDefinition
+      def to_builder
+        Jbuilder.new do |json|
+          json.first_name first_name
+          json.last_name last_name
+          json.date_of_birth date_of_birth
+
+          json.address Definitions::Address.generate(home_address)
```
zheileman, line 10: Be aware this attribute is now named `home_address` also in the JSON document.

app/serializers/submission_serializer/sections/case_details.rb
```diff
@@ -0,0 +1,30 @@
...
+        Jbuilder.new do |json|
+          json.case_details do
+            json.urn crime_application.case.urn
+            json.case_type crime_application.case.case_type
+            json.appeal_maat_id crime_application.case.appeal_maat_id
+            json.appeal_with_changes_maat_id crime_application.case.appeal_with_changes_maat_id
+            json.appeal_with_changes_details crime_application.case.appeal_with_changes_details
+            json.hearing_court_name crime_application.case.hearing_court_name
+            json.hearing_date crime_application.case.hearing_date
+            json.offences(crime_application.case.charges.map { |charge| Definitions::Offence.generate(charge) })
```
zheileman, line 16: Definitions probably should take care of rendering collections. But that can be done as part of a separate PR.

I'm thinking something like:

```ruby
json.offences Definitions::Offence.generate(crime_application.case.charges)
```

And then `Definitions::Offence` will deal with the collection, iterating through it to render the different items.

app/serializers/submission_serializer/sections/provider_details.rb
```diff
@@ -0,0 +1,23 @@
+module SubmissionSerializer
+  module Sections
+    class ProviderDetails < Sections::BaseSection
```
zheileman, line 3: We can clean up the example code here, and `min_version` because `provider_details` is now a required object in the schema. But for now it can be an empty object (`{}`)

app/serializers/submission_serializer/definitions/offence.rb
```diff
@@ -0,0 +1,13 @@
+module SubmissionSerializer
+  module Definitions
+    class Offence < Definitions::BaseDefinition
+      def to_builder
+        Jbuilder.new do |json|
+          json.name offence_name
```
zheileman, line 6: From a `Charge` instance you can get the `Offence` instance with `charge.offence` and then get the name and the class with `charge.offence.name` (this is the same as `charge.offence_name`) and `charge.offence.offence_class`

zheileman:
@kennyevil Looking great!

I added a few comments, but nothing major and some can be done as a separate PR I think.

Also if you see this PR grows a bit too much, maybe you can remove a couple sections to be completed in a separate PR 👍

kennyevil:
@zheileman For now, I've skipped over fully validating the output until we have the provider details in place.

zheileman:
@kennyevil I've pushed a few tweaks, I think the only remaining bit might be finalise the `applicant` section. We can do that in a separate PR maybe, as I think we will need to model it similar to the schema/structs, where a Person only declares a few basic/common attributes, and then Applicant extend it with more attributes, and Codefendant the same. It will make it easier when we introduce Partner too 😄

PR #200, opened 2022-11-30
Remove dev api gem + config for it

willmcb:
## Description of change
Now that we have the client API and can set up the datastore locally, there is no need for the direct dev-api gem so this PR removes it from apply.

## Link to relevant ticket
n/a


timpeat (approved):
👍

PR #209, opened 2022-12-05
CRIMAP-235 Spike SAML authentication

zheileman:
## Description of change
WIP. This is a draft. Ignore for now.


zheileman:
Another approach, without Devise for now, will be used. Will raise soon the PR.

PR #219, opened 2022-12-13
Crimap 105 add pagination

kennyevil:
## Description of change

Adds pagination to the applications page

## Link to relevant ticket

https://dsdmoj.atlassian.net/jira/software/projects/CRIMAP/boards/897?selectedIssue=CRIMAP-105

0 0 0 0_3000_applications (2)

If there are more than 10 applications in any given status, the pagination should kick in.


app/views/crime_applications/index.html.erb
```diff
@@ -37,5 +37,6 @@
         <% end %>
       </tbody>
     </table>
+    <%= paginate @applications %>
```
zheileman, line 40: I’ve seen there are some duplication of partials? I suppose it is to cope with the 2 different controllers we have (for in progress and for completed applications)? I’m no expert on kaminari but I believe there might be a way to configure the controller that kaminari will use perhaps so we reuse partials and code?

I believe something like `paginate @applications, params: { controller: 'foobar' }`

Maybe give it a try?

app/controllers/completed_applications_controller.rb
```diff
@@ -30,10 +30,7 @@ def applications_from_datastore
       status: status_filter, **pagination_params
     ).call
 
-    @paginator = InfinitePaginationV2.new(
```
zheileman, line 33: I think there are some locales we can also remove that were used for this infinite paginator.

app/views/completed_applications/index.html.erb
```diff
@@ -32,6 +32,6 @@
       </tbody>
     </table>
 
-    <%= render @paginator if @paginator %>
+    <%= render partial: 'pagination', locals: @pagination.symbolize_keys %>
```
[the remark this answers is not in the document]
zheileman: Yep I got it working (as far as I can tell in my tests) just by using this:

```ruby
<%= paginate Kaminari.paginate_array(@applications, total_count: @pagination['total_count']).page(params[:page]) %>
```

I removed the pagination partials inside `app/views/completed_applications/`, and it seems to be working as expected.

app/views/crime_applications/index.html.erb
zheileman: No need for this anymore, read my other comment 👍

zheileman:
Did you run `rails g kaminari:views default` to generate all the default views? I think that is a sensible approach, it will generate all the necessary partials, with comments included that are very handy to know what is going on. I suggest not removing the comments, otherwise, is difficult to know about all these magic local variables Kaminari exposes.

Then you can modify the markup etc. to our needs, but trying to still use as much as possible of what Kaminari offers, like localisation (they go by default to `views.pagination.*` I think is fine to have them as you already have in `pagination.*`).

app/views/completed_applications/index.html.erb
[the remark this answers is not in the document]
zheileman: Taking it further, we can have just `<%= paginate @applications %>` here too (same as in the other view for consistency).
And we tweak the method `applications_from_datastore` to return the `Kaminari.paginate_array ...` instead.

app/controllers/completed_applications_controller.rb
```diff
@@ -30,12 +30,7 @@ def applications_from_datastore
...
-      pagination: result.pagination,
-      params: params,
-    )
-
-    result
+    Kaminari.paginate_array(result, total_count: result.pagination['total_count']).page(params[:page])
```
zheileman, line 33: Nice!
Tiny thing but can we have the `.page(params[:page])` up in the `#index` action, as to unify the interface in both controllers. No blocker!

zheileman (approved):
🥳

PR #225, opened 2022-12-14
Crimap 238 graceful errors

kennyevil:
## Description of change

Add error pages on the applications page for when the datastore is unavailable.

## Link to relevant ticket

https://dsdmoj.atlassian.net/jira/software/projects/CRIMAP/boards/897?selectedIssue=CRIMAP-238

0 0 0 0_3000_completed_applications_q=returned (1)

Turn off the datastore and attempt to access the `submitted` or `returned` tabs on the application page.


app/controllers/completed_applications_controller.rb
```diff
@@ -3,7 +3,9 @@ class CompletedApplicationsController < DashboardController
                 :present_crime_application, only: [:show]
 
   def index
-    @applications = applications_from_datastore.page(params[:page])
+    @applications = Datastore::GetApplications.new(
+      **pagination_params
```
zheileman, line 7: I suggest to keep the filter params (status, but soon may be others) and the pagination params separate, it is confusing otherwise as not all is pagination.

app/services/datastore/get_applications.rb
```diff
@@ -0,0 +1,34 @@
...
+        status: status_filter,
+        page: @page,
+        per_page: @per_page,
+        sort: @sort
+      ).call
+
+      Kaminari.paginate_array(result, total_count: result.pagination['total_count'])
+    rescue StandardError => e
+      Rails.logger.error(e)
+      Sentry.capture_exception(e)
```
zheileman, line 21: Does this return a `nil`? The call to capture_exception might be returning any other thing and then the safe navigation `&.page` might trigger inadvertently. I think is best to explicitly add a `nil` as a return value to the rescue.

config/locales/en/errors.yml
```diff
@@ -29,6 +29,11 @@ en:
       heading: Access restricted
       lead_text: This page requires you to be signed in to LAA Portal. You will be redirected back to this service after you sign in.
       sign_in_button: Sign in with LAA Portal
+    datastore:
```
zheileman, line 32: It feels like this might be best to place in the existing `dashboard.yml` maybe under `en.shared.errors`?

These other errors in this locale are full page errors, each one being a routable URL.

zheileman (changes requested):
Cool, a few comments!

zheileman (approved):

PR #228, opened 2022-12-15
Crimap 246 no applications available

kennyevil:
## Description of change

Adds message for users if no applications are available

## Link to relevant ticket

https://dsdmoj.atlassian.net/jira/software/projects/CRIMAP/boards/897?selectedIssue=CRIMAP-243

0 0 0 0_3000_applications (3)

destroy all application records in all three states

### DIFF
--- a/app/views/completed_applications/index.html.erb
+++ b/app/views/completed_applications/index.html.erb
@@ -4,15 +4,15 @@
 
 <%= render partial: 'shared/subnavigation' %>
 
-  <div class="govuk-grid-row">
-    <div class="govuk-grid-column-full">
-      <% if @applications.nil? %>
-        <%= render partial: 'datastore_unavailable' %>
-      <% elsif @applications.empty? %>
-        <h2 class="govuk-heading-m">
-          <%= t('.no_records') %>
-        </h2>
-      <% else %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <% if @applications.nil? %>
+      <%= render partial: 'datastore_unavailable' %>
+    <% elsif @applications.empty? %>
+      <h2 class="govuk-heading-m">
+        <%= t('.no_records', status: (params[:q] ||= 'submitted')) %>
+      </h2>
+    <% else %>
       <table class="govuk-table app-dashboard-table">
         <thead class="govuk-table__head">
           <tr class="govuk-table__row">
@@ -32,14 +32,13 @@
                 <%= l(app.submitted_at) %>
               </td>
               <td class="govuk-table__cell">
-              <%= app.laa_reference %>
-            </td>
-          </tr>
-        <% end %>
-      </tbody>
-    </table>
-    <%= paginate @applications %>
-
+                <%= app.laa_reference %>
+              </td>
+            </tr>
+          <% end %>
+        </tbody>
+      </table>
+      <%= paginate @applications %>
     <% end %>
   </div>
 </div>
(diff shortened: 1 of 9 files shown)


app/views/completed_applications/index.html.erb
```diff
@@ -4,11 +4,15 @@
 
 <%= render partial: 'shared/subnavigation' %>
 
-<% if @applications.nil? %>
-  <%= render partial: 'datastore_unavailable' %>
-<% else %>
   <div class="govuk-grid-row">
```
zheileman, line 7: I think there might be some extra indentation, not a biggie tho.

app/views/completed_applications/index.html.erb
```diff
@@ -4,11 +4,15 @@
...
     <div class="govuk-grid-column-full">
+      <% if @applications.nil? %>
+        <%= render partial: 'datastore_unavailable' %>
+      <% elsif @applications.empty? %>
+        <h2 class="govuk-heading-m">
+          <%= t('.no_records') %>
+        </h2>
+      <% else %>
       <table class="govuk-table app-dashboard-table">
```
zheileman, line 16: Here as well some indentation seems off.

config/locales/en/dashboard.yml
```diff
@@ -50,6 +51,7 @@ en:
       status:
         in_progress: In progress
         submitted: Submitted
+      no_records: There are no applications in progress
```
zheileman, line 54: This needs to be a different message, as these are submitted/returned applications. Not in progress.

zheileman (changes requested):
Cool, just a few minor comments!

config/locales/en/dashboard.yml
kennyevil: Just realised that I misinterpreted what was meant by "the solution is valid for all three statuses". I'll change that now.

app/views/completed_applications/index.html.erb
```diff
@@ -4,11 +4,15 @@
...
-<% else %>
-  <div class="govuk-grid-row">
-    <div class="govuk-grid-column-full">
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <% if @applications.nil? %>
+      <%= render partial: 'datastore_unavailable' %>
+    <% elsif @applications.empty? %>
+      <h2 class="govuk-heading-m">
+        <%= t('.no_records', status: (params[:q] ||= 'submitted')) %>
```
zheileman, line 13: I feel this may be dangerous. We are blindly propagating whatever value is in the `q` query param to the locales. For once it will blow up for anything like `foobar`, but there might be other side effects.

I think the way to handle this is to have a helper method `status_filter` in the dashboard controller (as that is the base controller the other inherit from) that is implemented in both controllers, in one it just return `in_progress` for instance, and in the completed controller it takes and sanitises the `param[:q]` (then we don't need that code in the GetApplications service anymore, we just pass the proper sanitised value to the service object).

Then in locales we have add to the shared errors a new section `no_records` for instance and declare the 3 different messages. Propagating the raw `q` value doesn't work if we translate to welsh.

config/locales/en/dashboard.yml
```diff
@@ -14,6 +14,7 @@ en:
         in_progress: In progress
         submitted: Submitted
         returned: Returned
+      no_records: There are no applications in progress
```
zheileman, line 17: We can now remove this locale, and the other one as well, as now it is using `shared.errors` right?

zheileman (approved):
Nice!
A previous locales can now be removed. Other than that, approving 👍

PR #247, opened 2023-01-13
update hint text on partner page

kennyevil:
## Description of change

Remove hint text on the partnership page and replace with a list of definitions of a partner

## Link to relevant ticket

https://dsdmoj.atlassian.net/jira/software/projects/CRIMAP/boards/897?selectedIssue=CRIMAP-230

### Before changes:
0 0 0 0_3000_applications_8ee35ef9-fd58-4246-b1e5-2ee4122587cf_steps_client_has_partner (1)

### After changes:

0 0 0 0_3000_applications_a573c49a-97fb-4444-a493-fbea111f5d9e_steps_client_has_partner (2)

### DIFF
--- a/app/views/steps/client/has_partner/edit.html.erb
+++ b/app/views/steps/client/has_partner/edit.html.erb
@@ -6,7 +6,15 @@
     <%= govuk_error_summary(@form_object) %>
     <%= step_form @form_object do |f| %>
         <%= f.govuk_collection_radio_buttons :client_has_partner, @form_object.choices,
-                                             :value, inline: false, legend: { tag: 'h1', size: 'xl' } %>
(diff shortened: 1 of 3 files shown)


zheileman (commented):
I think this is not hint text in the new designs.

This will mostly be injected content, as per:
https://govuk-form-builder.netlify.app/building-blocks/injecting-content/

Also perhaps double checking the radios being inline or not. Based on the design system guidelines, when there are only 2 options and these are short (like Yes / No) inline is possible. I think we have inline in all the other simple Yes / No questions.

kennyevil:
Had a talk with Carol and she says it should be stacked. The convention we're following is that , if the yes/no question is the only question then we should stack it, but if it's a sub question on the page then it should be inline.

zheileman (approved):
👍

PR #251, opened 2023-01-13
Crimap 104 sorting

kennyevil:
## Description of change

Add sortable columns to crime applications pages

## Link to relevant ticket

https://dsdmoj.atlassian.net/browse/CRIMAP-226

## Notes for reviewer

After discussion, it was decided that sorting by name was not needed for these pages, so sorting by start date/submitted at/returned at is the only requirement

### Before changes:
0 0 0 0_3000_applications_order=created_at sort=desc (1)

### After changes:
0 0 0 0_3000_applications_order=created_at sort=desc


app/controllers/completed_applications_controller.rb
```diff
@@ -29,12 +29,18 @@ def amend
...
+      sort: sort_direction,
       page: params[:page],
       per_page: Kaminari.config.default_per_page,
     }
   end
 
+  def sort_direction
+    return nil if params[:sort].blank?
+
+    "#{params[:sort]}ending"
```
zheileman, line 41: This feels a bit sketchy 😅 

Maybe we can keep a map of what internally we use (`asc` / `desc`) and what the datastore uses (`ascending` / `descending`). It is sad we couldn't keep this consistent in the datastore. And perhaps hide that logic or mapping in the service object, `GetApplications`.

app/assets/stylesheets/local/custom.scss
```diff
@@ -83,3 +83,70 @@ $app-blue-button-colour: govuk-colour("blue");
 .autocomplete__wrapper * {
   @include govuk-typography-common();
 }
+
+[aria-sort] .sorting,
```
zheileman, line 87: As this is quite a lot of custom CSS, perhaps we can create a separate `sorting.scss` file and `@import` it from the application.scss file, similar to what we do with some other big chunks of styling and custom components like the info_card, or the suggestions.

zheileman (commented):
I need to play a bit more with this but looking great!
A couple suggestions, nothing major.

app/controllers/crime_applications_controller.rb
```diff
@@ -29,4 +29,24 @@ def destroy
...
+
+    params[:order]
+  end
+
+  def ordering_param_allowed?
+    %w[created_at].include? params[:order]
+  end
+
+  def sort_param
+    params[:sort]&.to_sym || :desc
```
zheileman, line 46: I think we will need to whitelist this param as well, because otherwise you can pass whatever you want in it and will try to trigger the sorting, which may pose a risk.

app/controllers/crime_applications_controller.rb
```diff
@@ -29,4 +29,24 @@ def destroy
...
+    return 'created_at' unless ordering_param_allowed?
+
+    params[:order]
+  end
+
+  def ordering_param_allowed?
+    %w[created_at].include? params[:order]
+  end
+
+  def sort_param
```
zheileman, line 45: Wondering if some of these methods are common to both controllers, the in progress applications (`crime_applications`) and the submitted/returned one (`completed_crime_applications`) in which case ideally we would want to raise those methods to the super class of both controllers, `dashboard_controller` to reuse as a common interface.

app/assets/stylesheets/application.scss
```diff
@@ -17,3 +17,6 @@
 
 // App-specific custom styles and overrides
 @import "local/custom";
+
+// Sorting styles for table headers
+@import "local/sorting";
```
zheileman, line 22: Tiny thing but components is best to all go above the `@import "local/custom";` declaration, leaving this one as the last one, because that way, without having to touch the components itself, we can tweak some things by using the local/custom file.

zheileman (approved):
👍

PR #252, opened 2023-01-16
CRIMAP-239 Preliminary approach to re-hydration

zheileman:
## Description of change
As part of the "returned application journey", on Apply we want to be able to select a returned application in the dashboard, and click a button "Update application" to edit it and re-submit.

The edition in reality is the creation of a brand new application, keeping all the entered information from the original (or parent) application, including same USN.

The provider is then presented with the `Check your answers` page for this new application, with everything pre-filled so they can change anything they are required to.

Then they can submit the application, which will create a new record in the datastore, with a new UUID, but same USN.

This is the first part of probably a few more PRs to finalise this re-hydration, taking a JSON payload and reconstructing a persisted (in Apply local DB) application in draft mode (`in_progress`), including relationships.

In this PR, a basic re-hydration of the top-level application attributes, and the `applicant` (including addresses if any) is included. Some other code tweaks to facilitate this, along with some renaming/refactoring.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-239

## Screenshots of changes (if applicable)
Screenshot 2023-01-16 at 13 17 20
Screenshot 2023-01-16 at 13 17 38
Screenshot 2023-01-16 at 13 17 50

## How to manually test the feature
For the time being, until Review starts "returning" applications, this can be tested by manually updating the status of a submitted application in your local instance of the datastore. Set the status to `returned` and it should appear in the returned tab on Apply dashboard, and then you can re-hydrate it back to the Apply database.

PR #253, opened 2023-01-17
CRIMAP-239 Rehydrate case details

zheileman:
## Description of change
Follow-up to previous PR #252.

In this PR the rehydration of the case details (except IoJ that will be done separate) is implemented.

Most of it has been refactoring previous code that coped with case details JSON from the datastore rendered in the application certificate, to use DB instances instead of Structs so is easier to persist back to the database as we keep the same data modelling in both places. The code has been simplified too.

Still, WIP so some things might not work as expected. Refer to previous PR for more details on how to test this.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-239

Screenshot 2023-01-17 at 10 59 11

Refer to PR #252


timpeat (approved):
makes sense 👍

PR #257, opened 2023-01-19
Use Dockerfile multi-stage builds and update deps

zheileman:
## Description of change
Although currently our docker image is quite contained in size, and it is fast enough to build, some gains can be obtained (and make it easier to expand in the future) by using multi-stage builds.

Removed a few build deps that it seems were not necessary, at least in my local tests, but until a full deployment is done I can't be certain for sure. Worst case scenario we need to add them back.

Additionally, the Alpine image has been bumped to the latest one, 3.16

Results for `time docker build --no-cache .`

```
Original Dockerfile

non-cached build
time - 1:01
size - 877MB

Improved Dockerfile (multi-stage)

non-cached build
time - 0:48
size - 597MB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,29 +1,42 @@
-FROM ruby:3.1.2-alpine3.15
+FROM ruby:3.1.2-alpine3.16 AS base
 MAINTAINER LAA Crime Apply Team
 
-RUN apk --no-cache add --virtual build-deps \
-  build-base \
+# dependencies required both at runtime and build time
+RUN apk add --update \
   postgresql-dev \
-  git \
-  bash \
-  curl \
-&& apk --no-cache add \
-  postgresql-client \
-  shared-mime-info \
-  linux-headers \
-  xz-libs \
   tzdata \
   yarn
 
 # Alpine does not have a glibc, and this is needed for dart-sass
 # Refer to: https://github.com/sgerrand/alpine-pkg-glibc
+ARG GLIBC_VERSION=2.34-r0
 RUN wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://alpine-pkgs.sgerrand.com/sgerrand.rsa.pub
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -33,28 +46,24 @@ RUN mkdir -p /usr/src/app && \
...
+# copy over the remaning files and code
 COPY . .
 
-# The following are ENV variables that need to be present by the time
+# Some ENV variables need to be present by the time
 # the assets pipeline run, but doesn't matter their value.
-ENV SECRET_KEY_BASE=needed_for_assets_precompile
-ENV ENV_NAME=needed_for_assets_precompile
+RUN SECRET_KEY_BASE=needed_for_assets_precompile \
+    RAILS_ENV=production \
```
timpeat, line 61: Always seems an unsatisfactory solution to me, but I've never found a better one. 

Initially I was unhappy with how you've used production for the stub.  But thinking about it, I wonder if using production, as you have done, may be safest as at least it would prevent a test run against the db etc. Was that your reasoning?

timpeat (approved):

Dockerfile
zheileman: These values here doesn't really matter. However, they have to be present for `rails assets:precompile` to work. And they need to "make some sense" (so to speak) as the command will invoke a rails environment. So for instance, `RAILS_ENV` will be checked against `config/environments/*.rb` (development, test and production). So `RAILS_ENV=foobar` will blow up.

And we want to ensure assets are compiled for a production target. Because the compilation could tweak things like _minifying_, _uglifying_, etc. depending on the environment.

The real ENV values for the actual running of the image are always taken from the Kubernetes environment (coming either from config_map, or secrets). Or from the `docker-compose.yml` when running the container locally (`docker-compose up`). Docker compose is never used in the cluster, is only there for our convenience.

Hope it makes sense? Let me know if it is easier to have a chat about it?

PR #260, opened 2023-01-20
CRIMAP-121 Propagate provider email address

zheileman:
## Description of change
On submission, the currently signed in provider email address gets injected into the application to be propagated to the datastore along with the already existing legal representative attributes.

NOTE: on staging we are currently using a somehow lacking SAML mock, which does not return to us any email address. In the interim until we properly integrate with LAA Portal staging environment, a fallback email will be used, just so the schema validates the payload.

Additionally, and as recently discussed, Apply will stop sending a `status` attribute when submitting applications to the datastore. This attribute is redundant on submissions and the datastore was already setting the status to `submitted` anyways.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-121

Submitted applications (if testing locally, requires a datastore running latest version of the schemas gem) will contain the provider email address (inside the `provider_details` attribute). Also, submissions from Apply will not send to the datastore the `status` attribute anymore.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -89,6 +89,10 @@ RSpec/Rails/InferredSpecType:
 Metrics/MethodLength:
   Max: 12
 
+# To be tweaked until we find the right balance
+Metrics/AbcSize:
+  Max: 18
+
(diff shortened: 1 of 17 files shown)


timpeat (approved):
Many thanks!

PR #263, opened 2023-01-24
Set the proper aria-sort value

zheileman:
## Description of change
According to the documentation for `aria-sort` [1], the values should be `ascending`, `descending` or `none`.

Added a few more tests.

[1] https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort

Do an assets clobber and then `rails dartsass:build` to ensure the styles are correct.


app/helpers/dashboard_helper.rb
```diff
@@ -31,8 +31,13 @@ def flipped_sort_direction
     sort_direction == 'asc' ? 'desc' : 'asc'
   end
 
+  # https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort
   def aria_sort(column)
-    sort_by == column ? sort_direction : 'none'
+    if sort_by == column
+      { 'asc' => 'ascending', 'desc' => 'descending' }.fetch(sort_direction)
```
willmcb, line 37: Do we need to have a conversion - why not just use the unabreviated words throughout the application? 🤔 

One thing that I was thinking is that we use desc in other places to mean "description" so I think having the same word mean two different things is a little confusing.

willmcb: I'm looking at the sorting stuff on the Review side as well atm as well fyi :)

zheileman: I like the abbreviated (asc / desc) version for sorting. Is the convention across the board almost in every service and API I've ever worked. The API supports both so feel free to use unabbreviated in Review if you like it more.

`desc` can be use in Grape for the description of the endpoints, but it has as alias `description` which can be use if desc / description causes confusion. Honestly, don't see the confusion with this.

zheileman: @willmcb heads up however, I'm looking at using the component directly from the MOJ frontend,
i.e. `@import "@ministryofjustice/frontend/moj/components/sortable-table/sortable-table";`
instead of having a copy of the file in our repo.

But I've discovered the MOJ uses buttons instead of links for the sorting headers. I will see if we can get around this or we have to use buttons too (they are styled as links anyways so look the same).

willmcb: No worries that makes sense :)

kennyevil (approved):

PR #266, opened 2023-01-30
add provider details to certificate page

kennyevil:
## Description of change

Add legal representative details to the certificate page

## Link to relevant ticket

https://dsdmoj.atlassian.net/jira/software/projects/CRIMAP/boards/897?selectedIssue=CRIMAP-257

0 0 0 0_3000_completed_applications_2a86cf90-d0e7-419b-9d82-25cc1f9dabd0

Submit an application and check the completed application page. Legal representative details should be at the bottom of the page.

### DIFF
--- a/app/presenters/summary/sections/legal_representative_details.rb
+++ b/app/presenters/summary/sections/legal_representative_details.rb
@@ -1,0 +1,45 @@
+module Summary
+  module Sections
+    class LegalRepresentativeDetails < Sections::BaseSection
+      def name
+        :legal_representative_details
+      end
+
+      def show?
+        provider_details.present? && super
+      end
+
+      # rubocop:disable Metrics/MethodLength
+      def answers
+        [
+          Components::FreeTextAnswer.new(
+            :office_code, provider_details.office_code,
+          ),
+
+          Components::FreeTextAnswer.new(
+            :provider_email, provider_details.provider_email,
+          ),
+
+          Components::FreeTextAnswer.new(
+            :legal_rep_first_name, provider_details.legal_rep_first_name,
+          ),
+
+          Components::FreeTextAnswer.new(
+            :legal_rep_last_name, provider_details.legal_rep_last_name,
+          ),
+
+          Components::FreeTextAnswer.new(
+            :legal_rep_telephone, provider_details.legal_rep_telephone, show: true,
+          ),
+        ].select(&:show?)
+      end
+      # rubocop:enable Metrics/MethodLength
+
+      private
+
+      def provider_details
+        @provider_details ||= crime_application.provider_details
+      end
+    end
+  end
+end
--- a/config/locales/en/summary.yml
+++ b/config/locales/en/summary.yml
@@ -17,6 +17,7 @@ en:
       next_court_hearing: Next court hearing
       justification_for_legal_aid: Justification for legal aid
       passport_justification_for_legal_aid: Justification for legal aid
+      legal_representative_details: Legal representative details
 
     questions:
       # BEGIN client details section
@@ -127,3 +128,16 @@ en:
         answers: 
           'true': Details provided do not require further justification for legal aid
       # END passport justification for legal aid section
+
+      # BEGIN legal representative details section
+      office_code:
+        question: Office account number
+      provider_email:
+        question: Email address
+      legal_rep_first_name:
+        question: First name
+      legal_rep_last_name:
+        question: Last name
+      legal_rep_telephone:
+        question: Telephone
+      # END legal representative details section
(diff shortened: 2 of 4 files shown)


app/presenters/summary/sections/legal_representative_details.rb
```diff
@@ -0,0 +1,45 @@
...
+          Components::FreeTextAnswer.new(
+            :legal_rep_first_name, provider_details.legal_rep_first_name,
+          ),
+
+          Components::FreeTextAnswer.new(
+            :legal_rep_last_name, provider_details.legal_rep_last_name,
+          ),
+
+          Components::FreeTextAnswer.new(
+            :legal_rep_telephone, provider_details.legal_rep_telephone, show: true,
```
zheileman, line 32: No need for `show: true` because this is a mandatory (so always present) value. Except maybe the datastore may have some corrupt data 😅

zheileman (commented):
Nice. Just in case, can you double check with @carolinapizatto because I believe she had a different design for this, where the legal rep details and the provider details (including account number and email address) along with some of the other details like date stamp, submission date, etc. are split into 2 sections, one `Overview` and another `Legal representative details` (the one you've just done) 👍

zheileman (commented):
Another thing to consider is perhaps ensure this section is not shown for application still `in_progress` as some details will only be present in a JSON datastore application, not a locally persisted application.

It is kind of an edge case admittedly as the CYA comes before the declaration, but if the declaration fails for whatever reason or the user hits the "Save and come back later" at that point, next time they render the CYA I have the feeling it may blow up.

zheileman (commented):
Sorry I missed this before, can we implement, like with any other summary presenter section, the corresponding unit tests in `spec/presenters/summary/sections`

That way we can test the in progress etc. Which at the moment is not covered.

app/presenters/summary/components/date_time_answer.rb
```diff
@@ -0,0 +1,9 @@
+module Summary
+  module Components
+    class DateTimeAnswer < BaseAnswer
+      def to_partial_path
+        'steps/submission/shared/date_time_answer'
```
zheileman, line 5: Cool I see what you tried to do, and this will totally work. But we can take advantage of the fact the `BaseAnswer` superclass is already prepared to take i18n options, and propagate them to the component.

So we can keep using the existing `DateTime` component, and we pass `i18n_opts: { format: :datetime }`.
There is a tiny change needed in the partial to use these i18n opts:

`l(date_answer.value, **date_answer.i18n_opts)`

And finally a little scenario in the view test here to double check it works as expected 😄 
`spec/views/steps/submission/shared/_date_answer.html.erb_spec.rb`

config/locales/en/summary.yml
```diff
@@ -17,8 +18,22 @@ en:
       next_court_hearing: Next court hearing
       justification_for_legal_aid: Justification for legal aid
       passport_justification_for_legal_aid: Justification for legal aid
+      legal_representative_details: Legal representative details
```
zheileman, line 21: Tiny thing, this will be "Legal representative" as per design, if I'm correct. We can rename the summary class as well to be consistent as we don't use `details` suffix in any of the other section classes.

config/locales/en.yml
```diff
@@ -3,6 +3,7 @@ en:
     DATE_FORMATS: &DATE_FORMATS
       default: '%-d %b %Y'  # 3 Jan 2019
       compact: '%d/%m/%Y'   # 03/01/2019
+      datetime: '%-d %B %Y %-l:%M%P'  # 3 January 2019 2:45pm
```
zheileman, line 6: 👍

spec/presenters/summary/sections/legal_representative_details_section_spec.rb
```diff
@@ -0,0 +1,69 @@
+require 'rails_helper'
```
zheileman, line 1: The name of the spec file contains an additional section suffix. Not critical 😅

spec/presenters/summary/sections/overview_section_spec.rb
```diff
@@ -0,0 +1,78 @@
+require 'rails_helper'
```
zheileman, line 1: Same with this file name.

spec/presenters/summary/sections/overview_section_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+
+describe Summary::Sections::Overview do
+  subject { described_class.new(crime_application) }
+
+  let(:crime_application) do
+    instance_double(
+      CrimeApplication,
+      to_param: '12345',
+      status: status,
+      reference: 'LAA-1',
```
zheileman, line 11: We can use a numeric value (given reference/USN will be numeric) here, and then below in the expectation we check it has been casted to a string, but this is a minor thing.

zheileman (approved):
Nice, approving! Some minor comments, no need to address immediately.

PR #267, opened 2023-01-30
CRIMAP-255 Handling of SAML metadata/certificates

zheileman:
## Description of change
This is a small iteration on top of what we already had with regards to LAA Portal authentication, to be able to consume metadata files locally stored, in addition to remote metadata endpoints (one, or another, they are mutually exclusive).

This reduces the amount of certificates we need to expose as secrets, given the metadata files are not considered secrets and can be stored in the public repo (the certificate they contain is the public one).

On the other side, we still need to handle with care our own (SP - Service Provider) certificate pair (public/private). Not so much the public, but the private. But for the sake of been together, we store both as secrets.

A certificate has been generated to be used for samlmock and Portal DEV. Samlmock will not use it, and we don't have a deployed DEV env so this can be used for local tests against Portal DEV once we are federated.

NOTE: If we can fetch the metadata endpoints directly in the future, then we can remove the need to consume them locally, or leave it as a fallback alternative. At the moment it is not possible due to network policies but it's being looked after.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-255

## Notes for reviewer
Additional context and comments in this ticket:
https://dsdmoj.atlassian.net/browse/TA-1979

Everything should continue working as before, however now it is possible to define SP certificates via secrets, and consume local metadata files in addition to remote ones.


.env.development
```diff
@@ -18,13 +18,15 @@ DATABASE_URL=postgresql://postgres@localhost/laa-apply-for-criminal-legal-aid
...
 # LAA Portal SAML authentication metadata endpoint
-LAA_PORTAL_IDP_METADATA_URL=https://samlmock.dev.legalservices.gov.uk/metadata
+# or path to a locally-stored metadata file, one or the other
+# LAA_PORTAL_IDP_METADATA_URL=https://samlmock.dev.legalservices.gov.uk/metadata
+LAA_PORTAL_IDP_METADATA_FILE=config/laa_portal/metadata/samlmock.xml
 
-# If the LAA Portal metadata URL is not set,
-# the following config will be used instead
+# LAA Portal service provider (that's us) certificate,
+# only required for environments other than samlmock
```
willmcb, line 26: 👍

.gitattributes
```diff
@@ -1,4 +1,5 @@
 *secrets* filter=git-crypt diff=git-crypt
+certificates.yml filter=git-crypt diff=git-crypt
```
willmcb, line 2: 👍

app/lib/laa_portal_setup.rb
```diff
@@ -50,19 +55,24 @@ def metadata_url
...
 
+  # rubocop:disable Metrics/MethodLength
   def metadata_config
-    return manual_override_config if metadata_url.blank?
-
-    # An explicit timeout is set, as the gem parser does not have one,
-    # which means it hangs for a very long time if URL is not reachable.
-    Timeout.timeout(3) do
-      OneLogin::RubySaml::IdpMetadataParser.new.parse_remote_to_hash(metadata_url)
+    if metadata_url.present?
```
willmcb, line 70: This is really nice as it allows us to change the source and not break things.

willmcb (approved):
Nice one looks good to me - one non-blocking suggestion you might want to consider :)

app/lib/laa_portal_setup.rb
```diff
@@ -74,13 +84,19 @@ def metadata_config
...
+  def metadata_from_server
+    # An explicit timeout is set, as the gem parser does not have one,
+    # which means it hangs for a very long time if URL is not reachable.
+    Timeout.timeout(3) do
+      OneLogin::RubySaml::IdpMetadataParser.new.parse_remote_to_hash(metadata_url)
+    end
+  end
+
+  def metadata_from_file
+    OneLogin::RubySaml::IdpMetadataParser.new.parse_to_hash(
```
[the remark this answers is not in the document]
zheileman: Absolutely, will do that, it makes sense 👍

PR #269, opened 2023-01-30
Add additional secretRef to deployment template

zheileman:
## Description of change
Follow-up to previous PR #267.

The new secrets (portal certificates) are created but if we don't specify them in the deployment template, they are not being picked up 🤦‍♂️


willmcb (approved):
👍

PR #270, opened 2023-01-31
CRIMAP-267 Better handling of application statuses

zheileman:
## Description of change
This is mostly a refactor and tidy up of some code that no longer makes sense because we only have in progress applications locally, and anything else is remote (datastore).

Previously, we had still a few coupling of the `status` values to locally persisted records, this is no longer the case.

To be extremely explicit, the `enum` in the `CrimeApplication` model now only support `in_progress` values so it forbids trying to persist to local DB a record in any other status (to limit mistakes).

Implement (similar to the others we already had) a `Datastore::GetApplication` service object that encapsulates the call to the API and handling of errors (including not found).

Awaiting confirmation from service designer on what to do with `superseded` applications, but for now, and for the sake of this code, I'm treating them as "not found" which means showing a "nice" error page to the user (this is an existing error page).

Finally, fixed a bug in the `ValueObject` superclass where inquiry methods were polluting other value object subclasses namespace.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-267

Everything should behave as before, but datastore applications not found (fake an UUID in the URL, for instance) will produce as expected a human not found error page. Same (for now, unless decided otherwise) for superseded applications.

### DIFF
--- a/app/services/datastore/get_application.rb
+++ b/app/services/datastore/get_application.rb
@@ -21,7 +21,7 @@ module Datastore
     private
 
     def superseded?(app)
-      ApplicationStatus.new(app.fetch('status')).superseded?
+      ApplicationStatus.new(app.status).superseded?
     end
   end
 end
--- a/app/services/datastore/get_application.rb
+++ b/app/services/datastore/get_application.rb
@@ -1,0 +1,27 @@
+module Datastore
+  class GetApplication
+    attr_reader :application_id
+
+    def initialize(application_id)
+      @application_id = application_id
+    end
+
+    def call
+      app = DatastoreApi::Requests::GetApplication.new(
+        application_id:
+      ).call
+
+      raise DatastoreApi::Errors::NotFoundError if superseded?(app)
+
+      app
+    rescue DatastoreApi::Errors::NotFoundError
+      raise Errors::ApplicationNotFound
+    end
+
+    private
+
+    def superseded?(app)
+      ApplicationStatus.new(app.fetch('status')).superseded?
+    end
+  end
+end
--- a/app/value_objects/value_object.rb
+++ b/app/value_objects/value_object.rb
@@ -21,13 +21,19 @@ class ValueObject
   end
 
   class << self
+    # Define inquiry methods for each value in the value object,
+    # i.e. `#coffee?` returns true for value `:coffee`, false for `:tea`
     def inherited(subclass)
-      TracePoint.trace(:end) do
-        # Define inquiry methods for each value in the value object,
-        # i.e. `#coffee?` returns true for value `:coffee`, false for `:tea`
+      TracePoint.trace(:end) do |trace|
         # :nocov:
-        subclass.values.each do |value|
-          define_method("#{value}?") { value.eql?(self) }
+        if subclass == trace.self
+          subclass.const_set(
+            :INQUIRY_METHODS, subclass.values.map { |value| "#{value}?".to_sym }
+          )
+
+          subclass.values.each do |value|
+            subclass.define_method("#{value}?") { value.eql?(self) }
+          end
         end
         # :nocov:
       end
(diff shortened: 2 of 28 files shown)


kennyevil (approved):

app/value_objects/value_object.rb
```diff
@@ -21,13 +21,19 @@ def hash
...
-      TracePoint.trace(:end) do
-        # Define inquiry methods for each value in the value object,
-        # i.e. `#coffee?` returns true for value `:coffee`, false for `:tea`
+      TracePoint.trace(:end) do |trace|
         # :nocov:
-        subclass.values.each do |value|
-          define_method("#{value}?") { value.eql?(self) }
+        if subclass == trace.self
+          subclass.const_set(
+            :INQUIRY_METHODS, subclass.values.map { |value| "#{value}?".to_sym }
```
willmcb, line 31: Why do you set constants as well as methods?

spec/value_objects/value_object_spec.rb
```diff
@@ -77,15 +73,21 @@
...
 
   describe 'inquiry methods' do
-    let(:fruit_one) { Fruit.new(:apple) }
-    let(:fruit_two) { Fruit.new(:banana) }
+    let(:yes_answer) { YesNoAnswer.new(:yes) }
+    let(:no_answer)  { YesNoAnswer.new(:no) }
+
+    it 'has a constant with the inquiry methods' do
+      expect(
+        YesNoAnswer::INQUIRY_METHODS
```
willmcb, line 81: This is interesting to me - I like it but I'm struggling to understand the application other than in this test 🤔

willmcb (approved):
Nice one - could we have a 5 min chat just on the inquiry method stuff? :)

app/services/datastore/get_application.rb
```diff
@@ -0,0 +1,27 @@
...
+    def call
+      app = DatastoreApi::Requests::GetApplication.new(
+        application_id:
+      ).call
+
+      raise DatastoreApi::Errors::NotFoundError if superseded?(app)
+
+      app
+    rescue DatastoreApi::Errors::NotFoundError
+      raise Errors::ApplicationNotFound
```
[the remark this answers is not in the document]
zheileman: The `Errors::ApplicationNotFound` is also used for Apply database (in progress) applications so is a way to unify both exceptions (API and Apply DB) under the same one which is then taken care in the concern `app/controllers/concerns/error_handling.rb`

It also keeps the API exception classes concentrated in these service objects that encapsulate each of the API calls (we have a few ones of these).

app/value_objects/value_object.rb
zheileman: So the constant will return this (for ApplicationStatus, but other value-objects behave the same):

```ruby
> ApplicationStatus::INQUIRY_METHODS
=> [:in_progress?, :submitted?, :returned?, :superseded?]
```

And then we have those instance methods declared and returning true/false depending if the value is what you are asking it for. Example:

```ruby
> s = ApplicationStatus.new(:returned)
> s.returned?
=> true
> s.submitted?
=> false
> s.in_progress?
=> false
> s.superseded?
=> false
> s.foobar?
(irb):8:in `<main>': undefined method `foobar?' for #<ApplicationStatus:0x000000010dee1980 @value=:returned> (NoMethodError)
```

Before, the constant was no needed, but I've now found a use case where it came handy to have an array of all the inquiry methods for a given value-object. In this case it was to perform delegation of all those methods (and not having to add one by one, or risk forgetting if we ever add another status, etc.)

spec/value_objects/value_object_spec.rb
zheileman: I use it here (introduced in this PR):
https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/pull/270/files#diff-88c2fd3a91d8fa60265ec5797a508c95bc1d6a4eea137880bc79ccf37f7cf0f1R3

😅

PR #275, opened 2023-02-07
Crimap 46 dwp check

kennyevil:
## Description of change

Adds DWP benefit checker service and service unreachable page

## Link to relevant ticket

https://dsdmoj.atlassian.net/jira/software/projects/CRIMAP/boards/897?selectedIssue=CRIMAP-46
https://dsdmoj.atlassian.net/jira/software/projects/CRIMAP/boards/897?selectedIssue=CRIMAP-256

0 0 0 0_3000_applications_c92e37a6-a17b-458c-9157-d1a2efe43ff4_steps_client_retry_benefit_check

Get the settings from `config\secrets.yml` and add them to `.env.development.local`. Set `BC_USE_DEV_MOCK` to false.

Use the following credentials to get a positive result

```
last name: Walker'
date of birth: 10/1/1980
national insurance number:  JA293483A
```

Go to `/client/retry_benefit_check` to see the retry page


zheileman:
Hey @kennyevil, this PR is a bit on the large size, mainly because it contains what I think were 2 separate pieces of work. The DWP API call and the network/credentials/etc. unhappy path mini-journey.

Would be great if we could split this PR into 2 smaller ones although I will be trying to review some things here and there and will be testing in local too.

Once thing, I've seen 2 variables in the secrets.yml that are not really secrets: the endpoint, and a mock true/false flag. Can we move these to the `config_map` instead? In fact, the `BC_USE_DEV_MOCK` might be best to have a default in code to `false` unless variable is explicitly declared (for example in our `.local` files if need be), but likely we want that always to `false` in staging, and surely production, so is best if we don't have to declare it to false in each environment that we end up having.

.github/workflows/test-build-deploy.yml
```diff
@@ -18,6 +18,11 @@ jobs:
       RACK_ENV: test
       RAILS_ENV: test
       DATABASE_URL: "postgresql://postgres:postgres@127.0.0.1/laa-apply-for-criminal-legal-aid-test"
+      BC_LSC_SERVICE_NAME: ${{ secrets.BC_LSC_SERVICE_NAME }}
```
zheileman, line 21: Please don't do this. Tests should not be using secrets at all. And definitely we don't want to add these secrets to Github.
If the ENV variables are needed (ideally they should not, for instance, in code we can do `ENV.fetch("BC_LSC_SERVICE_NAME", nil)`) let's use dummy values.

app/services/dwp/benefit_check_service.rb
```diff
@@ -1,10 +1,71 @@
 module DWP
   class BenefitCheckService
+    BENEFIT_CHECKER_NAMESPACE = 'https://lsc.gov.uk/benefitchecker/service/1.0/API_1.0_Check'.freeze
+    USE_MOCK = ActiveModel::Type::Boolean.new.cast(ENV.fetch('BC_USE_DEV_MOCK', nil))
```
zheileman, line 4: We probably don't want to use constants for ENV variables. Main reason is they are more difficult to mock in tests.
Better to use just a method that encapsulates the ENV.fetch (i.e. `#use_mock?`)

app/services/dwp/benefit_check_service.rb
```diff
@@ -1,10 +1,71 @@
...
+    REQUEST_TIMEOUT = 30.seconds
+
+    class ApiError < StandardError
+      include Nesty::NestedError
```
zheileman, line 8: Do we need this include?

app/services/dwp/benefit_check_service.rb
```diff
@@ -1,10 +1,71 @@
...
+      nil
+    end
+
+    private
+
+    attr_reader :applicant
+
+    def benefit_checker_params
+      {
+        clientReference: applicant.crime_application_id,
```
zheileman, line 46: What are the DWP requirements for this `clientReference` value? Out of curiosity. Maybe we can use the UUID of the applicant record instead.

app/views/steps/client/retry_benefit_check/edit.html.erb
```diff
@@ -0,0 +1,30 @@
+<% title t('.page_title') %>
+<% step_header %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">
```
zheileman, line 6: `govuk-heading-xl`

app/views/steps/client/retry_benefit_check/edit.html.erb
```diff
@@ -0,0 +1,30 @@
...
+    <p class="govuk-body">
+      <%= t('.saved_answers') %>
+    </p>
+  </div>
+</div>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= step_form @form_object do |f| %>
+      <%= f.continue_button(primary: :try_again, secondary: :apply_in_eforms) %>
```
zheileman, line 27: As per my previous comment, I think no need for secondary submit, we can `nil` it and use a link styled as button for eForms CTA.

spec/spec_helper.rb
```diff
@@ -1,8 +1,22 @@
 ENV['RAILS_ENV'] ||= 'test'
 
 require 'simplecov'
+require 'vcr'
```
zheileman, line 4: Would be great to remove the VCR dependency, as we are not using it anywhere else and most of the time it is just an overkill and difficult to manage.
In this case I can see there are just a couple fixtures, maybe 3 needed. We can save them as fixtures and in the tests use Webmock to mock the requests and responses.

app/controllers/steps/client/retry_benefit_check_controller.rb
```diff
@@ -0,0 +1,17 @@
+module Steps
+  module Client
+    class RetryBenefitCheckController < Steps::ClientStepController
```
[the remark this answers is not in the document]
zheileman: Check the styling on any of the pages with a green and a grey button next to each other and we can replicate it here to look the same. I think it may just need a wrapper `<div class="govuk-button-group">`

Gemfile
```diff
@@ -66,6 +72,7 @@ group :test do
   gem 'rubocop-rspec', require: false
   gem 'selenium-webdriver'
   gem 'simplecov', require: false
+  gem 'vcr'
```
willmcb, line 75: I am not keen on this gem - it creates an overhead that we don't need at the moment I think. Rerecording of cassettes etc can be a real headache.

app/services/dwp/benefit_check_service.rb
willmcb: What is the benefit of this gem over native errors? Is it a dependency that we really need? I don't think so given that it has three contributors/maintainers and was last updated in 2013... so I'm going to make a call here and say that we don't use this. Please use Ruby native errors.

https://github.com/skorks/nesty for reference

spec/spec_helper.rb
willmcb: +1

app/services/dwp/benefit_check_service.rb
```diff
@@ -4,10 +4,6 @@ class BenefitCheckService
     USE_MOCK = ActiveModel::Type::Boolean.new.cast(ENV.fetch('BC_USE_DEV_MOCK', nil))
     REQUEST_TIMEOUT = 30.seconds
 
-    class ApiError < StandardError
```
zheileman: I would say, even if get rid of Nesty, we still want to have a custom exception class to raise when rescuing from `rescue Savon::SOAPFault`. Maybe just remove nesty, leaving the rest of the code?

spec/services/dwp/benefit_check_service_spec.rb
```diff
@@ -1,35 +1,131 @@
...
+      hash_including(
+        message: hash_including(
+          clientReference: applicant.id,
+          surname: applicant.last_name.strip.upcase,
+          dateOfBirth: applicant.date_of_birth.strftime('%Y%m%d'),
+        ),
+      )
+    end
+
+    let(:benefit_checker_fixture) { File.read('spec/fixtures/benefit_checker_responses/successful.xml') }
```
zheileman, line 43: If we place the fixtures inside the existing `spec/fixtures/files` then we can use the helper `file_fixture('benefit_checker_responses/successful.xml').read`
And we keep it tidy and organised. There are already fixtures for address lookups.

spec/services/dwp/benefit_check_service_spec.rb
```diff
@@ -1,35 +1,131 @@
...
       last_name:,
       date_of_birth:,
-      nino:
+      nino:,
     )
   end
 
+  before do
+    Rails.configuration.x.benefit_checker.use_mock = false
+    Rails.configuration.x.benefit_checker.wsdl_url = 'http://benefit-checker/?wsdl'
```
zheileman, line 24: We can add this to the `.env.test` file, akin to what we did with the datastore, and that way we ensure it is always mocked for tests:

```
# DWP check endpoint (stubbed in tests)
BC_WSDL_URL=http://benefit-checker/?wsdl
```

I'm not a big fan of changing `Rails.configuration.x` values in the tests. These values will persist for other tests, and sometimes it produces flaky/intermittent tests difficult to diagnose.

As a suggestion, as you are initialising `BenefitCheckService` with the config, you may be able to stub that config in tests, and avoid having to change the configuration globally.

zheileman (commented):
A couple minor comments for you to consider. I think this is almost ready to go 🥳

zheileman (approved):
🥳

PR #277, opened 2023-02-13
Prepare Apply to use datastore JWT auth

zheileman:
## Description of change
Merging this will not enable JWT or change any functionality on Apply, as it is not enabled yet (to enable, change `config.auth_type = :jwt`).

Once enabled, the datastore API client will make use of a little middleware introduced in the `simple-jwt-auth` gem to generate and send a JWT bearer token.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-273

## Notes for reviewer
If wanted to test locally, a corresponding PR has been raised in the datastore repo, in order to have an end to end, locally-testable demo.


timpeat (approved):
Very nice 👍

PR #280, opened 2023-02-21
CRIMAP-286 Configure logging and filtering of PII

zheileman:
## Description of change
Logs to stdout so other systems can route these logs to final destinations for viewing and long-term archival. For example this is needed to see the logs in Kibana.

Also, added PII parameters to the filter. These are also used by Sentry when reporting exceptions.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-286

On local environments this new log stream to stdout can seen and tested by running the app in production mode, or by using docker-compose.
In our k8s namespaces, logs will output to stdout and can be accessed in Kibana.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,8 @@ gem 'faraday', '~> 2.7'
 gem 'govuk_design_system_formbuilder', '~> 3.3.0'
 gem 'jbuilder', '~> 2.11.5'
 gem 'kaminari'
+gem 'lograge'
+gem 'logstash-event'
 gem 'pg', '~> 1.4'
 gem 'puma'
 gem 'rails', '~> 7.0.3'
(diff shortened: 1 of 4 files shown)


timpeat (approved):
Looks good. I did wonder if we should just enable for all environments so we become more familiar.....

zheileman:
@timpeat Lograge removes ActionView logging, which also includes rendering times for partials. I think in development these can be quite useful. Additionally, depending on your IDE, these logs are interactive with clickable links to go directly to those files. See below.

The JSON one-liner is suitable for Kibana and similar but is not very human-friendly IMO, thus why I think is best to use it only in production.

Screenshot 2023-02-21 at 11 50 41

PR #282, opened 2023-02-23
(test) Portal DEV integration (SP-initiated)

zheileman:
## Description of change
This is just a test. We will not be using Portal DEV on staging, however in conversations with PCUAM team we want to try if SP-initiated auth works in a cloud environment as opposite to localhost.

I may try this twice, first with `idp_sso_service_binding: :redirect` followed with a second test with `idp_sso_service_binding: :post`.


willmcb (approved):
💥

PR #286, opened 2023-02-27
Offence suggestions a11y tweaks

zheileman:
## Description of change
This minor improvement will show in the "Suggestions" dropdown header, the number of offences returned. This is handy for non visually impaired users, as it may not be immediately evident to some that the suggestions box is scrollable when more than 5 or 6 offences are returned.

Additionally, it is even more important for visually impaired users, using screen readers, as the number of suggestions will be announced whenever there are any changes (polite so it waits for any other text to be read first).

Tested with VoiceOver on Mac, further testing or tweaks might be needed, these may come up as part of the accessibility audit.

NOTE: I've marked the code changes, as with other tweaks we've made to this component, with `[change]`, so it is easy to cherry pick these in case the component remote branch is updated.

### Before changes:
Screenshot 2023-02-27 at 09 24 35

### After changes:
Screenshot 2023-02-27 at 09 24 55

### DIFF
--- a/app/javascript/local/suggestions.js
+++ b/app/javascript/local/suggestions.js
@@ -26,6 +26,11 @@ Input.prototype.init = function () {
     this.$suggestionsHeader.textContent = this.$module.getAttribute('data-suggestions-header') || 'Suggestions'
     this.$suggestionsHeader.hidden = true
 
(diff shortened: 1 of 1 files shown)


timpeat (approved):
Nice s6n. Lovely PR :)

PR #290, opened 2023-02-27
CRIMAP-172 Use non-cluster specific domain

zheileman:
## Description of change
As per documentation:

https://user-guide.cloud-platform.service.justice.gov.uk/documentation/other-topics/migrating-from-live-1-domain-name.html

This is part of a broader ticket to use a custom domain for staging, however, for now this makes the service more resilient in case of cluster changes, until we have a production namespace and we set the proper final (and public) domains.

Old `*.apps.live` domain redirects to new domain to avoid breaking bookmarks etc.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-172

Both domains should work. If using old one (the `live` one) it will redirect to the non-cluster one (as in, no `live` in the URL).


arthurashman (approved):
👍

PR #294, opened 2023-03-02
CRIMAP-259 Show provider email in header

zheileman:
## Description of change
Also there is some work in preparation for CRIMAP-261, to allow or disallow providers using our service unless they've been onboarded to our private MVP.

In this PR the rules based on office code are not implemented yet, but I've implemented some boilerplate code to prepare for this, and to stop providers without LAA accounts (office codes) from using our service.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-259
https://dsdmoj.atlassian.net/browse/CRIMAP-261

This can be tested locally by checking out the branch, using the SAMLmock, and signing in with user "ABRA-S" who doesn't have any office codes. On staging I don't know how likely this scenario would be, probably very unlikely, but as some degree of gatekeeping is necessary, this is a fair check.
Email of the signed in provider is shown in the header, instead of the UID, as per service designer request.

### DIFF
--- a/app/controllers/providers/omniauth_callbacks_controller.rb
+++ b/app/controllers/providers/omniauth_callbacks_controller.rb
@@ -1,6 +1,7 @@
 module Providers
   class OmniauthCallbacksController < Devise::OmniauthCallbacksController
     skip_before_action :verify_authenticity_token
+    before_action :check_provider_is_allowed, only: [:saml]
 
     def saml
       provider = Provider.from_omniauth(auth_hash)
@@ -22,6 +23,19 @@ module Providers
 
     def auth_hash
       request.env['omniauth.auth']
+    end
+
+    def check_provider_is_allowed
+      gatekeeper = ProviderGatekeeper.new(auth_hash.info)
+
+      return if gatekeeper.access_allowed?
+
+      set_flash_message(
+        :alert, gatekeeper.reason,
+        scope: [:devise, :omniauth_callbacks, :provider_gatekeeper],
+      )
+
+      redirect_to root_path
     end
   end
 end
--- a/config/locales/en/devise.yml
+++ b/config/locales/en/devise.yml
@@ -15,6 +15,8 @@ en:
     omniauth_callbacks:
       failure: "Could not authenticate you from %{kind} because \"%{reason}\"."
       success: "Successfully authenticated from %{kind} account."
+      provider_gatekeeper:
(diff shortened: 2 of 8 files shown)


app/controllers/providers/omniauth_callbacks_controller.rb
```diff
@@ -23,5 +24,18 @@ def after_omniauth_failure_path_for(_)
     def auth_hash
       request.env['omniauth.auth']
     end
+
+    def check_provider_is_allowed
```
arthurashman, line 28: 'allowed' feels strange to me for some reason. Although I guess is clear. 
Just wondering if you'd consider 'authorised' or 'enrolled'

config/locales/en/devise.yml
```diff
@@ -15,6 +15,8 @@ en:
     omniauth_callbacks:
       failure: "Could not authenticate you from %{kind} because \"%{reason}\"."
       success: "Successfully authenticated from %{kind} account."
+      provider_gatekeeper:
+        no_office_codes: "Your account cannot use this service as it does not have any office codes."
```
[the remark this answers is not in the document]
zheileman: Hehe, yep it may be a bit obscure, but this error is only for when the user doesn't have any office accounts at all. This, to the best of my understanding, should not happen in production, but it is happening with some accounts (in SAMLmock for sure, maybe dev/staging envs too), rendering our service useless as applications need to always be linked to a provider's office account code.

The enrolment/authorisation I believe you are thinking about (the one based on office codes, firms (if we have all codes for a given firm), or even down to individual email addresses) will come next, it is not done in this PR, and will have different error message(s) 😄

app/controllers/providers/omniauth_callbacks_controller.rb
zheileman: The provider is technically authorised already at this point, as there is an additional (Devise) filter that handles the SAML response from Portal. That's why I didn't named it authorised. Just to avoid confusion!

Enrolment is a good one, I will be using that one when the rules based on office code or email are implemented 👍

arthurashman: Great ok, thanks for explaining!

config/locales/en/devise.yml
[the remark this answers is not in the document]
arthurashman: Right, yep, makes sense, thanks for explaining!!

PR #295, opened 2023-03-06
Refactor provider service objects

zheileman:
## Description of change
As we've now finalised (at least for the moment) the SAML authentication, and we are starting to work on the allow list, this PR does a little refactor to better encapsulate logic about the SAML attributes special handling (roles, office codes, email...) and puts all the service objects in their own `Providers` namespace.

Also now there is unit testing for the handling of these attributes, which wasn't explicitly tested before.

A separate PR will be raised to continue with the allow list and any other cleanups and refactors needed.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-261

## Notes for reviewer
No changes in functionality, only code refactor. It looks like a lot of changes because I renamed and moved classes to their own namespaces.


timpeat (approved):
Nice PR.

PR #297, opened 2023-03-06
CRIMAP-246 Provider's office account enrolment check

zheileman:
## Description of change
Implement an initial enrolment/allowlist based on office account codes.

The error page where the user is redirected is still TBD and copy and design will be provided once that has been agreed (there is a separate ticket).

Upon some comments in previous PR #294 I cleanup some of the previous gatekeeper code, and did some renaming too.

Note it may be we also want to allow by email address, but that has not been decided yet. If required, it is almost ready to implement, only caveat we would not want emails committed to the repo, so they might require hashing.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-246
https://dsdmoj.atlassian.net/browse/CRIMAP-303

## Notes for reviewer
Copy in the error page TBD. There is a separate ticket for design/content (CRIMAP-260)

This can be tested locally by using the SAMLmock and trying to sign in with user `ABRA-S`.
On staging it can be tested by trying to sign in with any user not in our list in confluence. I will update the page to add one specifically for exercising the unhappy path scenario.

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -15,6 +15,10 @@ class ErrorsController < ApplicationController
   end
 
   def unauthorized
+    respond_with_status(:unauthorized)
+  end
+
+  def not_enrolled
     respond_with_status(:forbidden)
   end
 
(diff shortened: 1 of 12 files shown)


timpeat (approved):
👍

PR #293, opened 2023-03-02
CRIMAP-300 Expose SP SLO url in metadata

zheileman:
## Description of change
This is a pre-requisite in order to test if Portal can trigger a SLO request to our service. The URL must be exposed in the metadata.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-300

## How to manually test the feature
After deploy, the metadata will expose the SLO url.

PR #298, opened 2023-03-07
CRIMAP-300 Tweaks for Portal SLO

zheileman:
## Description of change
This is a follow-up to PR #293 for the SLO investigations/spike.

Trying out `post` binding instead of `redirect` binding, as with redirect we get blocked with a 403 by cloudfront.

Also, do not blow up at the rack level if the endpoint can't process the SLO request (for example if it is called without a SAML payload), and show a nice page with a flash error instead (by bubbling up a re-raised `StandardError`).

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-300


zheileman:
@timpeat I ended up removing the method override for the time being. We can keep an eye to see if this becomes an issue, as we get Sentry notifications 👍

PR #300, opened 2023-03-08
CRIMAP-305 Fix govuk branded assets

zheileman:
## Description of change
Some assets coming from `govuk-frontend` were not being correctly referenced and part of the assets pipeline.

These mainly includes favicons of different sizes. Removed static favicon files in `public` directory as we don't want these to be used, but instead the govuk frontend ones.

Modernised the `govuk_template` and the references to these paths.

Additionally, set logging to stdout on production (this is separate to lograge that logs the requests and was already configured to use stdout), and add some additional logging in key areas.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-305

### DIFF
--- a/app/controllers/concerns/error_handling.rb
+++ b/app/controllers/concerns/error_handling.rb
@@ -14,7 +14,9 @@ module ErrorHandling
       else
         raise if Rails.application.config.consider_all_requests_local
 
+        Rails.logger.error(exception)
         Sentry.capture_exception(exception)
+
         redirect_to unhandled_errors_path
       end
     end
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -27,8 +27,9 @@ Rails.application.configure do
   # Compress CSS using a preprocessor.
   # config.assets.css_compressor = :sass
 
-  # Do not fallback to assets pipeline if a precompiled asset is missed.
-  config.assets.compile = false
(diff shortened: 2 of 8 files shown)


app/controllers/concerns/error_handling.rb
```diff
@@ -14,7 +14,9 @@ module ErrorHandling
       else
         raise if Rails.application.config.consider_all_requests_local
 
+        Rails.logger.error(exception)
```
willmcb, line 17: 👍

config/environments/production.rb
```diff
@@ -27,8 +27,9 @@
   # Compress CSS using a preprocessor.
   # config.assets.css_compressor = :sass
 
-  # Do not fallback to assets pipeline if a precompiled asset is missed.
-  config.assets.compile = false
+  # We are pre-compiling assets, so on production these will always be present, however
+  # for this to work this flag needs to be `true`. Seems like a Rails bug to me.
```
willmcb, line 31: Will probably apply to review in the same way - thanks for flagging 👍

willmcb (approved):
👍

config/environments/production.rb
zheileman: Yep I believe you may have the same issue as we both are using the same assets precompile process and pipeline.

PR #302, opened 2023-03-09
deps: Bundle update

zheileman:
## Description of change
Update gems to latest minor versions. No breaking changes. Fixes one new rubocop offence.

Also lock laa-criminal-legal-aid-schemas to a tag, to avoid introducing changes inadvertently.

Closes #301.


willmcb (approved):
👍

PR #305, opened 2023-03-13
Update client details task completed rules

zheileman:
## Description of change
Originally we had the `telephone_number` as a mandatory field, but some time ago we decided to make it optional, thus we don't have to check for presence in order to consider the task "completed", as `telephone_number` can be left blank.

Only `correspondence_address_type` is mandatory now.

Made the rules for the correspondence address more stringent by checking we have indeed the address record filled when `home address` or `other address` are selected.

### Before changes:
Screenshot 2023-03-13 at 10 36 15

### After changes:
Screenshot 2023-03-13 at 10 36 27

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -10,6 +10,10 @@ class Person < ApplicationRecord
   scope :with_name, -> { where.not(first_name: [nil, '']) }
 
   def home_address?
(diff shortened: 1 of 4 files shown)


timpeat (approved):
👍

PR #307, opened 2023-03-15
Enable omniauth test mode temporarily

zheileman:
## Description of change
The accessibility audit team needs testing on mobile devices, however the VPN is quite tricky to setup on mobile, thus temporarily we are going to disable LAA Portal auth and fake it with the omniauth test mode.


timpeat (approved):
👍

PR #309, opened 2023-03-16
CRIMAP-308 Fix offence dates validation

zheileman:
## Description of change
On the offences page, the dates are nested attributes. On submit these nested attributes were assigned to the `record` (`Charge` model) with `record.offence_dates_attributes=(params)`.

We use `accepts_nested_attributes_for :offence_dates` in `Charge` for this.

This works perfectly fine when the dates are valid. However if any of the dates is invalid, for instance month 13, or day 38, it blows up with `ActiveRecord::MultiparameterAssignmentErrors` because Rails internally was instantiating Date objects from the params, and `Date.new(2022, 13, 5)` is invalid.

What we want is to give a meaningful error to the user, and style as "red" the field giving the error, and show the error summary.

In order to do this we have to change the approach, instead of:
1. Assign the dates params (which instantiate Dates and then the `OffenceDate` objects).
2. Perform validation on the new instantiated (not yet persisted) `OffenceDate` records.
3. Persist if all was good.

We now do:
1. Instantiate from the params the OffenceDate objects
2. Perform validation on these with our `multiparam_date` validator
3. If all good persist the record and their offence dates.
4. If error, re-render the page showing the errors.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-308

Screenshot 2023-03-15 at 17 48 09
Screenshot 2023-03-15 at 17 48 30

### DIFF
--- a/app/forms/steps/case/charges_form.rb
+++ b/app/forms/steps/case/charges_form.rb
@@ -4,15 +4,14 @@ module Steps
       attribute :offence_name, :string
       validates :offence_name, presence: true
 
-      delegate :offence_dates_attributes=, to: :record
+      # transient attribute
+      attr_accessor :offence_dates_attributes
 
       validates_with ChargesValidator, unless: :any_marked_for_destruction?
 
       def offence_dates
-        @offence_dates ||= record.offence_dates.map do |offence_date|
-          OffenceDateFieldsetForm.build(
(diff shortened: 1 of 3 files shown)


timpeat (approved):
👍

PR #315, opened 2023-03-31
[CRIMAP-312] Change justification button text

mseedat-moj:
## Description of change
Alters button text to match figma

## Notes for reviewer
Note sure if button text 'Add justification information' is meant to be shown for all return types of just split case

### After changes:
Screenshot 2023-03-31 at 22 59 16

Create a new application in Apply

In Review return the application with the split case reason

In Apply look for 'returned applications' and select the originally submitted application.

### DIFF
--- a/app/views/completed_applications/_return_notification.html.erb
+++ b/app/views/completed_applications/_return_notification.html.erb
@@ -18,6 +18,6 @@
 
     <%= button_to recreate_completed_crime_application_path, method: :put,
                   class: 'govuk-button app-button--blue govuk-!-margin-bottom-2 app-no-print',
-                  data: { module: 'govuk-button' } do; t('.update_application'); end %>
+                  data: { module: 'govuk-button' } do; t('.add_justification'); end %>
   </div>
 </div>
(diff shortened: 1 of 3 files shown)


willmcb (approved):

app/views/completed_applications/_return_notification.html.erb
```diff
@@ -18,6 +18,6 @@
 
     <%= button_to recreate_completed_crime_application_path, method: :put,
                   class: 'govuk-button app-button--blue govuk-!-margin-bottom-2 app-no-print',
-                  data: { module: 'govuk-button' } do; t('.update_application'); end %>
+                  data: { module: 'govuk-button' } do; t('.add_justification'); end %>
```
arthurashman, line 21: So this should only be 'add_justification' if the return reason is add justification. Otherwise, for all other return reasons, update_application is correct.

config/locales/en/dashboard.yml
```diff
@@ -73,7 +73,7 @@ en:
         case_concluded: because the case has already concluded
         provider_request: at your request
         split_case: because the case has now been split and we require IOJ justification for all offences
-      update_application: Update application
```
arthurashman, line 76: We need both of these

arthurashman (changes requested):
We need both versions of the copy, one for the add justification return_reason, and the other for all other return_reasons

mseedat-moj:
Closing this PR as it makes more sense to have the required changes in #314 (CRIMAP-311)

PR #317, opened 2023-04-11
Fix routing for u18 post client details

arthurashman:
Currently all applicants route from client details to NINO u18 applicants are passported and do not need a passporting benefit therefore do not need a NINO to run a DWP check
This PR changes the routing for u18 applicants from client details to address lookup
u18 passporting feature flag turned on for test env

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-317

## Notes for reviewer
Feature flag for u18 passporting is on in the test env

Enter a dob on client details page that means a user is u18

### DIFF
--- a/app/services/age_calculator.rb
+++ b/app/services/age_calculator.rb
@@ -1,0 +1,12 @@
+class AgeCalculator
+  attr_reader :crime_application
+
+  def initialize(crime_application)
+    @crime_application = crime_application
+  end
+
+  def applicant_under18?
+    dob = crime_application.applicant.date_of_birth
+    dob + 18.years > Time.zone.today
+  end
+end
--- a/app/services/age_calculator.rb
+++ b/app/services/age_calculator.rb
@@ -1,12 +1,12 @@
 class AgeCalculator
-  attr_reader :crime_application
+  attr_reader :person
 
-  def initialize(crime_application)
-    @crime_application = crime_application
+  def initialize(person)
+    @person = person
   end
 
-  def applicant_under18?
-    dob = crime_application.applicant.date_of_birth
+  def under18?
+    dob = person.date_of_birth
     dob + 18.years > Time.zone.today
   end
 end
--- a/app/services/decisions/client_decision_tree.rb
+++ b/app/services/decisions/client_decision_tree.rb
@@ -6,7 +6,7 @@ module Decisions
       when :has_partner
         after_has_partner
       when :details
-        edit(:has_nino)
+        after_client_details
       when :has_nino, :retry_benefit_check
         after_has_nino
       when :benefit_check_result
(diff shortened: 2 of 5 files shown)


config/settings.yml
```diff
@@ -5,6 +5,7 @@ feature_flags:
     production: false
   u18_ioj_passport:
     local: true
+    test: true
```
zheileman, line 8: No harm, but not needed, as in test environments all feature flags are enabled by default unless disabled explicitly (i.e. `test: false`)

app/services/decisions/client_decision_tree.rb
```diff
@@ -30,6 +30,14 @@ def after_has_partner
       end
     end
 
+    def after_client_details
+      if IojPassporter.new(form_object).applicant_under18_passport?
```
zheileman, line 34: I think the `IojPassporter` class is initialised with a crime application instead the form object. In this case as both objects respond to `#applicant` it works, but might be best to change it to avoid future problems 😄

zheileman (changes requested):
Added a couple minor comments.

I'm afraid there is another issue uncovered by this. The current schema has `nino` as a required attribute.

Also, perhaps not critical, but it feels like we are mixing 2 kinds of "passporting" here. One is DWP check, and the other is IoJ? and we are reusing the class intended for IoJ.

In my head it may make sense for the DWP bypass to be done in some of the DWP service objects, although that imply entering a NINO for an under 18s. Do all under 18s lack NINO? Maybe a separate discussion...

app/services/decisions/client_decision_tree.rb
arthurashman: 👍 Thank you!

config/settings.yml
arthurashman: 👍 Thank you. I thought it might have a true default,  I didn't know whether that was a known paradigm so decided to explicitly set it. I will revert

zheileman:
@arthurashman I've noticed another thing that might be affected by this. Unfortunately, probably due to some (in light of this, wrong, assumptions), there are some places were we rely on the NINO being up until now a mandatory field. For instance in the task list if I remember correctly, to enable the following section (case details).
Might be other places. We need to revisit those places.

arthurashman:
@zheileman I have pulled out the age calculation into it's own service. I guessed that could be useful for calculating the age of the applicant <16, or if there is need to calculate ages of partners or codefendants in future too. I think that addresses your point, which I completely agree with - I don't know if we want a `MeansPassporter` service to make that even clearer still?

app/services/age_calculator.rb
```diff
@@ -0,0 +1,12 @@
+class AgeCalculator
```
zheileman, line 1: This is great, I think is more clear and as you mention might be differences depending on < 18 or < 16 so having this (even admittedly simple logic) in a separate class is cleaner.

Perhaps only suggestion, not important really, the class could be initialised with a person (in this case applicant) record directly instead of the whole crime application?
That way is easier to calculate dates for other `Person` subclasses like perhaps, if needed in the future, partners. As we don't couple the class to `crime_application.applicant`

arthurashman: 👍 I didn't think that through, thinking I didn't want to initialise with `applicant` - but yes, 100% thank you

zheileman (approved):

PR #319, opened 2023-04-13
CRIMAP-312 Implement split case IoJ passporting logic

zheileman:
## Description of change
Split case passported applications are returned to the provider asking them to provide IoJ details.

This PR implements a new boolean flag in the `Ioj` table `passport_override` (false by default). This flag gets set on rehydration of returned application that have been returned due to split case.

In a few places along the application, changes has been made to use this flag so:

1. On a returned application, the "Review" page shows an "Justification for legal aid" section with a different copy and a change link (NOTE: copy to be confirmed with content designer, can be done in a separate PR).

2. If the provider clicks any other change link before the IOJ and goes through the application, the `passport_override` flag is used to show the IOJ page, even if they are passported.

3. The task list has been tweaked as the logic has changed slightly.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-312

## Notes for reviewer
On staging, unless we enable the `u18_ioj_passport` feature flag, this will not be fully testable. I've checked and we can enable this on staging, which I've done in this PR.

### Before changes:
Screenshot 2023-04-12 at 15 14 47

### After changes:
Screenshot 2023-04-13 at 12 51 04

On localhost it can be tested by submitting a passported (under 18s for now only) application without IOJ, and on review returning it with `split_case` reason. The returned application will let the Provider enter IOJ details despite still being technically passported (i.e. applicant is still under 18)

### DIFF
--- a/app/presenters/summary/sections/passport_justification_for_legal_aid.rb
+++ b/app/presenters/summary/sections/passport_justification_for_legal_aid.rb
@@ -10,13 +10,25 @@ module Summary
           Components::ValueAnswer.new(
             :passport, passport_triggered?
           ),
+          Components::ValueAnswer.new(
+            :passport_override, passport_override?,
+            change_path: edit_steps_case_ioj_path
+          ),
         ].select(&:show?)
       end
 
       private
 
+      def ioj
+        @ioj ||= crime_application.ioj
+      end
+
       def passport_triggered?
-        crime_application.ioj.blank? && crime_application.ioj_passport.any?
+        ioj.blank? && crime_application.ioj_passport.any?
+      end
+
+      def passport_override?
+        ioj.present? && ioj.types.blank? && ioj.passport_override
       end
     end
   end
--- a/app/presenters/tasks/ioj.rb
+++ b/app/presenters/tasks/ioj.rb
@@ -18,8 +18,16 @@ module Tasks
     end
 
     def completed?
-      crime_application.ioj_passport.any? ||
+      return true if ioj_passported?
+
+      crime_application.ioj.present? &&
         crime_application.ioj.types.any?
+    end
+
+    private
+
+    def ioj_passported?
+      IojPassporter.new(crime_application).call
     end
   end
 end
--- a/app/services/datastore/application_rehydration.rb
+++ b/app/services/datastore/application_rehydration.rb
@@ -26,12 +26,20 @@ module Datastore
       crime_application.applicant.present?
     end
 
+    def split_case?
+      parent.return_details.reason.inquiry.split_case?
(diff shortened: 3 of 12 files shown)


app/presenters/summary/sections/passport_justification_for_legal_aid.rb
```diff
@@ -10,13 +10,25 @@ def answers
...
+            :passport_override, passport_override?,
+            change_path: edit_steps_case_ioj_path
+          ),
         ].select(&:show?)
       end
 
       private
 
+      def ioj
+        @ioj ||= crime_application.ioj
```
willmcb, line 23: Nice :)

app/presenters/tasks/ioj.rb
```diff
@@ -18,8 +18,16 @@ def in_progress?
     end
 
     def completed?
-      crime_application.ioj_passport.any? ||
+      return true if ioj_passported?
+
+      crime_application.ioj.present? &&
```
willmcb, line 23: Was wondering if it might be worth putting the:
```ruby
def ioj
  @ioj ||= crime_application.ioj
end
```
method in the base presenter?

zheileman: In this instance, the base presenter is for the task list, and each task class affects mostly one record (applicant, case, ioj, etc.) so I think is best not to raise these to the base class and that way everything is more self-contained and easy to read the code.

Perhaps if these task classes become more complex then definitely it might be worth considering shifting some code to the base task class.

app/services/datastore/application_rehydration.rb
```diff
@@ -26,12 +26,20 @@ def already_recreated?
...
     end
 
     def ioj
-      Ioj.new(parent.ioj.serializable_hash) if parent.ioj.present?
+      if parent.ioj.present?
+        Ioj.new(parent.ioj.serializable_hash)
+      elsif split_case?
+        Ioj.new(passport_override: true)
+      end
     end
```
mseedat-moj, line 43: can `parent.ioj.present?` and `split_case?` both be false?

zheileman: Yep, possible, in fact is a very likely scenario, but it is covered:

`parent.ioj` will be `nil` if a submitted application has been IoJ passported, as the provider would not have entered any IoJ details then, and this application has been returned for any reason other than split case (like `clarification_required`).

In these scenarios, this `#ioj` method will return `nil` which is the expected behaviour, as we are re-creating the application and the original one didn't have IoJ.

willmcb (approved):
👍

PR #321, opened 2023-04-17
CRIMAP-331 Refactor IoJ and means passporters

zheileman:
## Description of change
This PR is a follow-up to some edge cases detected on previous PR #317.

Previously we had only one IoJ passporter. Now we have another passporter named `MeansPassporter` to better encapsulate the logic and meaning of means passporting.

Means passporting can be done on age (under 18s) or if over 18, then based on DWP check, that can result on success or fail.

Refactor some of the previous code, moving passporter classes into its own namespace `Passporting`. Removed the `AgeCalculator` and converted to a concern `PersonWithDob`.

Fixed the task list, that was still relying on NINO being a mandatory field (it is not anymore, as under 18s are exempt of DWP check and thus we skip the NINO page).

Minor DRY tweaks along the way.

NOTE: this is the first part of this work. In a separate PR, this new means passporting attribute will be exposed in the JSON for submission, so review can use the resulting values, and also Apply can read back these.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-331

There are no functional changes, this is mostly a refactor of existing code and addition of a new class to track means passporting.


app/services/passporting/ioj_passporter.rb
```diff
@@ -0,0 +1,28 @@
...
+      ioj_passport << IojPassportType::ON_AGE_UNDER18 if applicant_under18?
+
+      crime_application.update(ioj_passport:)
+
+      passported?
+    end
+
+    def passported?
+      # IoJ passporting can be overridden for applications returned
+      # back to the provider due to the case being split
```
arthurashman, line 14: Caseworkers shouldn't return a split application for an under 18 applicant, but just wondering if the logic works here if they do. The under 18 applicant should still be passported (and also that calculation should be from their age on application submission, not re running the passporter at time of resubmission)

willmcb (approved):
Looks great :)

app/services/passporting/ioj_passporter.rb
zheileman: Yes, the logic works whether the IoJ passporting was due to under 18 (in fact this is the only one that we can trigger on staging) or, soon, due to offence passporting.

More granular behaviour or edge cases like the one you mention, will be best done as separate PR(s), as this one is already quite large unfortunately.

I think the most likely solution is that the age check will be run only on new applications and not on resubmissions.

app/services/passporting/means_passporter.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    def passported?
+      crime_application.means_passport.any?
+    end
+
+    private
+
+    def benefit_check_passed?
+      !!applicant.passporting_benefit
```
arthurashman, line 20: Is this doing the same as what `.present?` would do? - would that be more readable?

app/services/decisions/client_decision_tree.rb
```diff
@@ -31,19 +31,19 @@ def after_has_partner
...
         edit(:has_nino)
       end
     end
 
     def after_has_nino
-      DWP::UpdateBenefitCheckResultService.call(current_crime_application.applicant)
+      DWP::UpdateBenefitCheckResultService.call(applicant)
 
-      if current_crime_application.applicant.passporting_benefit.nil?
+      if applicant.passporting_benefit.nil?
```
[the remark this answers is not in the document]
zheileman: Yes I agree. I didn't want to refactor too many things outside the (already quite big) scope of this PR, and it has grown too large already.
But as there are some remaining bits and pieces hanging around regarding to edge cases with passported returned applications, this could be done separately to improve readability of the code 👍

arthurashman (approved):
Very nice! Much clearer, and great refactoring

app/services/passporting/means_passporter.rb
zheileman: Yep, it is really the same. I will change this in the follow-up PR. I agree it seems more readable on booleans 👍 
On strings however is not the same:

```ruby
> !!""
=> true
> "".present?
=> false
```

arthurashman: Good to know, thank you!

PR #322, opened 2023-04-17
CRIMAP-331 Propagate `means_passport` to datastore

zheileman:
## Description of change
Second part of previous PR #321.

It bumps the schemas gem and uses the new attribute when reading back applications and on rehydration.

Removed some previous, hardcoded, unused attributes.

NOTE: DoB edge case is not yet deal with here (where a submitted application is under18 but on re-hydration applicant has become over18).

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-331

Submission of under 18 passported applications (NINO will be nil as we don't ask for it) should be successful. On the submitted application JSON, the `ioj_passport` should contain `['on_age_under18']` and the `means_passport` should contain ['on_age_under18'].
IoJ page is not shown on first run, but it is shown on returned applications due to split case.

### DIFF
--- a/app/services/adapters/structs/applicant.rb
+++ b/app/services/adapters/structs/applicant.rb
@@ -1,12 +1,6 @@
 module Adapters
   module Structs
     class Applicant < BaseStructAdapter
-      # For MVP this will always be `yes`
-      # This is not part of the application JSON for now.
-      def has_nino
-        YesNoAnswer::YES
-      end
-
       def home_address
         HomeAddress.new(super.attributes) if super
       end
@@ -18,7 +12,7 @@ module Adapters
       def serializable_hash(options = {})
         super(
           options.merge(
-            methods: [:has_nino]
+            methods: []
           )
         )
       end
--- a/app/services/adapters/structs/applicant.rb
+++ b/app/services/adapters/structs/applicant.rb
@@ -1,14 +1,7 @@
 module Adapters
   module Structs
     class Applicant < BaseStructAdapter
-      # NOTE: for MVP this is always true, as we are doing
-      # screening of applicants with DWP. This will change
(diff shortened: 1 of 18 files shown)


spec/presenters/summary/sections/client_details_spec.rb
```diff
@@ -23,6 +24,8 @@
     )
   end
 
+  let(:means_passport) { ['foobar'] }
```
mseedat-moj, line 27: What would a real-world value for 'means passport' look like?

app/services/adapters/structs/applicant.rb
```diff
@@ -1,14 +1,7 @@
...
-      # post-MVP but we don't know yet how exactly.
-      def passporting_benefit
-        true
-      end
-
-      # Same as above, for MVP will always be `yes`
+      # For MVP this will always be `yes`
       # This is not part of the application JSON for now.
       def has_nino
         YesNoAnswer::YES
```
arthurashman, line 7: Is this OK to leave hardcoded as Yes?

zheileman: It is not used really, but post-MVP we might use it. I can remove it to avoid confusion.
I think it is one of those attributes that can be inferred from other bits of information (like `applicant.nino.presence?`)

arthurashman (approved):
Looks good. All working as expected locally

app/services/adapters/structs/applicant.rb
arthurashman: 👍 Ok, I didn't mind either way! Just checking.

zheileman: I've removed it 😄

PR #323, opened 2023-04-18
CRIMAP-337 Deal with returned passported applications

zheileman:
## Description of change
One edge case not covered in previous PR was the fact, a passported application (mostly on the basis of under 18, but could also be on the basis of the DWP check) that is returned back to the provider, should keep the original passporting and not try to re-run the checks or take the provider through a different journey to the original one.

In practice this means:

- if original application was for an under 18 client, then the returned application will not ask for NINO nor run the DWP check, and remain passported upon submission.
- same for over 18 with a successful DWP check (in this case, the NINO page is shown to the provider, but the DWP check is not triggered).

This PR introduce some refactor to make this easier, with some syntax-sugar methods in a concern `Passportable`.

Also, introduced (although enabled on staging) a feature flag to enable/disable the means passport on under 18 (i.e. ask or not for NINO + DWP check). Just in case we need to turn this off.

Additional edge cases need to be studied and covered with code if required, as follow-up PRs. Also some changes might be required to the IoJ Passporter once we introduced the offence passporting.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-337

Submit a passported application (either on age or on DWP check) and return it. Child application should inherit parent passporting and thus should not ask for NINO or for DWP check (depending on the original passporting). Child application can also be submitted and again returned etc.

### DIFF
--- a/app/models/concerns/passportable.rb
+++ b/app/models/concerns/passportable.rb
@@ -1,0 +1,24 @@
+module Passportable
+  extend ActiveSupport::Concern
+
+  delegate :age_passported?,
+           :benefit_check_passported?, to: :means_passporter
+
+  def means_passported?
+    means_passporter.passported?
+  end
+
+  def ioj_passported?
+    ioj_passporter.passported?
+  end
+
+  private
+
+  def means_passporter
+    Passporting::MeansPassporter.new(self)
+  end
+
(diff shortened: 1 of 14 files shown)


willmcb (approved):
LGTM 👍

PR #324, opened 2023-04-19
CRIMAP-338 Handle resubmission date stamp

zheileman:
## Description of change
Non date-stampable case types should obtain a new date stamp on resubmission, instead of inheriting their parent's.

This fixes a bug observed on staging with resubmission of `Indictable` and `Already in Crown Court` application case types.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-338

Submit 2 applications. One application with a date-stampable case type, (for instance, Summary only) and another with a non date-stampable case type (for instance, Indictable).
Return both applications back to the provider.
Resubmit both applications. Date stampable one will retain original date stamp. Non date-stampable one will generate a new date stamp on submission (equal to submitted_at date).

### DIFF
--- a/app/services/datastore/application_rehydration.rb
+++ b/app/services/datastore/application_rehydration.rb
@@ -13,7 +13,7 @@ module Datastore
       crime_application.update!(
         client_has_partner: YesNoAnswer::NO,
         parent_id: parent.id,
-        date_stamp: parent.date_stamp,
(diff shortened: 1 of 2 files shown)


timpeat (approved):
Looks Great!

arthurashman (approved):
Ace, thank you!

PR #325, opened 2023-04-20
CRIMAP-328 Switch datastore URL to internal svc

zheileman:
## Description of change
We are evaluating internal cross-namespace networking between Apply/Review and Datastore.

For the time being and until we decide so, external ingress will still be available.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-328

In theory everything should continue working as before, just slightly less latency when loading lists or paginating (thus difficult to evaluate as we don't have that many records yet).


willmcb (approved):
💥

PR #327, opened 2023-04-24
Bump ruby version to 3.2.2

zheileman:
## Description of change
And other minor version gems update.

Use latest alpine version in Docker image.


.rubocop.yml
```diff
@@ -76,6 +76,9 @@ RSpec/ContextWording:
 RSpec/StubbedMock:
   Enabled: false
 
+RSpec/IndexedLet:
```
timpeat, line 79: Any reason for this?

timpeat (approved):
Small comment about the rubocop changes, but otherwise good to go.

.rubocop.yml
zheileman: Oh, yep! Sorry should I've added a comment to explain it. After bumping rubocop, they've introduced new cops, one of them triggers a few offences in our specs because it doesn't like lets named with numbers, like `date1`, `codefendant2`, etc. which we use in a few places.
I thought about renaming the lets but then I also thought it was clear enough as it was, numbered! 🤣 So for now I've disabled the cop.

arthurashman (approved):

PR #332, opened 2023-04-28
CRIMAP-172 Add staging subdomain to ingress

zheileman:
## Description of change
This is a subdomain of our production domain.
Follow-up to PR #331.

Internal cloud platform URLs are still available and functional, but this is now an easier to remember domain.
The cloud URLs will redirect to their corresponding "nice" domains.

Some functionality in Portal will still redirect to old cloud URL until metadata is updated on their side.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-172


willmcb (approved):
Nice 👍

PR #335, opened 2023-05-03
CRIMAP-346 Under 18 IoJ on returned applications

zheileman:
## Description of change
This covers an edge case where the applicant is aged under 18 and thus is passported on IoJ on the first submission and then the application is returned by the caseworker when applicant is already over 18, or the applicant turns 18 before the application is resubmitted again.

In summary, if the application is a resubmission, we don't run again the age check, instead we use whatever result the initial submission had.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-346

Application remains IoJ passported on age, even when the applicant have turned 18 between the original application being submitted and the resubmitted application being submitted.

### DIFF
--- a/app/services/passporting/base_passporter.rb
+++ b/app/services/passporting/base_passporter.rb
@@ -16,6 +16,10 @@ module Passporting
     def passported?
       raise 'implement in subclasses'
     end
+
+    def passport_types_collection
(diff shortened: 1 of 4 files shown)


timpeat (approved):
Nice and clear 👍

PR #337, opened 2023-05-05
CRIMAP-355 Implement session lifespan

zheileman:
## Description of change
This is in addition to the existing idle session timeout mechanism. Implemented via a new custom Devise module `:reauthable`.

The session lifespan will sign out a user (provider), and force them to re-authenticate with Portal, after X minutes have passed from their last sign in, regardless of their activity.

It is configurable, so in localhost we can still have a longer or shorter period that can help with development by setting the `REAUTHENTICATE_AFTER_MINUTES` variable in the `.env.development.local` file.

Staging set to 48 hours. Production set to 24 hours. These are sensible defaults, can be tuned later on.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-355

On localhost, as per above instructions to set the ENV variable. On staging or prod, waiting for time to pass.

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -32,6 +32,10 @@ OMNIAUTH_TEST_MODE=true
 # or shorter timeout can be configured (default is 60 minutes)
 # SESSION_TIMEOUT_MINUTES=60
 
+# Time, in minutes, that a user's authentication session remains
+# active. When the lifetime is reached the session expires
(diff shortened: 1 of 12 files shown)


willmcb (approved):
Looks good 👍

PR #338, opened 2023-05-09
CRIMAP-146 Prepare for offence IoJ passporting

zheileman:
## Description of change
Still awaiting final CSV file that will mark the offences that are "passportable". This PR just assumes a new column, named `ioj_passport` and for now it is `nil` for all offences, thus none is passportable.

The logic however is implemented in this PR and also a feature flag to globally (instead of at the offence level) can enable/disable this IoJ passporting.

I've not wanted to make changes to our current CSV file because it will have to change soon again, probably the columns too, as we are splitting multi-class offences.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-146

User-facing there are no changes with this PR as everything is backend and no offences are passportable yet (CSV file needs to be updated). Can be tested by doctoring the CSV locally. Specs also cover the expected logic.

### DIFF
--- a/spec/services/passporting/ioj_passporter_spec.rb
+++ b/spec/services/passporting/ioj_passporter_spec.rb
@@ -3,12 +3,14 @@ require 'rails_helper'
 RSpec.describe Passporting::IojPassporter do
   subject { described_class.new(crime_application) }
 
-  let(:crime_application) { instance_double(CrimeApplication, applicant:, ioj:, parent_id:) }
+  let(:crime_application) { instance_double(CrimeApplication, applicant:, case:, ioj:, parent_id:) }
   let(:applicant) { instance_double(Applicant, under18?: under18) }
+  let(:case) { instance_double(Case, charges:) }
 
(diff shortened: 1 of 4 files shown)


arthurashman (approved):
Nice! Looks good to me

spec/services/passporting/ioj_passporter_spec.rb
```diff
@@ -118,4 +120,50 @@
...
+      end
+
+      context 'when there are no passported offences' do
+        let(:charges) { [non_listed_charge, non_passported_charge] }
+
+        it { expect(subject.offence_passported?).to be(false) }
+      end
+    end
+    # rubocop:enable RSpec/MultipleMemoizedHelpers
+  end
```
arthurashman, line 168: Any use in adding a spec to test a combination of passported on age and passported on offence?

zheileman: Yep, good point. I'll add some coverage for that 👍

PR #342, opened 2023-05-12
CRIMAP-9 Add basic google analytics setup

zheileman:
## Description of change
This is the basic setup to track visits, drop outs, etc.

More complex analytics (to be able to measure interactions within the step forms) will come as a separate PR. Some custom javascript is required for these.

Also, no cookies consent yet, as that's part of a separate ticket.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-9

Traffic can be seen in the google analytics website.

### DIFF
--- a/app/helpers/analytics_helper.rb
+++ b/app/helpers/analytics_helper.rb
@@ -1,0 +1,5 @@
+module AnalyticsHelper
+  def analytics_tracking_id
+    Rails.configuration.x.analytics.ga_tracking_id
+  end
+end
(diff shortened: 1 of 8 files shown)


app/helpers/analytics_helper.rb
```diff
@@ -0,0 +1,5 @@
+module AnalyticsHelper
```
zheileman, line 1: This helper seems pointless just now, but will contain some more methods once the cookies consent is implemented.

PR #343, opened 2023-05-15
CRIMAP-9 Add some custom GA events

zheileman:
## Description of change
This will enhance the basic analytics, with some custom events triggered for certain actions throughout the application steps.

Radio options, checkboxes and dates (just the year part) will be submitted as events on form submits.

Also, individual links can be annotated with the class `ga-pageLink` and a couple data attributes to also send an event. This is useful mostly for "action" links, for instance "change", "remove"... links. I've added a few examples but we can annotate more links if they seem useful.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-9

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,3 @@
+exclude_patterns:
+- "app/javascript/local/"
+- "spec/"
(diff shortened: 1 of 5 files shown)


willmcb (approved):
Nice looks great! Also good that we don't rely on jQuery so thanks for converting this over to plain JS :)

PR #347, opened 2023-05-18
Tweaks to header colour handling and smoke test

zheileman:
## Description of change
A few tweaks that makes it more robust. Follow-up to PR #346.

The CSS will in effect use the orange colour for any environment unless it is production (this is better than having to specify one by one the non-prod envs).

The phase tag text will also default to whatever is the name of the env, unless overridden by locales (as it is the case of `production` -> `alpha`). This is also better than having to specify one by one all the env names, as usually the phase tag just shows the same (`staging` -> `staging`, `local` -> `local`).

Added a simple smoke test to ensure the CSS classes are applied correctly and the phase tag text is also correct.

### DIFF
--- a/app/assets/stylesheets/local/custom.scss
+++ b/app/assets/stylesheets/local/custom.scss
@@ -15,14 +15,13 @@
 // Differentiate the header in non-production envs by
 // overriding the color of the header and phase tag
 .app-body {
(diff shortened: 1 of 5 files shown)


timpeat (approved):
👍

PR #349, opened 2023-05-19
CRIMAP-306 Cleanup deprecated SAMLmock code

zheileman:
## Description of change
Instead of SAMLmock, we can use Portal DEV during local development for a higher fidelity whenever we need to develop or debug anything related to Portal SAML authentication.

The SAMLmock was very low fidelity didn't behave like a real Portal SAML provider, only had 3 users, and these didn't return email addresses, among other things.

I will document in confluence how to obtain the certificates for DEV in order to encrypt/assert the payloads as this is needed (it wasn't really needed with SAMLmock thus the low fidelity).

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-306

There are no functional changes as this was only used for local development, unless the `OMNIAUTH_TEST_MODE` env variable was set to `true` which is the default in localhost.


MohammedHassanLatif93 (commented):
@zheileman approved 👍

PR #355, opened 2023-05-24
[a11y] No programmatic association between interactive controls and question

hiboabd:
## Description of change
The accessibility audit flagged instances of there being no programmatic association between input fields and their question. 

On the 'Clients details' page, context regarding who's details need to be entered is missing for screen reader users where screen readers only announce the label and the type of input, (e.g. 'Full name grouping' "First name edit blank").  In this instance, it is not necessarily clear that those details are about the client.

On the 'Offences summary' page, when inputting the offence of the client, users are given an opportunity to delete the offence. There is no programmatic association between the question 'Are you sure you want to delete this offence?' and the buttons, 'Yes, delete it' and 'No, do not delete it'. For blind and low vision screen reader users the content that is being deleted or not is not communicated in this instance and could therefore lead to confusion regarding the purpose of the input. 

This PR adds the required context.

## Link to relevant ticket
[CRIMAP-367](https://dsdmoj.atlassian.net/browse/CRIMAP-367)

### Before changes:
Screenshot 2023-05-24 at 11 26 05

### After changes:
Screenshot 2023-05-24 at 11 16 54

[CRIMAP-367]: https://dsdmoj.atlassian.net/browse/CRIMAP-367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/steps/case/charges/confirm_destroy.html.erb
+++ b/app/views/steps/case/charges/confirm_destroy.html.erb
@@ -3,7 +3,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl">
+    <h1 class="govuk-heading-xl" id='charges-confirm-destroy-desc'>
       <%=t '.heading' %>
     </h1>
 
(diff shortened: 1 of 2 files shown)


timpeat (approved):
Looks good. We have a few others on the manage users interface, but I don't think they need addressing yet.

PR #356, opened 2023-05-24
Allow bypass with an under 18 (developer tools)

zheileman:
## Description of change
Previously there was only one "bypass DWP", which picks a random "mock" user (dob, nino, surname) and creates the record.

However there was a mock under18 so randomly it could happen this under18 user was picked for bypass, leaving the underlying record with inconsistent data (like having NINO, or the `passporting_benefit` flag set to true), which is incorrect for under 18s as we don't ask for NINO nor perform DWP check for those.

Removed this mock user as is no longer needed (no DWP is performed for under 18 anyways) and added a new button to explicitly trigger a bypass of an under 18, which will set everything correctly and even advance a bit more, up to the case type step.

### Before changes:
Screenshot 2023-05-24 at 17 21 01

### After changes:
Screenshot 2023-05-24 at 17 20 48

### DIFF
--- a/app/controllers/developer_tools/crime_applications_controller.rb
+++ b/app/controllers/developer_tools/crime_applications_controller.rb
@@ -39,6 +39,34 @@ module DeveloperTools
       redirect_to edit_steps_client_benefit_check_result_path(crime_application)
     end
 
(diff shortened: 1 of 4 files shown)


mseedat-moj (approved):
👍

PR #359, opened 2023-05-26
CRIMAP-380 Update design system and form builder

zheileman:
## Description of change
Update GDS and MOJ design systems to their latest releases.

Update form builder to latest version 4.0.0
https://github.com/DFE-Digital/govuk-formbuilder/releases/tag/v4.0.0

This version finally incorporates the improvements to localisation in nested fields, so we can get rid of some superfluous code in the co-defendant and offence pages. https://github.com/DFE-Digital/govuk-formbuilder/pull/386

Also, a new `extra_letter_spacing` can be used in input texts like reference numbers, telephones, codes, etc. that makes it easier to read. I've added this flag where it makes sense in our service. https://github.com/alphagov/govuk-frontend/pull/2230

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-380

## Notes for reviewer
After checking out make sure to run `yarn` to pull the new versions of the node modules, and then `rails assets:clobber && rails dartsass:build` to ensure the new assets are recompiled, then `restart the app`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 ruby File.read('.ruby-version').chomp
 
 gem 'faraday', '~> 2.7'
-gem 'govuk_design_system_formbuilder', '~> 3.3.0'
+gem 'govuk_design_system_formbuilder', '~> 4.0.0'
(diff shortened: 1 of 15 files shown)


timpeat (approved):
🎁

PR #360, opened 2023-05-26
[a11y] Dynamic changes not announced to screen reader users

hiboabd:
## Description of change
The accessibility audit flagged instances throughout the user journey where changes that dynamically appear on screen are not communicated to screen reader users. These are fixed on the `'What is the case type'` and '`Reason for legal aid'` pages by adding aria-labels `aria-live` and `aria-label` that notify users of changes and overrides the visually presented text adding additional context to the user as to what is being requested

This pr also fixes an issue in a [previous PR](https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/pull/351) where newly added `aria-labelledby` attributes to delete buttons was subsequently only reading out the question and not the button text. This was replaced by an `aria-label` that will add further context as to what is being deleted

## Link to relevant ticket
[CRIMAP-366](https://dsdmoj.atlassian.net/browse/CRIMAP-366)

## Notes for reviewer
The ticket also includes [an instance of dynamic content not being announced on the offences list dropdown](https://uv2866-apply-for-criminal-legal-aid.uservisionaccessibility.co.uk/dynamic-changes-not-announced-to-screen-reader-users-high.html). This is when the number of suggested offences within the offences dropdown is inconsistently read out to the screen reader user. Through investigating, this appears to happen when a user continues to type letters but there is no change in the number of suggested results so it is not read out as there has been no update. However, no obvious solution could be found so this will be disclosed as an issue in the accessibility statement.

### Before changes:
Screenshot 2023-05-25 at 15 38 08

### After changes:
Screenshot 2023-05-26 at 11 30 47
Screenshot 2023-05-26 at 11 31 45
Screenshot 2023-05-26 at 11 48 10
Screenshot 2023-05-26 at 09 01 21

[CRIMAP-366]: https://dsdmoj.atlassian.net/browse/CRIMAP-366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/steps/case/case_type/edit.html.erb
+++ b/app/views/steps/case/case_type/edit.html.erb
@@ -18,7 +18,7 @@
 
             <%= f.govuk_radio_button :case_type, choice.value do %>
               <%= f.govuk_text_field :appeal_with_changes_maat_id, width: 'one-third', autocomplete: 'off' %>
-              <%= f.govuk_text_area :appeal_with_changes_details %>
+              <%= f.govuk_text_area :appeal_with_changes_details, 'aria-label': t('helpers.label.steps_case_case_type_form.appeal_with_changes_details_a11y') %>
             <% end %>
 
           <% else %>
--- a/app/views/steps/case/case_type/edit.html.erb
+++ b/app/views/steps/case/case_type/edit.html.erb
@@ -18,7 +18,10 @@
 
             <%= f.govuk_radio_button :case_type, choice.value do %>
               <%= f.govuk_text_field :appeal_with_changes_maat_id, width: 'one-third', autocomplete: 'off' %>
-              <%= f.govuk_text_area :appeal_with_changes_details, 'aria-label': t('helpers.label.steps_case_case_type_form.appeal_with_changes_details_a11y') %>
+              <%= f.govuk_text_area :appeal_with_changes_details,
+                                    'aria-live': 'polite',
+                                    'aria-label': t('helpers.label.steps_case_case_type_form.appeal_with_changes_details_a11y')
+                                 %>
             <% end %>
 
           <% else %>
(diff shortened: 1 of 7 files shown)


app/views/steps/case/case_type/edit.html.erb
```diff
@@ -20,7 +20,9 @@
             <%= f.govuk_radio_button :case_type, choice.value do %>
               <%= f.govuk_text_field :appeal_with_changes_maat_id, autocomplete: 'off', width: 'one-third',
                                      extra_letter_spacing: true %>
-              <%= f.govuk_text_area :appeal_with_changes_details %>
+              <%= f.govuk_text_area :appeal_with_changes_details,
+                                    'aria-live': 'polite',
+                                    'aria-label': t('helpers.label.steps_case_case_type_form.appeal_with_changes_details_a11y') %>
```
zheileman, line 25: We can use this, that way if the form object class changes name or module, this still will find the correct locale, without changes to the view:

`t("helpers.label.#{f.object_name}.appeal_with_changes_details_a11y")`

Give it a go and let me know if it doesn't work, I think I've done this in a few other places top of my head.

app/views/steps/case/ioj/edit.html.erb
```diff
@@ -10,7 +10,10 @@
         <% IojReasonType.values.each_with_index do |ioj_type, index| %>
           <span id="<%= ioj_type %>"></span>
           <%= f.govuk_check_box :types, ioj_type.to_s, link_errors: index.zero? do %>
-            <% f.govuk_text_area ioj_type.justification_field_name, label: { hidden: true } %>
+            <% f.govuk_text_area ioj_type.justification_field_name,
+                                 label: { hidden: true },
+                                 'aria-live': 'polite',
+                                 'aria-label': t('helpers.label.steps_case_ioj_form.ioj_justification_hint_a11y') %>
```
zheileman, line 16: Same here we can use `f.object_name`

zheileman (approved):
Nice. Just a non blocking comment!

PR #364, opened 2023-05-31
CRIMAP-373 Use new single-class offences list

zheileman:
## Description of change
And finalise some remaining work related to the IoJ passport based on offence, as this new list, in addition to be single-class, also introduce the new column `ioj_passport` (true / false).

Cleanup some old code related to multi-class (presenter, locales, tests, etc.)

Note, as a separate PR, I will add the new `passportable` attribute to the schemas, structs, etc. although there is no rush as even if it is submitted to the datastore, it is not read-back as we don't use it yet.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-146
https://dsdmoj.atlassian.net/browse/CRIMAP-373
https://dsdmoj.atlassian.net/browse/CRIMAP-377

### Before changes:
Screenshot 2023-05-31 at 15 46 49

### After changes:
Screenshot 2023-05-31 at 15 47 13

### DIFF
--- a/app/models/concerns/csv_queryable.rb
+++ b/app/models/concerns/csv_queryable.rb
@@ -3,6 +3,13 @@ require 'csv'
 module CsvQueryable
   extend ActiveSupport::Concern
 
+  CSV::Converters[:boolean] = lambda do |field|
+    case field
+    when /\Atrue\z/i  then true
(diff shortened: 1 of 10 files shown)


willmcb (approved):
Looks great 👍

PR #374, opened 2023-06-09
Update copy for accessibility page

arthurashman:
## Description of change
Update copy for accessibility page
## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-397
### After changes:
image

### DIFF
--- a/app/views/about/accessibility.en.html.erb
+++ b/app/views/about/accessibility.en.html.erb
@@ -17,25 +17,25 @@
 
     <p class="govuk-body">We’ve also made the website text as simple as possible to understand.</p>
 
-    <p class="govuk-body"><a aria-label="AbilityNet" href="https://mcmw.abilitynet.org.uk/">AbilityNet</a> has advice on making your device easier to use if you have a disability.</p>
+    <p class="govuk-body"><a href="https://mcmw.abilitynet.org.uk/">AbilityNet</a> has advice on making your device easier to use if you have a disability.</p>
 
     <h2 class="govuk-heading-m">How accessible this website is</h2>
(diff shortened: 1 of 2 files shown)


app/views/about/accessibility.en.html.erb
```diff
@@ -4,6 +4,41 @@
...
+    <p class="govuk-body">This website is fully compliant with levels A, AA and AAA of the Web Content Accessibility Guidelines – WCAG 2.1.</p>
+
+    <h2 class="govuk-heading-m">Reporting accessibility problems with this website</h2>
+    <p class="govuk-body">We’re always looking to improve the accessibility of this website. If you find any problems not listed on this page or think we’re not meeting accessibility requirements, contact us by email at <%= mail_to Settings.support_email, class: 'govuk-link--no-visited-state', 'aria-label': Settings.support_email %>.</p>
+
+    <h2 class="govuk-heading-m">Enforcement procedure</h2>
+    <p class="govuk-body">The Equality and Human Rights Commission (EHRC) is responsible for enforcing the Public Sector Bodies (Websites and Mobile Applications) (No. 2) Accessibility Regulations 2018 (the ‘accessibility regulations’).</p>
+    <p class="govuk-body">If you’re not happy with how we respond to your complaint, <a aria-label="contact the Equality Advisory and Support Service (EASS)" href="https://www.equalityadvisoryservice.com/">contact the Equality Advisory and Support Service (EASS)</a>.</p>
```
zheileman, line 30: Why the `aria-label` in all links?
I think is not necessary when the link copy is the same we want a screen reader to read.

zheileman (approved):
👍 Nice, just probably we don't need all the `aria-label` in links.

PR #380, opened 2023-06-14
CRIMAP-389 Copy changes to existing error pages

zheileman:
## Description of change
As per content designer. Note this is the first part of the work.

Also changed any references to "back to all applications" to "back to your applications", also as per content designer request.

Added a flash message (that shows on the home page) when the user signs out, this will be the only flash message we will have eventually, but the work to remove the other flash messages and convert to standalone error pages will be done in a separate PR so this one is not too large.

In addition to this, changed the behaviour of the "Start now" button on the home page, so if provider is not signed out, it goes directly to LAA Portal (instead of giving them an unauthorize error like before) and if they are already signed in, direct them to the dashboard. This also will hopefully reduce the amount of unhelpful 401 alerts we get.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-389

Screenshot 2023-06-14 at 10 29 15

All the error pages can be accessed by using their fixed URL, for instance:
`.../errors/unhandled`
`.../errors/invalid_session`
etc.

The sign out flash is shown in the screenshot.


app/views/errors/invalid_session.en.html.erb
```diff
@@ -0,0 +1,19 @@
+<% title 'We signed you out' %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      For your security, we signed you out
+    </h1>
+
+    <p class="govuk-body">
+      This is because you were inactive for <%= hours_or_minutes(Rails.configuration.x.session.timeout_in) %>.
```
arthurashman, line 10: Will this timeout_in value always be available? Is there any chance of it being nil and causing an error? i.e. do we need a fallback?

arthurashman (approved):

app/views/errors/invalid_session.en.html.erb
zheileman: It is always available, and there is already a fallback in case the ENV is not defined:
```ruby
config.x.session.timeout_in = ENV.fetch('SESSION_TIMEOUT_MINUTES', 60).to_i.minutes
```

PR #384, opened 2023-06-15
Handle user time zone (London) when presenting dates

zheileman:
## Description of change
The application by default is set to UTC time zone, same as Datastore and Review. In the database(s) we store dates in UTC across the board.

However in some places, the presented date/time should be in the corresponding user's timezone which by default for our service (and unlikely to change or be configurable) will be `London`.

In order to avoid repetition, a new view helper has been exposed, analogous to `#l()` named `#tz_l()` to localize dates in the time zone of the user. This is used in the views.

Also, although originally unrelated to this, ensure the date stamp and the submission date have both the same precision (`Time.current`)

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-409

### Before changes:
before

### After changes:
after

### DIFF
--- a/app/lib/refinements/localize_with_tz.rb
+++ b/app/lib/refinements/localize_with_tz.rb
@@ -1,0 +1,17 @@
+module Refinements
+  module LocalizeWithTz
+    DEFAULT_USER_TIME_ZONE = 'London'.freeze
+
+    def self.included(base)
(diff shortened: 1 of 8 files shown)


timpeat (approved):
Looks, great. I'm going to 'borrow' the helper for review :)

PR #386, opened 2023-06-16
[CRIMAP-408] Change wording for passporting

mseedat-moj:
## Description of change
Changes the output for "Justification for legal aid" section from `Passported` to `Not needed` - but only for standard applications. Still says 'Passported' for split case.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-408

### DIFF
--- a/config/locales/en/summary.yml
+++ b/config/locales/en/summary.yml
@@ -138,8 +138,8 @@ en:
 
       # BEGIN passport justification for legal aid section
       passport:
-        question: Passported
-        answers: 
+        question: Not needed
+        answers:
           'true': Details provided do not require further justification for legal aid
       passport_override:
         question: Passported


config/locales/en/summary.yml
```diff
@@ -138,8 +138,8 @@ en:
 
       # BEGIN passport justification for legal aid section
       passport:
-        question: Passported
-        answers: 
+        question: Not needed
+        answers:
           'true': Details provided do not require further justification for legal aid
```
zheileman, line 143: The wording here also require changes, according to content designer.

zheileman:
We need to confirm if this same wording applies to a returned split application. My feeling is, it should, as in also to be changed from `Passported` to `Not needed`.

zheileman (approved):

PR #388, opened 2023-06-19
CRIMAP-214 Reset home address

zheileman:
## Description of change
There was an edge case where, if the provider enters a home address, but then for whatever reason they retract from this and want to delete it, there is not easy way to do this.

If they select click the link `My client does not have a home address` the previous entered address would still be there and in the contact details page, they would still see the radio "Same as home address" in the correspondence address options.

This PR improves the situation, but cleaning up any previously entered home address if the user comes back to the postcode step and click the link (which is now a button, styled as a link, so it submits the form) `My client does not have a home address`.

It does not destroy the record as it is needed for the "back" navigation, it only nil all values of the address effectively deleting it.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-214

There are no visual changes, but internally, the link `My client does not have a home address` is now a submit button, styled as a link, so it perform form submit and handling of the reset is done at the controller/formobject level.

Screenshot 2023-06-19 at 10 35 37

Checkout, run `rails dartsass:build` so the new styles are properly compiled, restart server just in case. Go to the home address postcode step, enter a postcode and select an address, in the contact details you will see the "Same as home address" radio option.

Now go back to the postcode and click "My client does not have a home address", it will clean up entered address, and redirect to the contact details. No "Same as home address" radio option should appear. In the DB the home address attributes are all `nil`.


docker-compose.yml
```diff
@@ -22,6 +22,7 @@ services:
       RAILS_LOG_TO_STDOUT: "1"
       DATABASE_SSLMODE: disable
       DISABLE_HTTPS: "1"
+      BC_USE_DEV_MOCK: "true"
```
zheileman, line 25: Not really related to this work, but without this, the local docker image can't pass the DWP check so we can't do end-to-end tests using docker-compose. This has no effect on real envs, it is only for local machines.

arthurashman (commented):
LGTM, works as expected on local.

Good catch 👍

arthurashman (approved):

PR #391, opened 2023-06-20
Crimap 398 update privacy content

arthurashman:
## Description of change
Updated privacy policy as per https://docs.google.com/document/d/1fVLqKKDMeZ2gIWfR99ULjPWh6TFT-xNdm5f22rlFPUA/edit
Decision was made not to duplicate the LAA-wide privacy policy, but to link out to the help and info page where the pdf is hosted. 
Updated link text in the declaration to "privacy notice" which is the title on the help and info page to ensure that it is clear which pdf we are directing the user to. 
## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-398

### DIFF
--- a/app/views/about/privacy.en.html.erb
+++ b/app/views/about/privacy.en.html.erb
@@ -1,154 +1,163 @@
+<%# N.B. Update the updated at date at the bottom of this file when modifying this page %>
+
 <% title 'Privacy notice' %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl">Privacy notice</h1>
-    <h2 class="govuk-heading-l">Apply for Criminal Legal Aid Service</h2>
-    <p class="govuk-body">This service is provided by the
-      <%= link_to 'Legal Aid Agency (LAA)', 'https://www.gov.uk/government/organisations/legal-aid-agency/', rel: 'external' %>,
-      part of the Ministry of Justice (MoJ).
-    </p>
+    <div class="govuk-!-margin-bottom-9">
+      <h1 class="govuk-heading-xl">Privacy notice</h1>
+      <h2 class="govuk-heading-l">Apply for Criminal Legal Aid Service</h2>
+      <p class="govuk-body">This service is provided by the
+        <%= link_to 'Legal Aid Agency (LAA)', 'https://www.gov.uk/government/organisations/legal-aid-agency/', rel: 'external' %>,
+        part of the Ministry of Justice (MoJ).
+      </p>
 
-    <p class="govuk-body">This privacy notice covers how the digital service (“Apply for criminal legal aid”) collects personal data from providers who are submitting information on behalf of their clients.</p>
+      <p class="govuk-body">This privacy notice covers how the digital service (“Apply for criminal legal aid”) collects personal data from providers who are submitting information on behalf of their clients.</p>
 
-    <p class="govuk-body">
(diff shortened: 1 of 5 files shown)


mseedat-moj (approved):
Lgtm

app/views/about/privacy.en.html.erb
```diff
@@ -1,155 +1,154 @@
...
       </ul>
     </div>
-    <p class="govuk-body">We know how important it is to protect customers’ privacy and to comply with data protection laws. We will safeguard your personal data and will only disclose it where it is lawful to do so, or with your consent.</p>
 
-    <h2 class="govuk-heading-l">Types of personal data we process</h2>
-    <p class="govuk-body">We only process personal data that is relevant for the services we provide.</p>
-    <p class="govuk-body">The personal data which you've provided for this service will only be used for the purposes set out below.</p>
+    <p class="govuk-body">We make sure you cannot be identified by Google Analytics data. We do not pass any of your personal data to Google Analytics, and Google Analytics does not store your IP address.</p>
+    <p class="govuk-body">We will not combine analytics information with other data sets in a way that would identify who you are.</p>
+    <p class="govuk-body">Find out more about <a class="govuk-link" target="_blank" href="/cookies">how we use Google Analytics and other cookies</a>
```
zheileman, line 57: Can we use the rails path helper? Also better not to be open in new page/tab unless there is a very good reason for, and if so we need to say something like `(opens in new tab)`

zheileman (changes requested):
A few minor comments.

Also all over the place there are `target="_blank"`
I think unless there is a very good reason we shouldn't use that, and if we do, we have to mark all this links with `(open in new tab)`for accessibility reasons.

app/views/about/privacy.en.html.erb
```diff
@@ -1,155 +1,154 @@
...
+        <li>
+          your portal login and office account number provided by the LAA Portal service
+          (<a class=" govuk-link " rel=" external " target=" _blank " href="<%= Settings.portal_help_info_url %>">LAA Portal terms and conditions</a>)
+        </li>
+        <li>
+          your Internet Protocol (IP) address
+        </li>
+        <li>
+          details of which version of
+          <a class=" govuk-link " rel=" external " target=" _blank " href=" https: www.gov.uk support browsers ">web browser</a>
```
zheileman, line 33: There are extraneous spaces and the href is invalid here.

app/views/about/privacy.en.html.erb
```diff
@@ -1,155 +1,154 @@
...
       SK9 5AF<br>
-      Tel: 0303 123 1113<br>
-    </p>
-    <p class="govuk-body">
-      <a href='http://www.ico.org.uk'>ico.org.uk</a>
     </p>
+
+    <h2 class="govuk-heading-l">Changes to this policy</h2>
+
+    <p class="govuk-body">We may change this privacy policy. In that case, the ‘last updated’ date at the bottom of this page will also change. Any changes to this privacy policy will apply to you and your data immediately.</p>
```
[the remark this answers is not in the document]
arthurashman: Nope, have flagged

app/views/about/privacy.en.html.erb
arthurashman: I was following what was already there. V happy to change.

app/views/about/privacy.en.html.erb
arthurashman: One of the linters has done this and I didn't notice, sorry!

arthurashman:
Thanks @zheileman - sorry for wasting your time, I should have checked what the linters were doing

PR #392, opened 2023-06-20
CRIMAP-405 Appeal case types do not trigger IoJ passport

zheileman:
## Description of change
The appeal is a separate issue to the original proceedings trying the offence(s), and the defendant must demonstrate they have grounds for appeal by completing the IoJ reasons.

As a result, the offences on the appeal application cannot be IoJ passported either by offence or age.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-405

Where previously an application would have been IoJ passported, now selecting an appeal case type, will ignore the passporting (either under 18 or on offence). No other changes made.

### DIFF
--- a/app/services/passporting/ioj_passporter.rb
+++ b/app/services/passporting/ioj_passporter.rb
@@ -17,6 +17,9 @@ module Passporting
     end
 
     def age_passported?
+      # Appeal cases do not trigger IoJ passporting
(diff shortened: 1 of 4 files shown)


arthurashman (approved):
Very nice and neat 👍

PR #395, opened 2023-06-21
More precise `change` address link in Review page

zheileman:
## Description of change
Small QoL refinement.

We can direct the provider to a more precise page when they click the `change` link on an address in the Review application page.

- If the address has not been entered, on the Review page it shows as "None", and the change link will take the provider to the postcode lookup so they can start the address journey.

- If an address haas been selected via postcode lookup, the change link will take them back to the dropdown with the results, so they can either pick a different one, or go back to change the postcode and run again the lookup.

- If an address has been manually entered we take them directly to the details page where they can tweak any part of the address (line one, line two, city, country...)

### DIFF
--- a/app/presenters/summary/sections/contact_details.rb
+++ b/app/presenters/summary/sections/contact_details.rb
@@ -50,7 +50,18 @@ module Summary
       end
 
       def change_path(address)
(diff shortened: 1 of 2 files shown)


arthurashman (approved):
Nice, I like it 👍

PR #397, opened 2023-06-23
CRIMAP-379 Simplified prometheus exporter approach

zheileman:
## Description of change
As we are not running Puma in multi process mode (instead we can scale horizontally with more threads and pods), there is a simplified approach to prometheus exporter.

In this version, we don't need a separate process running in a separate pod (meaning no need for having another deployment manifest, or deal with hosts to connect client to server etc.)

The prometheus exporter server instead runs in each pod, in the rails application itself. This have the advantage, we can add custom metrics more easily as we have access the the full rails environment (models, activerecord, etc.) whereas if running the process separately, we don't have it, or we have to load it defeating the purpose.

Added active record and puma instrumentation. That's why the configuration lines have increased.

We can see how this approach works and if we consider it is easier, once we implement some custom metrics (next PR). It is relatively easy to change to previous approach.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-414

Can also be tested easily on local, by running `docker-compose up` and checking the `/metrics` endpoint (in port 9394) or by setting the local env variable `ENABLE_PROMETHEUS_EXPORTER` to true.
If you want some verbose output in the console, set `PROMETHEUS_EXPORTER_VERBOSE` to true.


arthurashman (approved):

arthurashman:
I remember we used this approach in a previous team - good plan. Worked well for us

PR #398, opened 2023-06-23
Updated from alpha to beta with survey link

MohammedHassanLatif93:
## Description of change
Updated phase banner from alpha to beta with survey link

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-416


config/locales/en/layouts.yml
```diff
@@ -2,9 +2,9 @@
 en:
   layouts:
     phase_banner:
-      text: This is an MVP. Some things may not work as expected.
+      text_html: This is a new service – your <a href="%{survey_url}">feedback</a> will help us to improve it
```
zheileman, line 5: In the footer link we added the rel="external"
Do you want to do the same here?

zheileman (approved):
Looks good, just a question, but Im approving.

config/locales/en/layouts.yml
arthurashman: Good idea, thanks @zheileman. @MohammedHassanLatif93 Let's take a look at this after our break

PR #399, opened 2023-06-23
Additional prometheus exporter tweaks

zheileman:
## Description of change
All working fine! Tiny tweaks:

There is no need for a custom `hostname` label as we get "for free" a "pod" label which is exactly what we need to identify each of the pod's metrics.

Also, exposed the port in the deployment. I think technically is not needed if we have a service mapping but just in case.


timpeat (approved):
Exciting.

PR #404, opened 2023-06-26
CRIMAP-410 Alter USN sequence start

zheileman:
## Description of change
Agreed with team to do this in all environments. New sequence start set to 10_000_000

Note fixtures are still on 6xxx although not really important. Probably we will only change the maat fixture.

Tweaked the prometheus exporter initialiser so it doesn't run on db migrate tasks, and others. We only want this to run on servers (maybe in the future on workers too if we had any).

Run migrations in localhost and create a new application. It should start in 10_000_000


timpeat (approved):
Nice

PR #407, opened 2023-06-26
CRIMAP-275 Remove unused DWP check param

zheileman:
## Description of change
Michael has confirmed the `dateOfAward` parameter is no longer needed, in any environment, so it is safe to be removed.

Make sure the `clientReference` can be traced back to one of our applications, so changed to the application UUID (because the applicant record UUID is not persisted anywhere post-submission).

Few other non-functional tweaks.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-275

Everything should keep working as before, no functional changes.


timpeat (approved):
👍

PR #417, opened 2023-06-30
Fix extraneous spacing in some offence names

zheileman:
## Description of change
I've let Chris know to also update the offences spreadsheet.

Added a couple additional common abbreviations/synonyms to speed up entering offences.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-426

ASB, will come up with `Breach of an anti-social behaviour order`
CBO, will come up with `Breach a criminal behaviour order`

### DIFF
--- a/config/data/offence_synonyms.yml
+++ b/config/data/offence_synonyms.yml
@@ -3,3 +3,7 @@ offence_synonyms:
   synonyms: [gbh]
 - regex: !ruby/regexp /section 18/i
   synonyms: [s18]
+- regex: !ruby/regexp /anti-social behaviour/i
+  synonyms: [asb]
+- regex: !ruby/regexp /criminal behaviour order/i
+  synonyms: [cbo]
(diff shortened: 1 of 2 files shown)


config/data/offence_synonyms.yml
```diff
@@ -3,3 +3,7 @@ offence_synonyms:
   synonyms: [gbh]
 - regex: !ruby/regexp /section 18/i
   synonyms: [s18]
+- regex: !ruby/regexp /anti-social behaviour/i
+  synonyms: [asb]
+- regex: !ruby/regexp /criminal behaviour order/i
+  synonyms: [cbo]
```
arthurashman, line 9: We have Section 4 and Section 4A that I imagine could well be S4 and S4A - worth adding, or researching that with providers?

Actual bodily harm commonly is ABH too - One of the offence names includes it, but one other unhelpfully doesn't

arthurashman (approved):

config/data/offence_synonyms.yml
zheileman: Oh nice, `ABH` I can add that one too in this PR.

Yep I think the best we could do is work with providers on what they usually refer as shortcuts or abbreviations, is quite difficult for us to know in advance.

arthurashman: Yeah, very difficult. I also imagine they don't know they do it! S18 is so normal to them they might not consider it an abbreviation

PR #418, opened 2023-06-30
Fix cookie banner JS module blocked by browsers

zheileman:
## Description of change
Some browsers come with built in extra privacy measures, and also some plugins can be used to block some "annoyances" like cookie banners.

Main issue was, the blocking was stopping other genuine modules from being properly loaded by the browser, making some functionality that relies on JS fail, like the offences list, or the print button.

Seems to be "fixed" by obfuscating the name of the JS module so it doesn't contain keywords that are likely to be blocked, like `cookie`, `banner` or `consent`.

Screenshot 2023-06-30 at 15 57 10

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -4,12 +4,12 @@
 import { initAll } from 'govuk-frontend'
 initAll()
 
-// Cookie banner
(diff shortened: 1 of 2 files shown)


arthurashman (approved):
Haha, this is such a funny workaround

PR #422, opened 2023-07-04
CRIMAP-427 First part of appeal details refactor

zheileman:
## Description of change
I'm worried the PRs might become too big so I'll try to split them up into (almost) functional smaller PRs.

In order to keep backward compatibility as much as possible with existing applications in datastore, I've done a few temporary workarounds that will be removed once all the work is finalised and the schemas/fixtures/datastore are also updated.

This first part cleanup the existing case type step, adds the new appeal details step, adapt the decision tree with the new logic, and also updates the serialisers and the summary presenters.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-427

## Notes for reviewer
Few things still to do, and separately there will be changes to schema/fixtures. Will be done next.

### Before changes:
Screenshot 2023-07-04 at 15 20 00

### After changes:
Screenshot 2023-07-04 at 15 30 19
Screenshot 2023-07-04 at 15 30 28

Go the case type page, select any appeal radio option, you will be directed to the appeal details page, validations should work, etc. And in the review application page, details should show depending if they've been entered or not.

### DIFF
--- a/app/forms/steps/case/appeal_details_form.rb
+++ b/app/forms/steps/case/appeal_details_form.rb
@@ -1,0 +1,34 @@
+module Steps
+  module Case
+    class AppealDetailsForm < Steps::BaseFormObject
+      include Steps::HasOneAssociation
+      has_one_association :case
+
+      attribute :appeal_lodged_date, :multiparam_date
+      attribute :appeal_with_changes_details, :string
+      attribute :appeal_maat_id, :string
+
+      validates :appeal_lodged_date,
+                presence: true, multiparam_date: true
+
+      validates :appeal_with_changes_details,
+                presence: true, if: -> { appeal_with_changes? }
+
+      def appeal_with_changes?
+        case_type.appeal_to_crown_court_with_changes?
+      end
+
+      private
+
+      def persist!
+        kase.update(
+          attributes
+        )
+      end
+
+      def case_type
+        CaseType.new(kase.case_type)
+      end
+    end
+  end
+end
--- a/config/locales/en/errors.yml
+++ b/config/locales/en/errors.yml
@@ -83,9 +83,18 @@ en:
           attributes:
             case_type:
               inclusion: Select a case type
+        steps/case/appeal_details_form:
+          attributes:
             appeal_with_changes_details:
-              blank: Add the details of what has changed
-              present: Details of what has changed are required for Appeal to Crown Court with changes to financial circumstances only
+              blank: Enter details of changes in financial circumstances
+            appeal_lodged_date:
+              blank: Enter the date the appeal was lodged
+              invalid: Enter a valid appeal lodged date
+              invalid_day: Enter a valid day
+              invalid_month: Enter a valid month
+              invalid_year: Enter a valid year
+              year_too_early: Date the appeal was lodged is too far in the past
+              future_not_allowed: Date the appeal was lodged must be today or in the past
         steps/case/has_codefendants_form:
           attributes:
             has_codefendants:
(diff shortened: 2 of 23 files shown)


app/forms/steps/case/appeal_details_form.rb
```diff
@@ -0,0 +1,34 @@
+module Steps
+  module Case
+    class AppealDetailsForm < Steps::BaseFormObject
+      include Steps::HasOneAssociation
+      has_one_association :case
+
+      attribute :appeal_lodged_date, :multiparam_date
+      attribute :appeal_with_changes_details, :string
+      attribute :appeal_maat_id, :string
```
zheileman, line 9: I'm not performing format validation on MAAT ID, but I believe they confirmed is a number, not alpha-numeric, so we can probably validate format (if entered, as it is optional).
Can do in follow-up PR.

app/forms/steps/case/appeal_details_form.rb
```diff
@@ -0,0 +1,34 @@
+module Steps
+  module Case
+    class AppealDetailsForm < Steps::BaseFormObject
+      include Steps::HasOneAssociation
+      has_one_association :case
+
+      attribute :appeal_lodged_date, :multiparam_date
```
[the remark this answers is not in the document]
arthurashman: Past tense makes more sense to me 👍

config/locales/en/errors.yml
```diff
@@ -83,9 +83,18 @@ en:
...
               inclusion: Select a case type
+        steps/case/appeal_details_form:
+          attributes:
             appeal_with_changes_details:
-              blank: Add the details of what has changed
-              present: Details of what has changed are required for Appeal to Crown Court with changes to financial circumstances only
+              blank: Enter details of changes in financial circumstances
+            appeal_lodged_date:
+              blank: Enter the date the appeal was lodged
+              invalid: Enter a valid appeal lodged date
```
arthurashman, line 92: Maybe just `Enter a valid date`?

config/locales/en/helpers.yml
```diff
@@ -64,7 +66,9 @@ en:
       steps_address_lookup_form:
         postcode: This must be a valid UK postcode. For example, SW1A 2AA.
       steps_case_urn_form:
-        urn: For example, ‘12 AB 3456789’.
+        urn: For example, ‘12 AB 3456789’
```
arthurashman, line 69: Is there a reason the . was removed?  Looks like it's still in figma

config/locales/en/errors.yml
zheileman: Yeah, I had my doubts, in some places we use just "enter a valid date", and in others like the hearing, we use "Enter a valid next hearing date".

In theory according to the design guidelines we should be explicit about what date it is, however given in this page (and other pages) there is only 1 date input (as opposite to the offences page for example) I agree "Enter a valid date" would be appropriate.
I can change it.

arthurashman (approved):
LGTM - working as expected on local

config/locales/en/helpers.yml
zheileman: Oh I think that was by mistake. Will re-add it.

zheileman: Hmm, now looking at it, perhaps we need to clarify the position and either remove all full stop from all the hint texts in the service, or add them where we are missing them 🤣 
Because we have a mix of style I believe. In the design system guide, they sometimes add them, sometimes don't, so it is a bit unclear!

PR #423, opened 2023-07-04
[a11y] Make offence input field accessible on ios

hiboabd:
## Description of change
The accessibility audit flagged an issue on the ‘What has your client been charged with’ page where the offence input field was not accessible on iOS using swipe gestures when voice over is active (see before changes video below).

This was fixed by setting the aria attributes on the offence input element that controls the listbox rather than the overall form container

## Link to relevant ticket
[CRIMAP-371](https://dsdmoj.atlassian.net/browse/CRIMAP-371)

## Notes for reviewer
Bare with my terrible swipe navigation skills! 

### Before changes:
https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/assets/51047911/d33ceaa2-c477-4693-a961-d2d3552e3ab9

### After changes:

https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/assets/51047911/8dcbf39a-2381-441b-8c09-ccb0d51b0057

[CRIMAP-371]: https://dsdmoj.atlassian.net/browse/CRIMAP-371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/local/suggestions.js
+++ b/app/javascript/local/suggestions.js
@@ -16,11 +16,6 @@ Input.prototype.init = function () {
   if (suggestionsSourceId) {
     this.suggestions = document.getElementById(suggestionsSourceId)
 
-    this.$formGroup.setAttribute('role', 'combobox')
-    this.$formGroup.setAttribute('aria-owns', this.$module.getAttribute('id'))
-    this.$formGroup.setAttribute('aria-haspopup', 'listbox')
-    this.$formGroup.setAttribute('aria-expanded', 'false')
-
     this.$suggestionsHeader = document.createElement('h2')
     this.$suggestionsHeader.setAttribute('class', 'govuk-input__suggestions-header')
     this.$suggestionsHeader.textContent = this.$module.getAttribute('data-suggestions-header') || 'Suggestions'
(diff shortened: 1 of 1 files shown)


app/javascript/local/suggestions.js
```diff
@@ -16,11 +16,6 @@ Input.prototype.init = function () {
   if (suggestionsSourceId) {
     this.suggestions = document.getElementById(suggestionsSourceId)
 
-    this.$formGroup.setAttribute('role', 'combobox')
-    this.$formGroup.setAttribute('aria-owns', this.$module.getAttribute('id'))
```
hiboabd, line 20: I removed aria-owns attribute as according to the [docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-owns), it doesn't need to set if the element is a child of the parent and in this case the input element is a child of the form container

zheileman: Oh interesting. It says:
> In older ARIA specs, this was aria-owns rather than aria-controls, so you may see aria-owns in older combobox implementations. The aria-owns in the code should be updated to aria-controls.

So yeah, I think removing the aria owns and leaving the aria-controls is correct in this case 👍

zheileman (approved):
Amazing! nice job. Pretty pesky issue! 🎉

app/javascript/local/suggestions.js
hiboabd: Ah perfect 👍

PR #425, opened 2023-07-05
CRIMAP-427 Validate format of previous MAAT ID

zheileman:
## Description of change
Simple validation to avoid fat fingers or other issues.

The attribute is optional so we still allow blanks. But if the attribute is filled, then we validate is a numeric 6-9 digits.

Ensure all existing regex in other places use `\z` (lowercase) instead of `\Z` (uppercase) as:
> \Z Matches the end of the string unless the string ends with a "\n", in which case it matches just before the "\n".

Which means by using \Z someone could sneak in a newline character and still pass validation.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-427

Screenshot 2023-07-05 at 13 46 57

Absense is allowed. If something is entered, only numeric values 6-9 digits length will pass validation.

### DIFF
--- a/config/locales/en/errors.yml
+++ b/config/locales/en/errors.yml
@@ -88,7 +88,7 @@ en:
             appeal_with_changes_details:
               blank: Enter details of changes in financial circumstances
             appeal_maat_id:
-              invalid: Enter a valid previous MAAT ID
+              invalid: Enter a valid MAAT ID
             appeal_lodged_date:
               blank: Enter the date the appeal was lodged
               invalid: Enter a valid date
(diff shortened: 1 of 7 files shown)


config/locales/en/errors.yml
```diff
@@ -87,6 +87,8 @@ en:
           attributes:
             appeal_with_changes_details:
               blank: Enter details of changes in financial circumstances
+            appeal_maat_id:
+              invalid: Enter a valid previous MAAT ID
```
arthurashman, line 91: I'd drop the word 'previous' makes it a little confusing.

willmcb (approved):
Nice looks good - I tested the regex myself just to be sure and it worked perfectly 👍

PR #426, opened 2023-07-06
CRIMAP-429 Refactor error reporting using Rails new API

zheileman:
## Description of change
This is a first part of at least a couple PRs intended to remove some technical debt by refactoring exception capture and error reporting using the modern Rails error reporting API introduced in Rails 7+

https://edgeguides.rubyonrails.org/error_reporting.html

Functionally it should be the same as before, but some boilerplate code is simplified.

Only thing I noticed is the Rails.error.handle, Rails.error.report etc methods do not log the exception, which makes sense, but without it, the exception is quietly hidden and particularly in localhost it can be misleading.

I've made a basic error subscriber just to perform the `Rails.logger.error(e)` so it behaves like before, adding one line of backtrace for additional context.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-429

In localhost is easy to test by not running the datastore, and going to the dashboard. It will log errors about not being able to connect to datastore.
You can also expose a fake `SENTRY_DSN` env variable to see how it "attempts" to report the exception to Sentry.

### DIFF
--- a/app/lib/logger_error_subscriber.rb
+++ b/app/lib/logger_error_subscriber.rb
@@ -1,0 +1,7 @@
+# This subscriber will log exceptions to stdout, when using
+# new Rails error reporting API (i.e. `Rails.error.handle`, etc.)
+class LoggerErrorSubscriber
+  def report(error, *)
+    Rails.logger.error [error, error.backtrace[0]].join("\n")
+  end
+end
--- a/app/services/datastore/get_application.rb
+++ b/app/services/datastore/get_application.rb
@@ -7,14 +7,15 @@ module Datastore
     end
 
     def call
-      app = DatastoreApi::Requests::GetApplication.new(
-        application_id:
-      ).call
+      app = Rails.error.record do
+        # Will report and re-raise the exception
+        DatastoreApi::Requests::GetApplication.new(application_id:).call
+      end
 
       raise DatastoreApi::Errors::NotFoundError if superseded?(app)
 
       app
-    rescue DatastoreApi::Errors::NotFoundError
+    rescue DatastoreApi::Errors::ApiError
       raise Errors::ApplicationNotFound
     end
 
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -4,6 +4,10 @@ Rails.application.config.to_prepare do
     config.breadcrumbs_logger = [:active_support_logger, :http_logger]
     config.environment = HostEnv.env_name
 
+    # Opt in to new Rails error reporting API
+    # https://edgeguides.rubyonrails.org/error_reporting.html
+    config.rails.register_error_subscriber = true
+
     # Filtering
     # https://docs.sentry.io/platforms/ruby/guides/rails/configuration/filtering/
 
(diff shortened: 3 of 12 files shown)


app/lib/logger_error_subscriber.rb
```diff
@@ -0,0 +1,7 @@
+# This subscriber will log exceptions to stdout, when using
+# new Rails error reporting API (i.e. `Rails.error.handle`, etc.)
+class LoggerErrorSubscriber
```
timpeat, line 3: Nice

app/services/datastore/get_application.rb
```diff
@@ -7,14 +7,15 @@ def initialize(application_id)
...
+      app = Rails.error.record do
+        # Will report and re-raise the exception
+        DatastoreApi::Requests::GetApplication.new(application_id:).call
+      end
 
       raise DatastoreApi::Errors::NotFoundError if superseded?(app)
 
       app
-    rescue DatastoreApi::Errors::NotFoundError
+    rescue DatastoreApi::Errors::ApiError
```
timpeat, line 18: Did you intent to change the error class here?

config/initializers/sentry.rb
```diff
@@ -4,6 +4,10 @@
     config.breadcrumbs_logger = [:active_support_logger, :http_logger]
     config.environment = HostEnv.env_name
 
+    # Opt in to new Rails error reporting API
+    # https://edgeguides.rubyonrails.org/error_reporting.html
+    config.rails.register_error_subscriber = true
```
timpeat, line 9: I thought this was the default. "Some error-reporting libraries, such as [Sentry's](https://github.com/getsentry/sentry-ruby/blob/e18ce4b6dcce2ebd37778c1e96164684a1e9ebfc/sentry-rails/lib/sentry/rails/error_subscriber.rb) and [Honeybadger's](https://docs.honeybadger.io/lib/ruby/integration-guides/rails-exception-tracking/), automatically register a subscriber for you. Consult your provider's documentation for more details."

timpeat (approved):
Couple of comments for clarification before you merge. Otherwise good.

app/services/datastore/get_application.rb
zheileman: Yep, `NotFoundError` is a subclass of `ApiError`, but there are other errors that can be raised on the Datastore that subclass from `ApiError` that had the risk of not being caught if using a more specific class (like `NotFoundError` before), so I decided it is best to use a more generic class so we are better covered.

config/initializers/sentry.rb
zheileman: Yep I thought so too, but they retracted their "default" position, and made it opt-in.
PR for reference:
https://github.com/getsentry/sentry-ruby/pull/1771

They seem to be still working out some kinks before making it default again.

app/lib/logger_error_subscriber.rb
```diff
@@ -0,0 +1,7 @@
...
+  def report(error, *)
+    Rails.logger.error [error, error.backtrace[0]].join("\n")
```
willmcb, line 5: Could we use `backtrace.first` here? Makes it a little easier to parse maybe

willmcb:
LGTM :)

app/services/datastore/get_application.rb
timpeat: 👍

app/lib/logger_error_subscriber.rb
zheileman: Changed in next PR 👍

PR #428, opened 2023-07-06
CRIMAP-430 Swap order of the URN step

zheileman:
## Description of change
Move URN step after the date stamp (i.e. after case type and/or appeal details).

Update task list so the case details task starts with the case type and not with the URN as it was until now.

For consistency, moved also the place where the URN shows in the review application page (and caused a bit of a mess in the tests, sorry).

Everything (except the order, that is) should continue working as before, nothing really has changed except the order of the steps.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-430

Start a new application and after the contact details, instead of taking you to the URN step, it will take you to the case type step. The date stamp logic remains, and after the date stamp (or for non date-stampable cases, after the case type) you will be taken to the URN page.

Task list also updated. Developer bypass for under 18 also updated.

### DIFF
--- a/app/controllers/developer_tools/crime_applications_controller.rb
+++ b/app/controllers/developer_tools/crime_applications_controller.rb
@@ -58,7 +58,6 @@ module DeveloperTools
           edit_steps_client_has_partner_path(crime_application),
           edit_steps_client_details_path(crime_application),
           edit_steps_client_contact_details_path(crime_application),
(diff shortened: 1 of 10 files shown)


arthurashman (approved):
Excellent, all looks good to me 🎉 

(The word 'answer' started making NO sense to my brain after a while 😂 )

PR #429, opened 2023-07-10
Replace cryptic Omniauth error flash alert

zheileman:
## Description of change
Although it should be very rare and edge case, there could be sometimes Portal authentication errors.

We were showing a rather obscure error message in a flash alert. This has now been replaced with the generic "unhandled" error page, and also we send the exception to Sentry so we are alerted.

More context on this slack thread:
https://mojdt.slack.com/archives/C03JS4V9TPU/p1688989509642419

### Before changes:
Screenshot 2023-07-10 at 12 41 10

### After changes:
Screenshot 2023-07-10 at 15 51 05

It's rather difficult, you need to disable omniauth test mode, and configure Portal locally, then disable something that Portal except, for example `authn_requests_signed` in the file `LaaPortal::SamlSetup`

### DIFF
--- a/app/controllers/providers/omniauth_callbacks_controller.rb
+++ b/app/controllers/providers/omniauth_callbacks_controller.rb
@@ -11,14 +11,15 @@ module Providers
       )
     end
 
+    def failure
+      Rails.error.report($ERROR_INFO, handled: true)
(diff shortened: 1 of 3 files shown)


arthurashman (approved):
👍 Nice, quick fixing!

PR #431, opened 2023-07-11
CRIMAP-435 Remove `hmcts_common_platform` gem dependency

zheileman:
## Description of change
Based on some recent conversations, and the fact the repo for `hmcts_common_platform` is archived and apparently unmaintained and outdated, one immediate easy thing we can do is bring over the (simplified) CSV file and use the same approach we used for offences.

Functionally everything should continue working as before, and we have more flexibility to update or tweak the CSV file to add/remove or rename courts.

In the future we can plan to move to an API based approach using FaCT.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-435
Slack thread:
https://mojdt.slack.com/archives/C02AFLHGK9S/p1689069962130059

Go to the hearing court page, and play around.

### DIFF
--- a/config/data/courts.csv
+++ b/config/data/courts.csv
@@ -1,0 +1,251 @@
+code,name
+B,Aberystwyth Magistrates' Court
+B,Aldershot Magistrates' Court
+B,Amersham Magistrates' Court
+C,Aylesbury Crown Court
+B,Banbury Magistrates' Court
+B,Barkingside Magistrates' Court
+B,Barnsley Magistrates' Court
+B,Barnstaple Magistrates' Court
+B,Barrow-in-Furness Magistrates' Court
+C,Basildon Crown Court
+B,Basildon Magistrates' Court
+B,Basingstoke Magistrates' Court
+B,Bath Magistrates' Court
+B,Bedford Magistrates' Court
+B,Berwick-upon-Tweed Magistrates' Court
+B,Beverley Magistrates' Court
+B,Bexley Magistrates' Court
+C,Birmingham Crown Court
+B,Birmingham Magistrates' Court
+B,Blackburn Magistrates' Court
+C,Blackfriars Crown Court
+B,Blackpool Magistrates' Court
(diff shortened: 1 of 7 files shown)


config/data/courts.csv
```diff
@@ -0,0 +1,251 @@
...
+B,Medway Magistrates' Court
+C,Merthyr Tydfil Crown Court
+B,Merthyr Tydfil Magistrates' Court
+B,Milton Keynes Magistrates' Court
+C,Minshull Street Manchester Crown Court
+C,Mold Crown Court
+B,Mold Magistrates' Court
+B,Newcastle Magistrates' Court
+C,Newcastle upon Tyne Crown Court
+B,Newcastle upon Tyne Magistrates' Court
```
willmcb, line 147: Do you think these are duplicates, on lines 145 / 147? Googling around, seems like these refer to the same place.

config/data/courts.csv
```diff
@@ -0,0 +1,251 @@
...
+B,Redditch Magistrates' Court
+B,Romford Magistrates' Court
+C,Salisbury Crown Court
+B,Salisbury Magistrates' Court
+B,Scarborough Magistrates' Court
+B,Sevenoaks Magistrates' Court
+C,Sheffield Crown Court
+B,Sheffield Magistrates' Court
+C,Shrewsbury Crown Court
+C,Shrewsbury Crown Court
```
willmcb, line 188: Seems to be a double entry here as well

willmcb (approved):
👍

config/data/courts.csv
zheileman: Yep, this is one of the courts @arthurashman was looking at. I just took the list of names from the gem and bring it over, without any changes.

config/data/courts.csv
zheileman: Good eye. This is definitely a duplicate. Will remove it and update the digest.

PR #432, opened 2023-07-12
Bump datastore API client gem (v1.0.0)

zheileman:
## Description of change
No functional changes, the gem dependencies have been refreshen and released as `v1.0.0`

Using the tagged version in the Gemfile, same as we do with the schemas gem, to avoid inadvertently introducing breaking changes when updating the bundle.


willmcb (approved):
👍

PR #433, opened 2023-07-13
CRIMAP-439 Ensure enter key triggers main submit action

zheileman:
## Description of change
Default browser behaviour when user focus in a form input, like a text field, and they press enter, is to submit the form.

This usually works just fine, as there is only one submit per form or the main submit comes before the others.

However, in some steps, like the postcode lookup, there is more than 1 submit action and the browser picks the first one in order of appearance in the DOM, which might not be the one that actually performs the “save and continue” (or the postcode search in this case).

This works around this behaviour in a progressive enhancement way with a bit of JS, by cloning the main submit action and positioning it (hidden, even for screen-readers) right at the top of the form so `enter key` picks this one always.

Only forms tagged with "data-multi-action-form=true" will have this progressive enhancement applied, as there is no need to do this with other forms.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-439

Observe previous behaviour (without the changes in this PR) in steps. Enter key will not act same as "save and continue".
- postcode lookup
- offences page
- codefendants page

With the changes in this PR, when focus is in a text input in any of those pages, and pressing `enter` key, the form will be submitted, same behaviour as if the user clicks the "green button / save and continue".

### DIFF
--- a/spec/helpers/form_builder_helper_spec.rb
+++ b/spec/helpers/form_builder_helper_spec.rb
@@ -13,10 +13,11 @@ RSpec.describe FormBuilderHelper, type: :helper do
   end
 
   describe '#continue_button' do
+    # rubocop:disable Layout/LineLength
     context 'when there is no secondary action' do
       let(:expected_markup) do
         '<button type="submit" formnovalidate="formnovalidate" class="govuk-button" ' \
-          'data-module="govuk-button" data-prevent-double-click="true">Save and continue</button>'
+          'data-module="govuk-button" data-prevent-double-click="true" data-main-action="true">Save and continue</button>'
       end
 
       it 'outputs only the continue button' do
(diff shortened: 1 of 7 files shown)


spec/helpers/form_builder_helper_spec.rb
```diff
@@ -13,10 +13,11 @@
   end
 
   describe '#continue_button' do
+    # rubocop:disable Layout/LineLength
     context 'when there is no secondary action' do
```
willmcb, line 17: Might be me reading this wrong but the wording on this test is confusing to me in this context. If there is no secondry action on the form then surely that form doesn't need its submit to go through the hoisting to the top.

I think the test is logically correct just the description of throws me a bit 🤔

willmcb (approved):
Have a look at that comment I made - some wording didn't make sense to me.

Minor non-blocking thing though, I like this solution pretty straight forward mechanism to get around this issue 👍

spec/helpers/form_builder_helper_spec.rb
zheileman: These tests are about the form builder. Secondary in this scenario is the "draft" button. We can have the green button, and most of the time next to it we will also have the grey save draft button. In some very specific steps we don't have this secondary (draft) button, like the first few steps when starting an application, and same blue pages.

Screenshot 2023-07-13 at 11 26 54

arthurashman (approved):
Nice fix! Works as expected on manual testing.

spec/helpers/form_builder_helper_spec.rb
willmcb: Ah right! I get it - was thinking about the word secondry more generally. Thanks for clearing this up for me 👍

zheileman: Yep, grey buttons are all considered secondary in terms of the design system. The form builder uses `secondary: true` in the helper, and then the applied css class is `govuk-button--secondary` thus the naming 😄

PR #434, opened 2023-07-13
CRIMAP-299 Remove http basic auth

zheileman:
## Description of change
Preparation to going live.

Still Portal authentication is required so this does not mean we are opening the service to everybody.

NOTE: I've also generated as discussed a new `SECRET_KEY_BASE` for production.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-299

After deployment, going to either staging or production will not have http basic auth.


config/kubernetes/staging/ingress.yml
```diff
@@ -6,8 +6,6 @@ metadata:
   annotations:
     external-dns.alpha.kubernetes.io/set-identifier: ingress-staging-laa-apply-for-criminal-legal-aid-staging-green
     external-dns.alpha.kubernetes.io/aws-weight: "100"
-    nginx.ingress.kubernetes.io/auth-type: basic
-    nginx.ingress.kubernetes.io/auth-secret: secrets-staging
     nginx.ingress.kubernetes.io/server-snippet: |
       if ($host = 'laa-apply-for-criminal-legal-aid-staging.apps.live.cloud-platform.service.justice.gov.uk') {
         return 301 https://laa-apply-for-criminal-legal-aid-staging.apps.cloud-platform.service.justice.gov.uk;
```
willmcb, line 11: Do we want to remove the line `auth_basic off;` @ Line 19 as well? If we don't have it then is it confusing this is still here?

I guess if we ever put it back on then the security.txt will need to have it off so perhaps it is okay to leave. Up to you - though I would just flag as it cuaght my eye 👀

willmcb (approved):
One comment just to falg something :-) 

LGTM 👍

config/kubernetes/staging/ingress.yml
zheileman: I left that on purpose as the likes of enabling back http basic auth and forgetting to also add this annotation to the snippet are high 😄 
It does no harm at all leaving it.

PR #435, opened 2023-07-13
Fix navigation after entering 'other' address to go to case type not URN

arthurashman:
## Description of change
When we updated user journey to go from correspondence details to case type (instead of to URN) the journey when inputting an alternative address covered by the address step controller remained with URN as the next step - meaning that case type was never asked.

This PR fixed the routing so that after inputting an alternative address, the user is sent to the case type page and all steps are shown 

Start an application as normal up to correspondence address step
Select Somewhere else on the radios
Input an address by lookup or manual input
Click save and continue
Expect to go to Case type step


zheileman (approved):
Nice catch, thanks!

PR #441, opened 2023-07-18
Add restriction to dob that raises error if applicant under 10

arthurashman:
## Description of change
Client cannot be under age 10 - below this age, children cannot be held criminally responsible. 

MAAT has a validation whereby trying to save application details with a DoB that means the applicant is under the age of 10 it errors. Crime Apply should replicate this error to prevent incompatible data reaching MAAT, but rather to raise an error with the provider as they input the DoB to get them to check the DoB for errors

Adds validation that input date results in the client being 10 years old or older. Adds an error if not and raises this to the user

## Link to relevant ticket
https://dsdmoj.atlassian.net/jira/software/projects/CRIMAP/boards/897?selectedIssue=CRIMAP-451
### After changes:
image

Start an application and fill client details with a date of birth more recent than 18/07/2013 - e.g. 01/08/2013

### DIFF
--- a/app/forms/steps/client/details_form.rb
+++ b/app/forms/steps/client/details_form.rb
@@ -12,7 +12,8 @@ module Steps
       validates_presence_of :first_name,
                             :last_name
 
-      validates :date_of_birth, presence: true, multiparam_date: true
+      validates :date_of_birth, presence: true,
+                multiparam_date: { restrict_past_under_ten_years: true }
 
       private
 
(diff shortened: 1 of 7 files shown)


zheileman (commented):
Looks good to me! I think is a bit cleaner and simpler than using the multiparam validator 👍 

Just one comment, for the shared examples, as I think this is best tested in the form object as is a very specific validation only happening in there.

PR #442, opened 2023-07-18
Set `Cache-Control` response header to no-store

zheileman:
## Description of change
It avoids possibly exposing personal information when using the browser back button after the current session has been terminated.

This is similar to the [approach taken on Review](https://github.com/ministryofjustice/laa-review-criminal-legal-aid/pull/361) however instead of adding it to all controller actions, this is only set on those exposing possible personal information.

There are many other pages on Apply (all footer pages, and some others) that are not sensitive.

Note pragma header not set as is deprecated and no longer recommended. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Pragma

## Notes for reviewer
I'm not 100% sure this works well with Safari. Tested well in all other browsers including Firefox, Chrome, Opera and Edge.

Sign in, go to the dashboard, then go to submitted applications for instance, open one of those applications. Then while having it opened, sign out, then click the browser's back button, one or more times, you should not see the previous application details, and a page to sign in should be presented instead.

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -2,6 +2,10 @@ class DashboardController < ApplicationController
   helper_method :in_progress_count, :returned_count, :sortable_columns
   delegate :returned_count, to: :application_counters
 
+  # Avoid the browser caching any of the dashboard pages so the
+  # back button after signing out can't load any sensitive details
+  before_action :no_store
+
   private
 
   # Implement in sub-controllers to narrow down allowed columns
--- a/app/controllers/steps/base_step_controller.rb
+++ b/app/controllers/steps/base_step_controller.rb
@@ -3,6 +3,10 @@ module Steps
     before_action :check_crime_application_presence
     before_action :update_navigation_stack, only: [:show, :edit]
 
+    # Avoid the browser caching any of the step pages so the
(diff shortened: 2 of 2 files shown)


zheileman:
@arthurashman I'm hoping perhaps this behaviour with Safari only happens when in localhost? Maybe a "real" environment like staging behaves differently? I guess there is a bit of further try&error and investigation.

arthurashman:
OK, happy to approve and we can test in stg?

arthurashman (approved):
👍

zheileman:
@arthurashman yeah, I've merged it, easy to test anyways.

However, there is something else, have you noticed after clicking back and getting the `unauthenticated` error page, if you then click the "Sign in" button on that page, you get an `ActionController::InvalidAuthenticityToken` error and redirects to unhandled error page.

This is the token being somehow cached. When using a crude cache busting technique like adding a `?t=1689693371` it works however. So something still going on. This happens to me with all browsers, and sometimes it works first time but fails after redoing a few more times.

Something else to investigate 😮

arthurashman:
Oh, yes! Not great

zheileman:
@arthurashman are you happy to do some testing on Staging and report back? Same testing you've done in your localhost, see if there is any difference at all on staging? Thanks in advance!

PR #443, opened 2023-07-19
Discern between invalid session and invalid token errors

zheileman:
## Description of change
Omniauth authentication could fail due to an invalid authenticity token. For instance, by re-submitting the same form twice. In this scenario we were redirecting to the unhandled error page, which is not correct and confusing to the user.

Now we just let `InvalidAuthenticityToken` exceptions bubble up, so it can be rescued and handled accordingly within the generic error handling block.

Also previously we were bunching together the invalid session and the authenticity token errors, but they are not really the same and to a user's point of view, showing the invalid session error when the error was produced by an invalid authenticity token (forgery protection) might be confusing as we tell them:

"For your security, we signed you out
This is because you were inactive for 48 hours."

Now each error is separate and the `InvalidAuthenticityToken` will redirect to a different error page with a more lax error message and instruct users to go back, reload and retry.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-459

### Before changes:
When produced in omniauth flow:
Screenshot 2023-07-19 at 11 49 35

When produced anywhere else:
Screenshot 2023-07-19 at 13 57 45

### After changes:
Screenshot 2023-07-20 at 10 29 54

Sign in, go back and click Sign in again. Without the changes in this PR, you will get an unhandled error page. With the changes in this page, you will get a "Sorry, there is a problem with your request" page.

As part of the steps journey: start an application, and when in a step form, edit the page and change the form param `authenticity_token` and submit to produce an `InvalidAuthenticityToken` exception. It shows the "Sorry, there is a problem with your request" page.

### DIFF
--- a/app/controllers/concerns/error_handling.rb
+++ b/app/controllers/concerns/error_handling.rb
@@ -4,7 +4,9 @@ module ErrorHandling
   included do
     rescue_from Exception do |exception|
       case exception
-      when Errors::InvalidSession, ActionController::InvalidAuthenticityToken
+      when ActionController::InvalidAuthenticityToken
+        redirect_to unauthenticated_errors_path
+      when Errors::InvalidSession
         redirect_to invalid_session_errors_path
       when Errors::ApplicationNotFound
         redirect_to application_not_found_errors_path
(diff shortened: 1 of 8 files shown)


timpeat (approved):
👍

timpeat (approved):

PR #447, opened 2023-07-20
Tweak rails Grafana dashboard

zheileman:
## Description of change
Updated the Request duration and Request count widgets, to `avg` or `sum` requests grouped by controller and action.

Previously it would show n-times the same controller/action, as many times as pods (4 in production, 2 in staging).

I've applied it already so it so we can see how it looks.


willmcb (approved):
👍

PR #452, opened 2023-07-26
Refactor custom prometheus collectors

zheileman:
## Description of change
As we now have 3 collectors, a bit of refactor to DRY some code was due.

Functionally they are exactly the same and produce the same metrics with the same names as before.

Added a new collector to gather the total number of unique offices across all enrolled providers and a Grafana widget to visualise this.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-506

As before, these can be tested locally either by exposing the ENV variable `ENABLE_PROMETHEUS_EXPORTER=true` or by using docker-compose, then accessing the /metrics endpoint.


app/lib/prometheus_metrics/collectors/applications_count_collector.rb
```diff
@@ -0,0 +1,53 @@
...
+          total_count,
+          in_progress_count,
+          date_stamped_count,
+          stale_count,
+        ]
+      end
+
+      # Some applications can be submitted very fast, so we reduce
+      # the cache from the default 5m to 2m in this collector
+      def expires_in
```
willmcb, line 23: 👍 nice

spec/lib/prometheus_metrics/collectors/offices_count_collector_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'rails_helper'
+
+# Very light-touch smoke test as there is no point in having these
+# too elaborated, as metrics will probably change frequently
```
willmcb, line 4: 👍

willmcb (approved):
Nice, lgtm 👍

PR #453, opened 2023-07-26
Remove unneeded webdrivers gem

zheileman:
## Description of change
And cleanup Webmock initialiser as now there is nothing needed in it, the default is to `allow_localhost` anyways.

Note: we need to observe if this runs properly on CI. It `works on my machine ™`


hiboabd (approved):
Fingers crossed

timpeat (approved):

PR #456, opened 2023-08-09
CRIMAP-548 Add explicit 'no conflict' caption to CYA

arthurashman:
## Description of change
Add explicit 'no conflict' caption to CYA so provider can see what has been input should they need to change it. 

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-548

### Before changes:
image

### After changes:
image

### DIFF
--- a/app/presenters/summary/sections/codefendants.rb
+++ b/app/presenters/summary/sections/codefendants.rb
@@ -42,7 +42,7 @@ module Summary
         if YesNoAnswer.new(codefendant.conflict_of_interest).yes?
           [codefendant.full_name, I18n.t('summary.questions.conflict_of_interest_html')].join
         else
-          codefendant.full_name
+          [codefendant.full_name, I18n.t('summary.questions.no_conflict_of_interest_html')].join
         end
       end
     end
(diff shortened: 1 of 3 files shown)


app/presenters/summary/sections/codefendants.rb
```diff
@@ -42,7 +42,7 @@ def name_and_conflict(codefendant)
         if YesNoAnswer.new(codefendant.conflict_of_interest).yes?
```
zheileman, line 42: We could now refactor this so there is no need for IF-ELSE
We can pass the value of `codefendant.conflict_of_interest` to the I18n

Like:
```[codefendant.full_name, I18n.t(codefendant.conflict_of_interest, scope: 'summary.questions.conflict_of_interest_html')].join```

zheileman (approved):
We can do a bit of refactor if you feel like it, but otherwise approving 👍

app/presenters/summary/sections/codefendants.rb
arthurashman: Yep, of course. Makes sense. Thank you 👍

PR #458, opened 2023-08-11
Docker compose works with apple silicon

timpeat:
## Description of change

Add gcombat and update dart-sass to allow ```docker-compose up``` with Apple Silicon

## Notes for reviewer

gcombat allows glibc programs to be run in Alpine Linux (https://wiki.alpinelinux.org/wiki/Running_glibc_programs)

It was also necessary to update dart-sass to a version that correctly understands the architecture it is working in.

To make sure this does not introduce a regression on Intel Mac, remove any docker files you may already have and run  ```docker-compose up```


Dockerfile
```diff
@@ -5,7 +5,8 @@ MAINTAINER LAA Crime Apply Team
 RUN apk add --update \
   postgresql-dev \
   tzdata \
-  yarn
+  yarn \
+  gcompat
 
 # Alpine does not have a glibc, and this is needed for dart-sass
```
zheileman, line 11: As mentioned on Review PR, I do believe we can now happily remove all these glibc related setup as either `gcompat` or maybe the bump to dart-sass makes it unnecessary, which is great!
I've tested in my non-M1 removing these lines and everything keeps working so far. Would be good to double check on an M1 machine too.

PR #459, opened 2023-08-14
CRIMAP-527 Create basic upload function, add page to flow

arthurashman:
## Description of change
Added evidence upload decision tree, controller, route, view and feature flag
## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-527
## Notes for reviewer
To try out uploading something, add a folder called 'uploads' to the 'public' folder
### After changes:
image

Start an application and go to 
/applications/[application_uuid]/steps/evidence/upload

### DIFF
--- a/app/controllers/steps/evidence/upload_controller.rb
+++ b/app/controllers/steps/evidence/upload_controller.rb
@@ -11,15 +11,13 @@ module Steps
       def update
         if params[:steps_evidence_upload_form]
 
-        uploaded_files = params[:steps_evidence_upload_form][:upload_files]
+          uploaded_files = params[:steps_evidence_upload_form][:upload_files]
 
-        uploaded_files.each do |uploaded_file|
-          File.open(Rails.root.join('public', 'uploads', uploaded_file.original_filename), 'wb') do |file|
-            file.write(uploaded_file.read)
+          uploaded_files.each do |uploaded_file|
+            Rails.public_path.join('uploads', uploaded_file.original_filename).binwrite(uploaded_file.read)
           end
         end
-      end
-      # :nocov:
+        # :nocov:
 
         update_and_advance(UploadForm, as: :upload)
       end
--- a/app/controllers/steps/evidence/upload_controller.rb
+++ b/app/controllers/steps/evidence/upload_controller.rb
@@ -1,0 +1,28 @@
+module Steps
+  module Evidence
+    class UploadController < Steps::EvidenceStepController
+      def edit
+        @form_object = UploadForm.build(
+          current_crime_application
+        )
+      end
+
+      # :nocov:
+      def update
+        if params[:steps_evidence_upload_form]
+
+        uploaded_files = params[:steps_evidence_upload_form][:upload_files]
+
+        uploaded_files.each do |uploaded_file|
+          File.open(Rails.root.join('public', 'uploads', uploaded_file.original_filename), 'wb') do |file|
+            file.write(uploaded_file.read)
+          end
+        end
+      end
+      # :nocov:
+
+        update_and_advance(UploadForm, as: :upload)
+      end
+    end
+  end
+end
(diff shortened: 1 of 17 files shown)


config/locales/en/helpers.yml
```diff
@@ -88,6 +88,9 @@ en:
         expert_examination_justification: *ioj_justification_hint
         interest_of_another_justification: *ioj_justification_hint
         other_justification: *ioj_justification_hint
+      steps_evidence_upload_form:
+        #TODO update max file size
```
willmcb, line 92: Do we want to have an overall limit as well?

config/locales/en/steps.yml
```diff
@@ -178,6 +178,18 @@ en:
       ioj:
         edit:
           page_title: Why should your client get legal aid?
+    evidence:
+      upload:
+        edit:
+          page_title: Upload supporting evidence
+          heading: Upload supporting evidence
+          list:
+            intro: 'Use this page to provide:'
```
willmcb, line 187: Not sure we need quotes here

app/services/decisions/case_decision_tree.rb
```diff
@@ -108,7 +108,11 @@ def after_hearing_details
     # rubocop:disable Style/GuardClause
     def after_ioj
       if Passporting::MeansPassporter.new(current_crime_application).call
-        edit('/steps/submission/review')
```
zheileman, line 111: Can we for now, just ignore feature flags etc.? Do not introduce any decision in previous trees at all.
Trying to self contained this to the absolute minimum without polluting any other parts of the application.

app/views/steps/evidence/upload/edit.html.erb
```diff
@@ -0,0 +1,29 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <div>
+      <h1 class="govuk-heading-xl">
+        <%=t '.heading' %>
+      </h1>
+
+      <p><%= t('.list.intro') %></p>
+
+      <ul class="govuk-list govuk-list--bullet">
+        <li><%= t('.list.option_1') %></li>
```
zheileman, line 14: I think is safe to assume this list will be dynamic. The design is a mock up. For now I propose hardcoding it in the view, we will figure out the logic to show "items" in the list later. Let's keep this focused and simple for now.

config/settings.yml
```diff
@@ -15,6 +15,10 @@ feature_flags:
     local: true
     staging: true
     production: true
+  evidence_upload:
```
zheileman, line 18: It's ok leaving the feature configuration but don't think we need to worry about using it right now. Only way to hit the page should be manually knowing the URL.

app/controllers/steps/evidence/upload_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+  module Evidence
+    class UploadController < Steps::EvidenceStepController
+      def edit
+        @form_object = UploadForm.build(
+          current_crime_application
+        )
+      end
+
+      # :nocov:
+      def update
```
zheileman, line 11: Upload will not happen here, this will advance to next step, uploads have happened before this.
This however will be used to update the categories of the files. But that is for later.

zheileman (changes requested):

config/locales/en/steps.yml
arthurashman: For the colon

config/locales/en/helpers.yml
arthurashman: I'll put it in the TODO. Had just been following the designs

app/views/steps/evidence/upload/edit.html.erb
arthurashman: Oh, I hadn't heard anything about it being dynamic. Makse sense - I'll change it back to hard code

config/locales/en/steps.yml
arthurashman: Also removed based on Jesus' suggestion

PR #460, opened 2023-08-15
CRIMAP-529 Add MOJ javascript and dropzone

zheileman:
## Description of change
Add all the stuff needed to render an MOJ multi-upload component (dropzone behind the scenes).

Note in order to do this we now have to add to the application the MOJ javascript plus dropzone dependencies.

I'm raising this PR just to have it all set, but is obviously not functional, and I doubt it can be functional out of the box without modifying MOJ javascript because for once, the JS is not propagating the authenticity token and I don't see a way to configure this, meaning only way to propagate the token is to modify the javascript.

My opinion is we shouldn't be using MOJ component, which couples us to their javascript (and requires jQuery nonetheless) and instead code our own JS handling and apply our own styling (which can be the same as the MOJ styling if we wish).

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-529

## Notes for reviewer
In order to avoid having jQuery loaded globally in each of the pages in the application, I'm only loading it ad-hoc in the evidence upload page (`javascript_include_tag`).

Screenshot 2023-08-15 at 11 26 46

Checkout, run `yarn`, run `rails assets:clobber && rails dartsass:build` and restart server. Go to the evidence upload and the drop zone should be "functional" and the styling look like the one on MOJ component page.

### DIFF
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -1,5 +1,10 @@
 //= link_tree ../images
+//= link_tree ../builds
+
 //= link_tree ../../javascript .js
 //= link_tree ../../../vendor/javascript .js
-//= link_tree ../builds
+
 //= link_tree ../../../node_modules/govuk-frontend/govuk/assets
+//= link_tree ../../../node_modules/@ministryofjustice/frontend/moj/assets
+
+//= link jquery/dist/jquery.js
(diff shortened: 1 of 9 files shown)


arthurashman (approved):
Nice work 👍 

Works and looks as expected on local

PR #462, opened 2023-08-16
CRIMAP-529 Dropzone without MOJ component

zheileman:
## Description of change
This is a counterpart to previous PR #460.

We may decide this is a better way, as I do believe we will need to heavily modify and customise MOJ multi-upload component anyways, so we might as well avoid using it, and instead just use a standalone dropzone module (that doesn't require jQuery) and our own styling and JS for events handling.

I've made a head start on styling and JS events preparing the structure and foundation. We can either migrate styling and JS handling from MOJ component, or from other service like Civil Apply, or Tax Tribunals, etc. that use dropzone standalone too.

Note how, out of the box, without any additional JS, dropzone handles the Ajax call, and the authenticity token is propagated, etc.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-529

Screenshot 2023-08-16 at 08 37 50

As before, checkout, run `yarn`, run `rails assets:clobber && rails dartsass:build` and restart server. Go to the evidence upload page and the drop zone should be "functional" but it will have out of the box dropzone-supplied styling for now.


timpeat (approved):
❤️

willmcb (approved):
This looks great - agree we will need to change styles etc., but personally I like the way it looks and the little preview Icons are quite cool as well.

config/gatekeeper.yml
```diff
@@ -27,3 +27,5 @@ production:
     # Second onboarding tranche - 31/07/2023
     - 2A041V
     - 1W261L
+    # Third onboarding tranche - 18/08/2023
+    - 1Q921U
```
timpeat, line 31: Thank you :)

config/routes.rb
```diff
@@ -135,7 +135,7 @@ def show_step(name)
         edit_step :ioj
       end
 
-      namespace :evidence do
+      namespace :evidence, constraints: -> (_) { FeatureFlags.evidence_upload.enabled? } do
```
timpeat, line 138: 👍

timpeat (approved):
LGTM

PR #466, opened 2023-08-23
Reduce restrictions on URN validation

arthurashman:
## Description of change
URN field was raising validation errors on valid URNs. The rules we have been told are not correct, so we need to reduce restrictions given we now don't know what the specific format is.

Following suit of VCD, which has validations that URN is alphanumeric and fewer than 20 chars. This PR also sets a minimum of 6 chars, but otherwise the same as VCD


willmcb (approved):
🐛 💨

PR #467, opened 2023-08-23
Add flat file for inactive offices

willmcb:
## Description of change
An implementation using a flat file as an interim solution to filter out inactive office codes from the portal.

Idea her is to test on staging and then I will raise another PR with the inactive prod codes and deploy this again.

**The flow of this is:**

1. sign in is user attempts sign-in from portal 
2. callback in `OmniauthCallbacksController` checks they are allowlisted/enrolled with the `Providers::GateKeeper`
3. the `saml` method is called in the controller which in turn calls `Provider.from_omniauth` method with the `auth_hash`
4. `from_omniauth` finds or initialises a new provider, or updates them if they are already present
5. In the update statement, the `active_office_codes` method calls the new `ActiveOfficeChecker` service which filters our inactive offices based on the flatfile.
6. Inactive offices should then not show for newly logged in providers provided they are added to the flatfile.

I have also checked the way that the prometheus logs are collated and I think they should still work, [essentially they just call the `.office_codes` method on the `Providers` class](https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/blob/4cace39fa81d5c248a6db7f851965056220eb15c/app/lib/prometheus_metrics/collectors/offices_count_collector.rb#L14C2-L14C2), which is [updated as a provider signs into the system.](https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/blob/4cace39fa81d5c248a6db7f851965056220eb15c/app/models/provider.rb#L21C1-L21C1) So the metric should reflect "live" offices as we would expect it to.

## Link to relevant ticket
[CRIMAP-566](https://dsdmoj.atlassian.net/browse/CRIMAP-566)

You can see in the LaaPortal::SamlStrategy that I have added some more codes the the `mock_auth` these extra codes are added to the `inactive_offices.yml` so are filtered out, this means you should not see these codes as selectable locally. You can remove the codes from he `yml` file and reload the app and you should see these codes appear.

As you change this you should also see them update in the `office_codes` method of the Provider record in you're local database.

Please also test that you can go through the application e2e and there are no downstream affects of this change. 

[CRIMAP-566]: https://dsdmoj.atlassian.net/browse/CRIMAP-566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/services/providers/active_office_checker_spec.rb
+++ b/spec/services/providers/active_office_checker_spec.rb
@@ -28,7 +28,7 @@ RSpec.describe Providers::ActiveOfficeChecker do
       context 'inactive codes are not in office code list' do
         let(:inactive_office_codes) { %w[3C345D 4C777Z] }
 
-        it 'returns the an unaltered office code list' do
+        it 'returns an unaltered office code list' do
           active_codes = %w[1A123B 2A555X 2B234C 3B666Y]
           expect(subject.active_office_codes).to match(active_codes)
         end
--- a/spec/services/providers/active_office_checker_spec.rb
+++ b/spec/services/providers/active_office_checker_spec.rb
@@ -1,0 +1,67 @@
+require 'rails_helper'
+
+RSpec.describe Providers::ActiveOfficeChecker do
+  subject { described_class.new(auth_info) }
+
+  let(:auth_info) do
+    double(
+      office_codes:,
+    )
+  end
+
+  let(:office_codes) { %w[1A123B 2A555X 2B234C 3B666Y] }
+  let(:inactive_office_codes) { %w[2B234C 3B666Y] }
+
+  describe '#active_office_codes' do
+    before do
+      allow(subject)
+        .to receive(:inactive_office_codes)
+        .and_return(inactive_office_codes)
+    end
+
+    context 'multi-code office' do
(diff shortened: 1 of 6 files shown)


spec/services/providers/active_office_checker_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    context 'multi-code office' do
+      it 'returns a list of active office codes' do
+        active_codes = %w[1A123B 2A555X]
+        expect(subject.active_office_codes).to match(active_codes)
+      end
+
+      context 'inactive codes are not in office code list' do
+        let(:inactive_office_codes) { %w[3C345D 4C777Z] }
+
+        it 'returns the an unaltered office code list' do
```
mseedat-moj, line 31: bit of a typo here in the test name

mseedat-moj (commented):
minor typo in test. Looks good - solution works well. Regarding testing - does there need to be integration tests/end-to-end test showing what happens during login when an inactive provider tries?

arthurashman (approved):
LGTM. Tested locally and works as expected

PR #469, opened 2023-08-29
[CRIMAP-544] List uploaded evidence for in progress application

hiboabd:
## Description of change
This PR introduces the initial functionality to upload evidence for a draft application.

The `DocumentController` receives the file via the submit form and creates a local `Document` record with attributes like file_name etc. One attribute will be nil at this point, `object_key`.

The document is validated on size and content criteria and if valid, the `Upload` service requests an upload presigned URL via the datastore API and the presigned URL is then used to perform the upload directly to the bucket.

Upon successful completion of the upload, the returned `object_key` is written back to the `Document` record, and the transaction is finalised.

Upload failures are handled in the controller for JS and non-JS versions.

## Link to relevant ticket

[CRIMAP-544](https://dsdmoj.atlassian.net/browse/CRIMAP-544)

## Notes for reviewer
Apologies for the chunky PR!

Note: we have disabled multi file uploads for the non js version as the request payload included all the documents and it was not uploading one document at a time. As it's very unlikely that anyone will use the non js version, we did not feel it was worth the technical effort to handle both transaction types so the upload has been limited to one file at a time. 

Note: Application types list has been lifted from another application and may need dev review but should do for the time being. Design and content need reviewing, particularly for the error text.

Instructions to test are below.

### Before changes:
Screenshot 2023-09-12 at 17 36 54

### After changes:
Screenshot 2023-09-12 at 17 18 34

https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/assets/51047911/c9c850a9-cdf8-4047-956a-fc9600b22e9d

This requires having both datastore and apply running

1. In your datastore repo, navigate to `main` and run `bundle install` to ensure gems are up to date. 
2. In one terminal, run `rails server` to spin up datastore 
3. In another terminal, run `docker-compose up localstack` to spin up localstack (required to upload to s3) 
4. In apply, navigate to this branch and run `bundle install` to ensure gems are up to date and spin up apply by running `rails server`. You may need to run a migration.
5. Then navigate to apply in your chosen browser and create a draft application (or use an existing draft application) and navigate to `/applications/{application_uuid}/steps/evidence/upload`. You should then be able to upload a file

Please disable JS in your chosen browser and test the non js upload functionality. 

[CRIMAP-544]: https://dsdmoj.atlassian.net/browse/CRIMAP-544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/views/steps/evidence/upload/edit.en.html.erb
```diff
@@ -7,32 +7,61 @@
...
+      <%= t('.upload_files_heading') %>
+      <span class="govuk-caption-m govuk-!-padding-top-2"><%= t('.upload_files_info') %></span>
+    </h2>
+
     <%= form_for @form_object, html: { id: 'dz-evidence-upload-form', class: 'dropzone' },
-                 url: crime_application_documents_path,
+                 url: bundle_documents_path(@form_object.current_document_bundle),
                  method: :post, multipart: true do |f| %>
 
+      <%= f.button t('.choose_files_button'),
```
arthurashman, line 26: When js is switched off, we've ended up with double the buttons, I think this whole button section needs deleting.
image

app/views/steps/evidence/upload/edit.en.html.erb
```diff
@@ -7,32 +7,61 @@
...
+
       <div class="fallback">
         <%= f.govuk_file_field :documents, multiple: true,
                                label: { size: 'm' }, class: 'govuk-file-upload' %>
 
         <%= f.button t('.upload_file_button'),
                      class: 'govuk-button govuk-button--secondary',
                      data: { module: 'govuk-button' } %>
+
+        <div class="moj-multi-file__uploaded-files">
```
arthurashman, line 40: And the list is doubled up too, so this whole div needs deleting too (I think)

app/views/steps/evidence/upload/edit.en.html.erb
arthurashman: Playing with it and switching js on and off, which 'Choose files' button shows needs to be conditional on if the fallback is used

arthurashman: I was looking into this and it's easier said than done I fear. I think we will need to give the js button a hidden class that is removed by js (therefore _not_ removed if js isn't on)

hiboabd: I've amended this

app/controllers/documents_controller.rb
```diff
@@ -4,19 +4,58 @@ class DocumentsController < ApplicationController
...
+  end
+
+  def current_document_bundle
+    @current_document_bundle ||= DocumentBundle.find(document_bundle_id)
+  end
+
+  # TODO: unify if possible the submission params
+  # Currently XHR and HTML requests have different structure
+  # Also needs proper error handling for non-JS version if user click
+  # upload button without having selected a file
```
hiboabd, line 44: Currently, this routes to the generic there is 'something wrong with the service' page. Can write a ticket to do this but i think this should go in to unblock other tickets

app/views/steps/evidence/upload/_uploaded_file.html.erb
```diff
@@ -0,0 +1,15 @@
+<tr class="govuk-table__row">
+  <td id="<%=document.id%>" class="govuk-table__cell">
+    <span><%= document.filename %></span>
+
+    <% if document.s3_object_key %>
+      <strong class="govuk-tag govuk-tag--green">Uploaded</strong>
+    <% else %>
+      <% document.valid?(:storage) %>
```
hiboabd, line 8: Error messages are not saved so when documents are retrieved they need to be validated against our criteria so the correct error message can be displayed

arthurashman:
Just a tiny thing, but with JS on, the upload button remains with yellow focus state. It feels like it should return to grey after file selected

arthurashman:
Hmmm v weird, that's how the example works - https://design-patterns.service.justice.gov.uk/components/multi-file-upload/

Although if you take the visually hidden and js to show the button out then drag and dropping doesn't have the continued highlight (although this _is_ on the MoJ design component.

Maybe ignore me and leave this as it is

app/validators/file_upload_validator.rb
```diff
@@ -0,0 +1,57 @@
+class FileUploadValidator < ActiveModel::Validator
+  MIN_FILE_SIZE = 5 # KB
+  MAX_FILE_SIZE = 20 # MB
+
+  ALLOWED_CONTENT_TYPES = %w[
+    application/msword
+    application/pdf
+    application/rtf
+    application/vnd.ms-excel
```
arthurashman, line 9: Are we meant to be allowing excel files? It's not listed on the error on the design

arthurashman: Civil don't but do allow csv files

app/validators/file_upload_validator.rb
```diff
@@ -0,0 +1,57 @@
...
+    application/vnd.oasis.opendocument.text
+    application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
+    application/vnd.openxmlformats-officedocument.wordprocessingml.document
+    image/bmp
+    image/gif
+    image/jpeg
+    image/png
+    image/tiff
+    image/x-bitmap
+    text/csv
```
arthurashman, line 19: If we're going to allow these, we need to double check that's ok, and update the copy in the designs to say csv

app/validators/file_upload_validator.rb
```diff
@@ -0,0 +1,57 @@
...
+  MIN_FILE_SIZE = 5 # KB
+  MAX_FILE_SIZE = 20 # MB
+
+  ALLOWED_CONTENT_TYPES = %w[
+    application/msword
+    application/pdf
+    application/rtf
+    application/vnd.ms-excel
+    application/vnd.oasis.opendocument.text
+    application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
```
arthurashman, line 11: This also not allowed according to the copy (and Civil don't allow it)

app/validators/file_upload_validator.rb
```diff
@@ -0,0 +1,57 @@
...
+    application/vnd.openxmlformats-officedocument.wordprocessingml.document
+    image/bmp
+    image/gif
```
arthurashman, line 14: This also not allowed according to the copy (and Civil don't allow it)

app/validators/file_upload_validator.rb
arthurashman: Civil does allow (if that was the intention, to match what they do)

config/locales/en/steps.yml
```diff
@@ -184,6 +184,14 @@ en:
           page_title: Upload supporting evidence
           heading: Upload supporting evidence
           upload_file_button: Upload file
+          upload_files_heading: Upload files
+          upload_files_info: The maximum file size is 20MB. Files must be a DOC, DOCX, RTF, ODT, JPG, BMP, PNG, TIF or PDF.
+          choose_files_button: Choose files
+          uploaded_files_heading: Uploaded files
+          uploaded_files_info: When you submit this application, the caseworker will also get these files.
+          file_name: File name
+          uploaded: Uploaded
```
arthurashman, line 193: uploaded and uploading aren't used as far as I can tell

We could do with a translation for uploaded as it is hardcoded in `_uploaded_file.html.erb`

So:
```
file_name: File name
uploaded_file:
    uploaded: Uploaded
```

app/views/steps/evidence/upload/_uploaded_file.html.erb
```diff
@@ -0,0 +1,15 @@
+<tr class="govuk-table__row">
+  <td id="<%=document.id%>" class="govuk-table__cell">
+    <span><%= document.filename %></span>
+
+    <% if document.s3_object_key %>
+      <strong class="govuk-tag govuk-tag--green">Uploaded</strong>
```
arthurashman, line 6: If you add thee suggestion in the translation file, then we can do:
`      <strong class="govuk-tag govuk-tag--green"><%= t('.uploaded') %></strong>
`

arthurashman (approved):
This all looks great to me. 

Have obviously been using it lots in my own branch, so I'm also v happy that it all works as we want it to!

Nice work 😊

app/javascript/local/dropzone-cfg.js
```diff
@@ -12,18 +12,74 @@ DropzoneCfg.prototype.init = function () {
...
+    this.$tableCell = document.getElementById(file.upload.uuid)
+    this.$tableCell.append(error)
+    this.$statusTag = this.$tableCell.querySelector(".govuk-tag")
+    this.$tableCell.removeChild(this.$statusTag)
+  })
 }
+function createTableRow(file) {
+  let row = document.createElement("tr")
+  let cell = document.createElement("td")
+  let statusTag = createStatusTag("Uploading")
```
arthurashman, line 52: Yeah, this is hardcoded here, so I don't think you need it in the translation file

app/validators/file_upload_validator.rb
hiboabd: This was lifted from civil and not reviewed 😄 so there's a mismatch between the what the designs list as acceptable file types and we what we have in the backend. I'll clean it the list to keep it in line with design

app/validators/file_upload_validator.rb
hiboabd: I have no idea where this list came from because gifs is hilarious 😆

PR #478, opened 2023-09-19
[CRIMAP-596] Add ClamAV virus scanning

mseedat-moj:
## Description of change
Adds asynchronous virus scanning for uploaded files using `clamav`

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-596

## Notes for reviewer
1. Usage of the `clam_scan` gem is debatable as it has not been updated in 8 years (https://github.com/jschroeder9000/clam_scan/tree/master). An alternative library `clamby` has not been updated in 2 years (https://github.com/kobaltz/clamby) but it's argument parsing seemed to be less robust than `clam_scan`. There's a question mark about needing a library at all but I've decided to use one to have better argument parsing to prevent injection attacks e.g. through a cleverly crafted filename perhaps? Is a command line argument injection possible? `clam_scan` also auto-deletes dodgy files immediately whereas `clamby` does not.
2. Should files that are scanned with a positive virus match be deleted immediately? Removed this feature for now instead just log the possible violation to Sentry logs.
3. ~~The scanning only takes place on 'production' like environments not localhost. Is this a bad idea?~~
4. The clamav can be used as an executable but I've opted to use it as a daemon. This means additional startup overhead and additional memory usage. However it means quicker virus scanning and therefore from a user perspective more rapid feedback and less likely for request timeout. Need to check repercussions in production.
5. Additional work to ensure the clam database is updated will need to be done as well
6. Installing clam will increase Pod startup times quite drastically
7. Because each application server installs a copy of clam and the virus database, there will be a situation where application servers have different virus scan databases without a freschclam sync job in place.

N/A

Begin an application and upload a number of files on STAGING. Not sure how to test a false-positive file at the moment.

### DIFF
--- a/.github/workflows/test-build-deploy.yml
+++ b/.github/workflows/test-build-deploy.yml
@@ -61,6 +61,9 @@ jobs:
       - name: Setup test database
         run: bin/rails db:prepare
 
+      - name: Install clamav
+        run: sudo apt-get install clamav clamav-daemon freshclam
+
       - name: Run linters and tests
         run: bundle exec rake
 
--- a/.github/workflows/test-build-deploy.yml
+++ b/.github/workflows/test-build-deploy.yml
@@ -62,7 +62,7 @@ jobs:
         run: bin/rails db:prepare
 
       - name: Install clamav
-        run: sudo apt-get install clamav clamav-daemon freshclam
+        run: sudo apt-get install clamav clamav-daemon -y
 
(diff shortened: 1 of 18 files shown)


mseedat-moj:
Requires further discussion and testing prior to merging

mseedat-moj:
This work is superseded by #490

PR #483, opened 2023-09-21
[CRIMAP-620] Prevent legacy means_passport being reused

mseedat-moj:
## Description of change
Form reset when key details are changed did not change the application `means_passport` back to an empty array.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-620

## Notes for reviewer
Existing behaviour did not reset means_passport value meaning any legacy values persist even if other key details e.g. name, nino, etc are changed.

Create an application which has means_passport, change name details before submitting, after submitting means_passport should be blank (check database).

### DIFF
--- a/app/forms/steps/client/details_form.rb
+++ b/app/forms/steps/client/details_form.rb
@@ -27,8 +27,6 @@ module Steps
       def persist!
         return true unless changed?
 
-        crime_application.update(means_passport: [])
-
         applicant.update(
           attributes.merge(
             # The following are dependent attributes that need to be reset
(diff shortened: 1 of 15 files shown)


arthurashman (approved):
I think this would be good to get in and get on staging for some manual testing by everyone. Then on prod ASAP. 
But I wonder whether we should return to look at testing the resetting functionality as a tech debt point

app/forms/steps/client/details_form.rb
```diff
@@ -27,6 +27,8 @@ def date_of_birth_ten_or_more_years_ago
       def persist!
         return true unless changed?
 
+        crime_application.update(means_passport: [])
```
timpeat, line 30: Is there are reason not to add this to the attributes_to_reset hash ?

PR #487, opened 2023-09-27
CRIMAP-636 Implement end-of-journey fulfilment validations

zheileman:
## Description of change
Due to the loosely coupled nature of the step-by-step architecture where, as much as possible, each step is isolated from others and perform their own validations, and branching out can produce many different paths for a user, there may be rare situations (or by malice) where some important details of the application are not properly calculated.

Majority of these problems are caught by the JSON schema validation at point of submission, however there is not a proper user interface to present these errors.

This PR introduces the concept of "fulfilment validation" that happens at point of submission (technically right after declaration is signed) and raise any issues that hopefully can be recovered from by re-running some of the steps.

The validator allows to introduce further validations in the future, but for now we are focusing on the "passporters" (IoJ and Means).

If a problem is found, we present an error to the user. The error and design is TBD, in this PR most of the work is backend, but to "see something" I've added the errors to the page with some placeholder copy.

We also report to Sentry the problems to be alerted and take further measures if this is a reoccurring problem. Again is important to note the rare nature of these issues (unless tampering on purpose with the form).

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-636

Slack convo:
https://mojdt.slack.com/archives/C03JS4V9TPU/p1695656457701139

## Notes for reviewer
* As stated, copy is placeholder and possibly design may change.
* The "Change" link takes the provider to a meaningful page if possible. So for instance for the means passport it takes them to the Client Details page, and for the IoJ passport it takes them to IoJ page.
* At the moment there are 2 validations (ioj passport and means passport). If both fail we present 2 errors. In the future we could introduce more validations. We can decide if we want to show all errors or just one error, the first one, as in the end the provider will be able to action only one at a time (with the Change link).

Screenshot 2023-09-27 at 10 06 22
Screenshot 2023-09-26 at 16 29 14

Use instructions provided by @arthurashman on slack:
https://mojdt.slack.com/archives/C03JS4V9TPU/p1695634016744189

### DIFF
--- a/app/services/decisions/case_decision_tree.rb
+++ b/app/services/decisions/case_decision_tree.rb
@@ -101,22 +101,20 @@ module Decisions
     # NOTE: for MVP, this is the last step of the application,
     # however post-MVP there will be means assessment steps,
     # unless the applicant has been passported. As this is not
-    # yet implemented, and for MVP we only let means passported
-    # applicants, raise an exception (this should NOT happen but
-    # in case it happens, we get alerted).
+    # yet implemented, for now both branches behave the same.
+    # However before submission we perform a validation to
+    # ensure applicant is means-passported (array is not empty).
     #
-    # rubocop:disable Style/GuardClause
+    # rubocop:disable Style/IdenticalConditionalBranches,Lint/DuplicateBranch
     def after_ioj
       if Passporting::MeansPassporter.new(current_crime_application).call
         edit('/steps/submission/review')
       else
         # TODO: post-MVP implement means assessment steps
-        # For MVP, this branch should not be reachable, and if
-        # we end up here, we want to know about it (Sentry)
-        raise InvalidStep, 'application is not means-passported'
+        edit('/steps/submission/review')
       end
     end
-    # rubocop:enable Style/GuardClause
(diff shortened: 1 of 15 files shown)


app/services/decisions/case_decision_tree.rb
```diff
@@ -101,22 +101,20 @@ def after_hearing_details
...
-    # rubocop:disable Style/GuardClause
+    # rubocop:disable Style/IdenticalConditionalBranches,Lint/DuplicateBranch
     def after_ioj
       if Passporting::MeansPassporter.new(current_crime_application).call
         edit('/steps/submission/review')
       else
         # TODO: post-MVP implement means assessment steps
-        # For MVP, this branch should not be reachable, and if
-        # we end up here, we want to know about it (Sentry)
-        raise InvalidStep, 'application is not means-passported'
```
zheileman, line 116: I've removed the raised exception here, as we now want to catch the means passport problem with the new "pre-submission" validation, and present an error to the user to hopefully recover.
I've left on purpose the 2 branches even if they go to the same place for clarity and is silly.

timpeat (approved):
Elegant 🥇

PR #488, opened 2023-09-27
Refactor evidence data model removing bundle abstraction

zheileman:
## Description of change
Bit of a refactor before this becomes too entangled to remove the DocumentBundle intermediate table and abstraction as it seems is of no use and makes code a bit more convoluted and creates additional indirection.

To the best of my efforts everything should continue working as before, no changes to external URLs etc. have been made, and is all internal.

As part of this I've taken the opportunity to introduce in the Documents DB table 2 new attributes:

- `submitted_at` - can be populated on rehydration from a submitted application, and this can signalise which documents are already submitted and thus we shouldn't do a hard delete on S3 (just soft delete setting the `deleted_at` attribute)

- `annotations` - jsonb attribute that initially will be used for the virus scanner to annotate files that have been scanned and its result. Can be used for some additional functionality later on.

## Link to relevant ticket
This is part of the overall evidence upload work.

No visual changes.

Everything should keep working as before.

### DIFF
--- a/app/serializers/submission_serializer/sections/supporting_evidence.rb
+++ b/app/serializers/submission_serializer/sections/supporting_evidence.rb
@@ -3,20 +3,14 @@ module SubmissionSerializer
     class SupportingEvidence < Sections::BaseSection
       def to_builder
         Jbuilder.new do |json|
-          json.supporting_evidence Definitions::Document.generate(documents)
+          json.supporting_evidence Definitions::Document.generate(documents.stored)
         end
       end
 
       private
 
       def documents
-        # the upload page is not part of the flow yet so document bundles may not be initialised
-        if crime_application.document_bundles.first.nil?
-          @documents = {}
-        else
-          # we only want documents successfully uploaded to s3 to be serialised
-          @documents ||= crime_application.document_bundles.first.documents.stored
-        end
+        crime_application.documents
       end
     end
   end
--- a/config/locales/en/steps.yml
+++ b/config/locales/en/steps.yml
@@ -198,8 +198,8 @@ en:
           file_name: File name
           uploaded: Uploaded
           delete:
-            success: Document has been sucessfully deleted
-            failure: Document was not sucessfully deleted
+            success: Document has been successfully deleted
+            failure: Document was not successfully deleted
         uploaded_file:
           remove_button: Delete
     submission:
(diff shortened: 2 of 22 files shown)


config/locales/en/steps.yml
```diff
@@ -198,8 +198,8 @@ en:
           file_name: File name
           uploaded: Uploaded
           delete:
-            success: Document has been sucessfully deleted
-            failure: Document was not sucessfully deleted
+            success: Document has been successfully deleted
+            failure: Document was not successfully deleted
```
arthurashman, line 202: 😬 oops, thank you! Good spot

app/serializers/submission_serializer/sections/supporting_evidence.rb
```diff
@@ -3,20 +3,14 @@ module Sections
...
 
       def documents
-        # the upload page is not part of the flow yet so document bundles may not be initialised
-        if crime_application.document_bundles.first.nil?
-          @documents = {}
-        else
-          # we only want documents successfully uploaded to s3 to be serialised
-          @documents ||= crime_application.document_bundles.first.documents.stored
-        end
+        crime_application.documents
```
hiboabd, line 13: 😍

spec/controllers/steps/evidence/upload_controller_spec.rb
```diff
@@ -3,40 +3,36 @@
...
-
         allow_any_instance_of(Datastore::Documents::Delete).to receive(:call).and_return(false)
 
         expect(
           subject
         ).to receive(:update_and_advance).with(
           Steps::Evidence::UploadForm, as: :delete_document,
-          flash: { alert: 'Document was not sucessfully deleted' },
-          record: document.document_bundle
+          flash: { alert: 'Document was not successfully deleted' }
```
hiboabd, line 32: This might change right? Do you want to reference the actual i18n value

hiboabd (approved):
Non blocking comment but LGTM

spec/controllers/steps/evidence/upload_controller_spec.rb
zheileman: Didn't make major changes to what was done already.
In this case as it is a spec I'm not too worry of this having hardcoded copy. If copy changes the spec will fail and is easy to fix.

In any case the whole deletion will need to be refactored/improved as currently does hard delete on S3 and also tries to do so for not stored files (without s3 key) which fails again, and again 🥲

PR #489, opened 2023-09-28
[CRIMAP-595] Display supporting evidence on CYA page

hiboabd:


app/presenters/summary/html_presenter.rb
```diff
@@ -19,8 +19,9 @@ def sections
         Sections::JustificationForLegalAid.new(crime_application),
         Sections::PassportJustificationForLegalAid.new(crime_application),
         Sections::LegalRepresentativeDetails.new(crime_application),
+        Sections::SupportingEvidence.new(crime_application),
```
zheileman, line 22: Although for now it doesn't matter as it is feature-flagged, I think this section will go before the Legal rep details (which is the Declaration and thus the last step).

app/presenters/summary/sections/supporting_evidence.rb
```diff
@@ -0,0 +1,36 @@
+module Summary
+  module Sections
+    class SupportingEvidence < Sections::BaseSection
+      def name
+        :supporting_evidence
+      end
+
+      def show?
+        crime_application.documents.present? && super
```
zheileman, line 9: As we've defined a `documents` method, we could reuse it here:
`documents.any? && super`

app/services/adapters/structs/crime_application.rb
```diff
@@ -14,6 +14,13 @@ def case
       def ioj
         Structs::InterestsOfJustice.new(interests_of_justice)
       end
+
+      def documents
+        # TODO: Filter out deleted documents
```
zheileman, line 19: Seems like we will not get back deleted documents, thus no need to filter, so we can remove this comment I introduced in previous refactor 😄

zheileman (approved):
Nice. Just a couple minor suggestions!

app/presenters/summary/html_presenter.rb
hiboabd: Oh ok, will move it

PR #492, opened 2023-09-29
CRIMAP-576 Add `has_benefit_evidence` page

arthurashman:
## Description of change
If benefit check fails, details are correct, user can register that they have evidence they intend to upload later
Adds virtual attribute, only needed locally in apply to resurface later at evidence upload step.

FYI I have also removed the feature flag for the benefit type page - So this would make it a live service change on deploy to prod.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-576

## Notes for reviewer
As I am away this week, please feel free to commit any changes and merge as needed! I'm not precious!

### Before changes:

https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/assets/45827968/287006df-15c8-4241-8eed-2d2143a84460

### After changes:

Uploading Hasevidence.mov…

### DIFF
--- a/app/controllers/steps/client/evidence_exit_controller.rb
+++ b/app/controllers/steps/client/evidence_exit_controller.rb
@@ -1,0 +1,7 @@
+module Steps
+  module Client
+    class EvidenceExitController < Steps::ClientStepController
+      def show; end
+    end
+  end
+end
--- a/app/controllers/steps/client/has_benefit_evidence_controller.rb
+++ b/app/controllers/steps/client/has_benefit_evidence_controller.rb
@@ -2,18 +2,12 @@ module Steps
   module Client
     class HasBenefitEvidenceController < Steps::ClientStepController
       def edit
-        @form_object = HasBenefitEvidenceForm.new(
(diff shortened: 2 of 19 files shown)


mseedat-moj (approved):
Nice!

zheileman:
@arthurashman For reference:

I've reconciled this branch with latest main code (now released to prod).

I made a small change:

Removed for now the `expected_evidence` attribute, as it may be early to know how best to really use this (there were also no tests for this). I'm trying to avoid a situation similar to the `passporters` where the logic might sometimes not be run and we end up with out of sync values in the array 😞 

Perhaps a better approach is to always calculate the expected evidence based on the answers provided, instead of keeping a persisted list.
The answers are needed regardless because we need to rehydrate returned applications, thus why I've persisted `has_benefit_evidence` answer.
Currently is a bit pointless as answering NO exits the journey, but soon we will need to allow to proceed (means test) so we need this answer persisted to accurately rehydrate and re-play each answer given to the provider.

PR #495, opened 2023-10-02
Refactor clamav deployment to use StatefulSet

zheileman:
## Description of change
This ensures stickiness of persistent volumes on rollouts and scaling.

Delete unused file. We can access the service with the short name as it is inside the same namespace, no need to use the fully qualified name.

This is a follow-up to previous PR #490.


hiboabd (approved):
LGTM

PR #496, opened 2023-10-03
CRIMAP-646 Refactor deletion of evidence documents

zheileman:
## Description of change
This is a first pass on fixing/adapting the delete functionality to the latest agreed behaviour.

Evidence can be hard-deleted while it is still in draft (never submitted) mode. Once an application is submitted, its evidence can still be deleted in a resubmission, but is a soft-delete, meaning it remains in the S3 bucket as previous version(s) of the application will still reference these files.

In order to do this on rehydration we just set the `submitted_at` attribute in the Document object, as a way to signalise this was a previously submitted document.

Refactor the `Delete` service object to protect against accidental (hard)deletions of submitted files, and also to not try to delete files that have not been stored really (`s3_object_key` is nil). Better handling of errors.

Extracted the draft application purge functionality to a little service object that also performs cleanup of S3 documents if necessary (and according to rules above) before destroying the local application record along with all their relationships.
This is only for in-progress applications deleted through the dashboard.

## Link to relevant ticket
There is no specific ticket for this I believe, the closer I could find was this bug report:
https://dsdmoj.atlassian.net/browse/CRIMAP-646

Nothing visually has been changed.

Try with draft applications, and returned applications, etc. Deleted documents in a resubmission, will still be accessible on Review when accessing the version where originally were submitted.

So basically:
- on a draft application (never submitted before), if provider deletes this draft from their dashboard, we would be deleting all stored documents (as the application was never submitted nor will be after deleting it)
- on a draft application coming from a submission (a returned application, rehydrated) all documents will have the `submitted_at` as we set it on rehydration, if the provider uploads some more documents, but then never submits the application, delete it from dashboard, we will delete only stored documents with `submitted_at` nil (not previously submitted documents)
- on submission we will not delete anything as all stored will go to the datastore, and anything not stored (failed uploads for instance) doesn’t matter for clean up purposes. We don't use the purger here.

### DIFF
--- a/app/services/application_purger.rb
+++ b/app/services/application_purger.rb
@@ -1,0 +1,29 @@
+class ApplicationPurger
+  attr_reader :crime_application
+
+  def self.call(crime_application)
+    new(crime_application).call
+  end
+
+  def initialize(crime_application)
+    @crime_application = crime_application
+  end
+  private_class_method :new
+
+  def call
+    delete_orphan_stored_documents
+    crime_application.destroy!
+  end
+
+  private
+
+  def delete_orphan_stored_documents
+    orphan_documents.each do |document|
+      Datastore::Documents::Delete.new(document:).call
+    end
+  end
+
+  def orphan_documents
+    crime_application.documents.stored.not_submitted
+  end
+end
--- a/app/services/application_purger.rb
+++ b/app/services/application_purger.rb
@@ -11,18 +11,11 @@ class ApplicationPurger
   private_class_method :new
 
   def call
-    delete_orphan_stored_documents unless application_submitted?
+    delete_orphan_stored_documents
     crime_application.destroy!
   end
 
   private
-
-  # When an application is submitted to the datastore, right before
-  # deleting it from the local database along with all relationships,
-  # we consider successfully stored documents as 'submitted' as well
-  def application_submitted?
-    crime_application.submitted_at.present?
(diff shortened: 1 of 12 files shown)


app/services/application_purger.rb
```diff
@@ -0,0 +1,29 @@
+class ApplicationPurger
```
willmcb, line 1: Just wondering about the name - tiny thing but is it purging the application itself or just its evidence / documents. Perhasp `ApplicationEvidencePurger` might be clearer.

If you think this might be used to purge other things though I think makes sense as is.

Anywany, just a thought I don't have a massively strong opinion on this.

willmcb (approved):
lgtm :) 

Small non-blocking comment to think about 👍

app/services/application_purger.rb
zheileman: Is purging both. It is the replacement for previous `current_crime_application.destroy`

It encapsulates any cleanup needed. At the moment that is evidence (plus the application itself) but there might be other cleanup required in the future thus it is not only about evidence.

zheileman:
It requires a change to mark at submission time the documents as submitted or similar approach as to not being accidentally deleted on submission. Good catch by Hibo.
Will raise a change soon.

hiboabd (approved):
LGTM thanks

PR #497, opened 2023-10-03
CRIMAP-644 Show benefit type in CYA page

zheileman:
## Description of change
Added the benefit type to the check your answer page, in place of the previous "Passporting benefit" essentially merging both rows into one as confirmed by content designer.

If the question is not asked (under 18s) then the row will not show up.

Ensured `benefit_type` attribute is reset in a couple places if some key details are changed, as this value is linked to the DWP benefit checker.

Feature is still being feature flag, need to be removed if we want this in production at some point in the future once QA'ed.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-644
Slack thread: https://mojdt.slack.com/archives/C03JS4V9TPU/p1696328705798759

Screenshot 2023-10-03 at 14 33 09

For under 18s this row will not show.
For over 18s, run through the DWP check successfully and reach the CYA page, you should see whatever benefit was selected from the list presented.

### DIFF
--- a/app/controllers/developer_tools/crime_applications_controller.rb
+++ b/app/controllers/developer_tools/crime_applications_controller.rb
@@ -32,6 +32,7 @@ module DeveloperTools
           edit_steps_client_has_partner_path(crime_application),
           edit_steps_client_details_path(crime_application),
           edit_steps_client_has_nino_path(crime_application),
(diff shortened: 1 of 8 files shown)


hiboabd (approved):
Thanks LGTM

PR #501, opened 2023-10-04
Scope applications by selected office code

timpeat:
## Description of change
Scope current crime application by selected office code.
Allow host to be set in development from env.


zheileman (approved):
Awesome! thx!

PR #502, opened 2023-10-06
CRIMAP-651 Skip attack-sqli check on session cookie

timpeat:
## Description of change
Prevent sql injection false positives on the rails session cookie

## Link to relevant ticket

[CRIMAP-651](https://dsdmoj.atlassian.net/browse/CRIMAP-651)

## Notes for reviewer
We've identified a number of attack-sqli false +ve on the rails session cookie and confirmed this PR's changes work. After testing on staging, this change will need to be applied to the production ingress.


zheileman (approved):
Awesome!

PR #503, opened 2023-10-06
CRIMAP-576 Link evidence upload to wider journey

zheileman:
## Description of change
Create the task implementation, in the task list.

A basic wrapper to hold the logic on whether evidence is needed or not, and which types. This for now is up in the air so I've done it as basic as possible.

Link the evidence upload page to the journey, post IoJ, depending on the value of the previously asked question "has_benefit_evidence"

Modify the fulfilment validator because now it will be possible to have non means passported applications, but with evidence supplied.

No datastore/schema/rehydration yet.

Some unknowns yet to be confirmed/decided.

Note all this still remains behind feat flag, so nothing will show on production even if we release.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-576

Screenshot 2023-10-06 at 16 05 30

Run a DWP check that comes up without benefit, answer YES to the question about having evidence. Then proceed all the way to the end, where after the IoJ you will see the upload evidence page.
Also the task in the task list will show either not applicable, in progress or completed, depending on the state of the application.

### DIFF
--- a/app/services/evidence/requirements.rb
+++ b/app/services/evidence/requirements.rb
@@ -1,0 +1,35 @@
+module Evidence
+  class Requirements
+    attr_reader :crime_application
+
+    # NOTE: we don't have enough detail yet on how best to implement this.
+    # The assumption is, along the journey, there will be "check-points" or
+    # questions that will make certain evidence mandatory. For now we only
+    # have one question that will trigger the evidence upload page.
+    # This class is just a simple wrapper to start with.
+
+    def initialize(crime_application)
+      @crime_application = crime_application
+    end
+
+    def any?
+      [benefits?].any?
+    end
+
+    def none?
+      !any?
+    end
+
+    def benefits?
+      YesNoAnswer.new(
+        applicant&.has_benefit_evidence.to_s
+      ).yes?
+    end
+
+    private
+
+    def applicant
+      @applicant ||= crime_application.applicant
+    end
+  end
+end
--- a/app/validators/application_fulfilment_validator.rb
+++ b/app/validators/application_fulfilment_validator.rb
@@ -22,7 +22,7 @@ class ApplicationFulfilmentValidator < ActiveModel::Validator
   def perform_validations
     errors = []
 
-    unless Passporting::MeansPassporter.new(record).call || has_evidence?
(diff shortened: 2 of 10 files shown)


app/services/evidence/requirements.rb
```diff
@@ -0,0 +1,35 @@
+module Evidence
+  class Requirements
+    attr_reader :crime_application
+
+    # NOTE: we don't have enough detail yet on how best to implement this.
+    # The assumption is, along the journey, there will be "check-points" or
+    # questions that will make certain evidence mandatory. For now we only
+    # have one question that will trigger the evidence upload page.
+    # This class is just a simple wrapper to start with.
```
willmcb, line 9: 👍

willmcb:
Looks good - functions as expected for me 👍

willmcb (approved):

app/validators/application_fulfilment_validator.rb
```diff
@@ -22,7 +22,7 @@ def default_url_options
   def perform_validations
     errors = []
 
-    unless Passporting::MeansPassporter.new(record).call
+    unless Passporting::MeansPassporter.new(record).call || evidence_present?
```
arthurashman, line 25: If this is specific to means_passport, therefore evidence for passporting benefit, should we also be checking that the evidence requested was for benefits? 

Maybe just one for the future once there is an option to unlock the evidence upload step for a different reason, but remain non-means passported and never asked for benefit evidence

arthurashman (approved):
Just a thought on the final fulfilment validation re what evidence was requested

app/validators/application_fulfilment_validator.rb
zheileman: Yes, not sure currently how we would do this. We can only work with what we have, which is checking if some document has been "stored" (as in successfully being submitted to S3).

We can assume is in the best interest to the provider to upload what they are asked to upload. If they instead upload an IKEA assembly instructions document, application will probably be rejected by the case worker 😅

zheileman: Once we introduce evidence categorisation further validations might be unlocked 👍

PR #514, opened 2023-10-24
CRIMAP-647 Added new offence Intentional Strangulation

MohammedHassanLatif:
## Description of change
Added new offence Intentional Strangulation to list of existing offences

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAP-647

### Before changes:
image

### After changes:
Screenshot 2023-10-24 at 16 15 36

Screenshot 2023-10-24 at 16 16 51

- User creates a new application in Apply
- User gets to 'What has your client been charged with?' page
- User clicks on offence box to add an offence and enters 'Intentional Strangulation'
- 'Intentional Strangulation' will appear in the list of offences


arthurashman (commented):
Very nice @MohammedHassanLatif looks good to me. 

Just have a question for @willmcb - do we need to keep track of changes / put in a comment if we're adding a new offence and why? Or will having it version controlled be good enough for this?

arthurashman (approved):

PR #531, opened 2023-11-07
Crimap 676 add income model and bf tax form

willmcb:
## Description of change

Adds the "clients_income_before_tax"  to the income journey.

Also adds a new `IncomeDetails` object that we can use going forward to collect income related information in subsequent steps.

the relationship to crime_application is:

`crime_application.applicant.income_details`

i.e. it is a relationship through the applicant, but this is defined on the `Person` model so should be ameanable to `Partner` when the time comes to add those steps in as well. 

## Link to relevant ticket
[CRIMAP-676](https://dsdmoj.atlassian.net/browse/CRIMAP-676)

### Before changes:
n/a

### After changes:
Screenshot 2023-11-08 at 17 02 48

- Start an application and go through the first step so you enter client details.
- then head to `applications/<application id>/steps/income/clients_income_before_tax`
- try to submitt the form without aswering the question, you should get a validation error in this case.

You could also test the API functions correctly in the Rails console.

After creating the application locally run:
- `rails c` to open the console
- then you can see the relationships working as such

```
# i.e. the latest applicant - the one you just created
a = Applicant.last

# get their (albeit limited at the moment) income details
icd = a.income_details

# get current application from applicant
c = applicant.crime_application

# there is a convenience method to return all income_details related to a crime application
c.income_details 
>>>  [ <array of income details records> ]
```

[CRIMAP-676]: https://dsdmoj.atlassian.net/browse/CRIMAP-676?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/forms/steps/income/income_before_tax_form.rb
+++ b/app/forms/steps/income/income_before_tax_form.rb
@@ -1,0 +1,25 @@
+module Steps
+  module Income
+    class IncomeBeforeTaxForm < Steps::BaseFormObject
+      include Steps::HasOneAssociation
+      has_one_association :income_details, through: :applicant
+
+      # threshold being £12,475 currently
+      attribute :income_above_threshold, :value_object, source: YesNoAnswer
+
+      validates_inclusion_of :income_above_threshold, in: :choices
+
+      def choices
+        YesNoAnswer.values
+      end
+
+      private
+
+      def persist!
+        income_details.update(
+          attributes
+        )
+      end
+    end
+  end
+end
--- a/app/forms/steps/income/income_before_tax_form.rb
+++ b/app/forms/steps/income/income_before_tax_form.rb
@@ -4,7 +4,7 @@ module Steps
       include Steps::HasOneAssociation
       has_one_association :income_details, through: :applicant
 
-      # threshold being £12,475 currently
+      # threshold being £12,475 as of Nov 2023
       attribute :income_above_threshold, :value_object, source: YesNoAnswer
 
       validates_inclusion_of :income_above_threshold, in: :choices
--- a/app/services/decisions/income_decision_tree.rb
+++ b/app/services/decisions/income_decision_tree.rb
@@ -7,9 +7,12 @@ module Decisions
       when :lost_job_in_custody
         # TODO: link to next step when we have it
         edit(:manage_without_income)
-      when :manage_without_income
+      when :income_before_tax
         # TODO: link to next step when we have it
         show('/home', action: :index)
+      when :manage_without_income
+        # TODO: link to next step when we have it
+        show('/', action: :index)
       else
         raise InvalidStep, "Invalid step '#{step_name}'"
       end
--- a/app/services/decisions/income_decision_tree.rb
+++ b/app/services/decisions/income_decision_tree.rb
@@ -1,4 +1,7 @@
 module Decisions
+  # TODO: remove this rubocop opt out when
+  # branches are known and are therefore non-duplicate
+  # rubocop:disable Lint/DuplicateBranch
   class IncomeDecisionTree < BaseDecisionTree
     def destination
       case step_name
@@ -12,11 +15,12 @@ module Decisions
         show('/home', action: :index)
       when :manage_without_income
         # TODO: link to next step when we have it
-        show('/', action: :index)
+        show('/home', action: :index)
       else
         raise InvalidStep, "Invalid step '#{step_name}'"
       end
     end
+    # rubocop:enable Lint/DuplicateBranch
 
     private
 
--- a/app/views/steps/income/income_before_tax/edit.html.erb
+++ b/app/views/steps/income/income_before_tax/edit.html.erb
@@ -1,0 +1,17 @@
+<% title t('.page_title') %>
+<% step_header(path: crime_applications_path) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <span class="govuk-caption-xl"><%= t('steps.income.caption') %></span>
+    <%= govuk_error_summary(@form_object) %>
+    <%= step_form @form_object do |f| %>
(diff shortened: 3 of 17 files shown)


app/views/steps/income/income_before_tax/edit.html.erb
```diff
@@ -0,0 +1,17 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <span class="govuk-caption-xl"><%= t('steps.income.caption') %></span>
+    <%= govuk_error_summary(@form_object) %>
+    <%= step_form @form_object do |f| %>
+      <%= f.govuk_collection_radio_buttons :income_above_threshold, @form_object.choices,
+                                           :value, inline: false,
+                                           legend: { text: t('.income_above_threshold'), tag: 'h1', size: 'xl' } do %>
+          <p><%= t('.income_above_threshold_info') %></p>
```
hiboabd, line 12: Think this should be a caption
```suggestion
                  <span class="govuk-caption-m govuk-!-margin-bottom-4"><%= t('.income_above_threshold_info') %></span>
```

app/forms/steps/income/income_before_tax_form.rb
```diff
@@ -0,0 +1,25 @@
+module Steps
+  module Income
+    class IncomeBeforeTaxForm < Steps::BaseFormObject
+      include Steps::HasOneAssociation
+      has_one_association :income_details, through: :applicant
+
+      # threshold being £12,475 currently
```
hiboabd, line 7: Are we expecting this to change? Also shall we specify the time of the comment 'threshold being £12,475 as of Nov 2023'

db/migrate/20231107202222_create_income_details.rb
```diff
@@ -0,0 +1,13 @@
+class CreateIncomeDetails < ActiveRecord::Migration[7.0]
+  def change
+    create_table :income_details, id: :uuid do |t|
+      t.uuid :person_id, null: false
+      t.string :income_above_threshold
+
+      t.timestamps
```
hiboabd, line 7: Never seen [this](https://apidock.com/rails/ActiveRecord/ConnectionAdapters/Table/timestamps) before, this is handy!

hiboabd (changes requested):
Code looks good to me, just some small suggestions. When testing locally however, the validation check when the form is empty doesn't seem to be right
Screenshot 2023-11-09 at 11 09 15
t..

app/forms/steps/income/income_before_tax_form.rb
willmcb: Was thinking we could tell from the commit but will amend as that would be clearer 👍

db/migrate/20231107202222_create_income_details.rb
willmcb: Yep pretty useful :)

willmcb:
@hiboabd have pushed chages as per your review :)

hiboabd (approved):

spec/forms/steps/income/income_before_tax_form_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  let(:applicant) { instance_double(Applicant, income_details:) }
+  let(:income_details) { instance_double(IncomeDetails) }
+
+  let(:income_above_threshold) { nil }
+
+  describe '#choices' do
+    it 'returns the possible choices' do
+      expect(
+        subject.choices
+      ).to eq([YesNoAnswer::YES, YesNoAnswer::NO])
```
mseedat-moj, line 23: This is totally subjective, but I'm of the opinion in tests the 'expected' values should be hard-coded e.g. 'yes' to prevent circucar logic causing tests to pass but fail in production. It's really subjective tbh.

mseedat-moj (approved):
Nice one

app/forms/steps/income/income_before_tax_form.rb
```diff
@@ -0,0 +1,25 @@
+module Steps
+  module Income
+    class IncomeBeforeTaxForm < Steps::BaseFormObject
+      include Steps::HasOneAssociation
+      has_one_association :income_details, through: :applicant
+
+      # threshold being £12,475 as of Nov 2023
```
arthurashman, line 7: Is there any thought to us storing the number that we have asked them?

arthurashman: Unsure if there is any value to doing so. Mostly just thinking out loud

app/services/decisions/income_decision_tree.rb
```diff
@@ -1,4 +1,7 @@
 module Decisions
+  # TODO: remove this rubocop opt out when
```
arthurashman, line 2: We discussed just chaining all the pages together not worrying about order. The comment I put on line 11 was intending to say this as page it directs the user to on line 12 isn't correct as per thee flow. 

I'm not worried either way.

spec/forms/steps/income/income_before_tax_form_spec.rb
arthurashman: I kinda get what you mean @mseedat-moj but I don't think this is circular in this case. Especially if that YesNoAnswer Type is tested. My understanding is that part of the value of using the YesNoAnswer is we could change the values to True False and this would all be fine. And it's the YesNoAnswer spec that is responsible for making sure the values are correct

arthurashman (approved):
Looks great!

app/services/decisions/income_decision_tree.rb
willmcb: Can tidy up in future PR if its annoying - its all going to change anyway but I like the idea is gives us a "journey" to start working with.

PR #533, opened 2023-11-08
[CRIMAP-715] Remove BaseDecorator in favour of BasePresenter

MohammedHassanLatif:
## Description of change
Refactored code changed from BaseDecorator to BasePresenter. Removed other decorator code and files.

## Link to relevant ticket
https://dsdmoj.atlassian.net/jira/software/projects/CRIMAP/boards/897?selectedIssue=CRIMAP-715


app/decorators/address_form_decorator.rb
```diff
@@ -1,4 +1,4 @@
-class AddressFormDecorator < BaseDecorator
+class AddressFormDecorator < BasePresenter
```
willmcb, line 1: Why not make this a `AddressFormPresenter` as well that way we aren't mixing concepts

app/presenters/crime_application_presenter.rb
```diff
@@ -14,6 +14,8 @@ def applicant_dob
   end
 
   def applicant_name
+    return unless applicant?
```
willmcb, line 17: Is this a boy scout style addition unrelated to the general refactor you're doing here in relation to removing the concept of Decorators?

willmcb (changes requested):
Few minor things to think about - looks good though happy to merge once they have been looked at / we have had a chat.

hiboabd (approved):
Looks good to me! 👏

app/presenters/crime_application_presenter.rb
willmcb: This can be removed / put back to its orignal state

willmcb (approved):
Looks good :)

PR #535, opened 2023-11-10
Presenter bug fix

willmcb:
We had removed the decorator class but were still calling it when we should have been calling the "present" method in this partial.

Create a new applicaion locally
Go enter basic details and ten do the DWP bypass
When you click save and continue no exeption should be raised.


MohammedHassanLatif (approved):
Happy with the changes, thank you for catching this! @willmcb 😃

PR #572, opened 2023-12-05
Serialize client owns property

arthurashman:
## Description of change
Serialize client owns property

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-209

## Notes for reviewer
Needs schema release first

1. bundle install
2. Create a new application, fill all details but DO NOT SUBMIT
3. Go to /steps/income/what_is_clients_employment_status
4. Select 'My client is not working', Select 'No'
6. Click Save and continue
10. Select 'No' to £12,475
11. Click Save and continue
12. Select 'No' to frozen assets
13. Click Save and continue
14. Select 'No' to property
11. Click Save and continue
12. Select 'No' to savings
13. Select 'No' to dependants
11. Click Save and continue
12. Select any for how manage
13. Click Save and continue
14. Reopen that application
15. Click Confirm declaration and submit application
16. Make sure submits ok

### DIFF
--- a/spec/services/adapters/structs/income_details_spec.rb
+++ b/spec/services/adapters/structs/income_details_spec.rb
@@ -15,9 +15,12 @@ RSpec.describe Adapters::Structs::IncomeDetails do
           'ended_employment_within_three_months' => 'yes',
           'lost_job_in_custody' => 'yes',
           'date_job_lost' => Date.new(2023, 9, 1),
-          'income_above_threshold' => 'yes',
+          'income_above_threshold' => 'no',
           'has_frozen_income_or_assets' => 'no',
+          'client_owns_property' => 'no',
           'has_savings' => 'no',
+          'manage_without_income' => 'other',
+          'manage_other_details' => 'Another way they manage'
         )
       )
     end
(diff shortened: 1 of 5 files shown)


spec/services/adapters/structs/income_details_spec.rb
```diff
@@ -15,9 +15,12 @@
           'ended_employment_within_three_months' => 'yes',
           'lost_job_in_custody' => 'yes',
           'date_job_lost' => Date.new(2023, 9, 1),
-          'income_above_threshold' => 'yes',
+          'income_above_threshold' => 'no',
```
mseedat-moj, line 18: Is this a dramatic change? If so does there need to be a test for checking the various combinations of these fields?

arthurashman: Not dramatic. I changed it to be consistent with the actual flow - the user would not be asked the subsequent questions (frozen assets, has property, or has savings) if their income was above the threshold. 

I changed it on the fixture in the schema, so this did fail a test and needed changing to no IIRC

PR #576, opened 2023-12-12
CRIMAPP-210 rehydrate property

willmcb:
## Description of change

Adds "Does your client own their home, or any other land or property?" answer to the CYA page

### After changes:
Screenshot 2023-12-13 at 16 36 41

- create new application
- go through inital application stages
- click on the dev tools to go to the income pages answer in the following way:
1. What is your client's employment status? -> My Client is not working -> No
2. £12,475 a year before tax? -> No
3. assets under a restraint or freezing order? -> No
4. Does your client own their home, or any other land or property? -> No
5. Does your client have any savings or investments? -> No
6. Does your client have any dependants? -> No
7. How does your client manage with no income? -> select any
8. Continue should take you back to application list
9. Click on the application and go to "Review the application" in the task list
10. Scroll to income section, you should see the new "Does your client own their home, or any other land or property?" Question in the section.
11. Submit the application, then send it back, go to CYA page and check that it still shows there and that it has rehydrated correcly.
12. change its value and and go back to CYA to check that it reflected there too.

### DIFF
--- a/app/presenters/summary/sections/income_details.rb
+++ b/app/presenters/summary/sections/income_details.rb
@@ -9,7 +9,7 @@ module Summary
         income.present? && super
       end
 
-      def answers
+      def answers # rubocop:disable  Metrics/MethodLength
         [
           Components::ValueAnswer.new(
             :income_above_threshold, income.income_above_threshold,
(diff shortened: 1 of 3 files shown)


hiboabd (approved):
LGTM, worked for me locally 👍

PR #579, opened 2023-12-14
[CRIMAPP-98] Add payments page

arthurashman:
## Description of change
Introduces income payments page and related setup code

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-98

## Notes for reviewer
- the implementation overrides the current `Steps::BaseFormObject` implementation. This was necessary because of the way GovUKComponents generate checkbox field names, the use of the 'exclusive none' option, the need to output all possible payment types with pre-populated data where available, the need to have configurable sub-forms per payment type, the need to reference error fields in a non-standard way, the need maintain submitted data between invalid requests, etc. The general approach of 'virtual attributes' is still followed.
- Is storing 1 database row per payment type the correct approach? IoJ stores 1 row per crime application, which each justification specified as a database column. Dependants are stored 1 row per dependant record, with a potentially unlimited number of dependant entries. However there can only ever be 10 payment rows (10 different types). Therefore the database design could allow multiple entries per payment type - is validation through application code enough?
- The `IncomePaymentsForm` will always wipe then re-create the payment entries on save. This means all `metadata` has to be regenerated too, however the form does not do this at present, as the only `metadata` field is `details`. This could cause problems if metadata should maintain state between changes.
- Is the validation good enough?
- Ticket https://dsdmoj.atlassian.net/browse/CRIMAPP-466 raised to unify money conversion

#### Entry with validation errors showing which payment type was invalid
Screenshot 2024-02-08 at 12 12 35

#### Tampering with HTML shows validation error if attempting to manipulate metadata
Screenshot 2024-02-08 at 12 13 39

#### Database has 1 row per payment type, with metadata jsonb
Screenshot 2024-02-08 at 02 12 09

1. Create a new crime application
2. Navigate to `/applications/<crime application id>/steps/income/which_payments_does_client_get`
3. Complete form


spec/support/shared_examples/payment_shared_examples.rb
```diff
@@ -0,0 +1,100 @@
...
+    end
+
+    context 'when not a valid frequency' do
+      let(:frequency) { 'every other day' }
+
+      it { is_expected.not_to be_valid }
+      it { expect(subject.errors.of_kind?(:frequency, :inclusion)).to be(true) }
+    end
+  end
+end
```
mseedat-moj, line 100: Made testing the fieldset deliberately 'shared' to allow quick testing of new Benefits and Child Maintenance pages. Subsequent PR will make the main form also a shared test to ensure general behaviours remain consistent across all similar 'checkbox list' based payment pages.

db/schema.rb
```diff
@@ -139,7 +150,6 @@
     t.string "ended_employment_within_three_months"
     t.string "client_has_dependants"
     t.string "has_savings"
-    t.string "payments", default: [], array: true
```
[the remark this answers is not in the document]
mseedat-moj: Checked production and staging servers. Safe to delete this entry in schema.rb

app/views/steps/income/income_payments/edit.html.erb
```diff
@@ -0,0 +1,38 @@
...
+    <span class="govuk-caption-xl"><%= t('steps.income.caption') %></span>
+    <%= govuk_error_summary(@form_object) %>
+
+    <%= step_form @form_object do |f| %>
+      <%= f.govuk_check_boxes_fieldset :income_payments, legend: { tag: 'h1', size: 'xl' } do |x| %>
+
+        <% @form_object.ordered_payment_types.each do |type| %>
+          <%= f.govuk_check_box :types, type.to_s, checked: f.object.checked?(type), link_errors: true do %>
+            <%= f.fields_for type.to_sym, f.object.public_send(type.to_s) do |g| %>
+              <%= g.govuk_fieldset legend: nil, id: "income_payment_#{type}" do %>
```
hiboabd, line 15: Does the legend have to be nil or can it just not be set?

config/locales/en/errors.yml
```diff
@@ -377,6 +377,18 @@ en:
               blank: "Enter %{ordinal} dependant's age"
               less_than: "The %{ordinal} dependant must be under 18 years old"
               greater_than_or_equal_to: "The %{ordinal} dependant must be at least 0 years old"
+        steps/income/income_payment_fieldset_form:
+          summary:
+            amount_in_pounds:
+              greater_than: "Amount for %{payment_type} must be more than 0"
+            details:
+              invalid: "Cannot add additional details for %{payment_type}"
```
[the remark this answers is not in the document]
mseedat-moj: It's an error message that only shows if a user decides to tamper with the frontend HTML and try to 'hack' the form to send through 'details' that is only applicable if the user selected 'other'

app/views/steps/income/income_payments/edit.html.erb
mseedat-moj: Without the fieldset the form field naming scheme changes and therefore the form breaks

spec/forms/steps/income/income_payments_form_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+        )
+      end
+
+      before do
+        subject.valid?
+      end
+
+      it 'saves nothing' do
+        expect(subject.crime_application.income_payments.size).to eq 0
+        expect(subject.save).to be true # Always true
```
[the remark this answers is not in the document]
mseedat-moj: Good question - it's deliberate to show the form is a departure from other implementations in that the save will always occur, because by this point the actual records have already been inserted into `income_payments` and calling `subject.save` is part of the general form object lifecycle. Hope that makes sense. Do you think a fuller comment would be useful in the form code itself?

app/views/steps/income/income_payments/edit.html.erb
RoseSAK: Should the legend actually be nil or just be hidden? Just wondering if this may cause problems with accessibility. It's quite a complex form so wondering if it still makes sense without any legend or if it would be better to have `#{payment_type} details` or something like this, just a suggestion though

config/locales/en/helpers.yml
```diff
@@ -77,8 +86,8 @@ en:
       steps_outgoings_council_tax_form:
         council_tax: Does your client pay Council Tax where they usually live?
       steps_submission_more_information_form:
-        need_more_information: Do you need to add any more information to this application? 
-        
+        need_more_information: Do you need to add any more information to this application?
+
```
arthurashman, line 90: Delete this extra line

mseedat-moj (approved):
Approving on behalf of @arthurashman

PR #581, opened 2023-12-18
Create outgoings table, add outgoings more than income page

arthurashman:
## Description of change
Create outgoings table, add outgoings more than income page

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-226

image

1. rails db:migrate
2. start a new application
3. go to steps/outgoings/are_clients_outgoings_more_than_income
4. submit with no selection to check errors https://docs.google.com/spreadsheets/d/1gdDpuyqWO6MJgUT1SOy0nbHd8AAdp1YebBzh1WDG040/edit#gid=0&range=B58:B59
5. select yes and input no text, submit to check errors
6. enter text and submit to check submits and doesn't raise an error

### DIFF
--- a/app/services/decisions/outgoings_decision_tree.rb
+++ b/app/services/decisions/outgoings_decision_tree.rb
@@ -1,0 +1,13 @@
+module Decisions
+  class OutgoingsDecisionTree < BaseDecisionTree
+    def destination
+      case step_name
+      when :outgoings_more_than_income
+        # TODO: link to next step when we have it
+        show('/home', action: :index)
+      else
+        raise InvalidStep, "Invalid step '#{step_name}'"
+      end
+    end
+  end
+end
--- a/app/views/steps/outgoings/outgoings_more_than_income/edit.html.erb
+++ b/app/views/steps/outgoings/outgoings_more_than_income/edit.html.erb
@@ -1,0 +1,26 @@
+<% title t('.page_title') %>
+<% step_header(path: crime_applications_path) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <span class="govuk-caption-xl"><%= t('steps.outgoings.caption') %></span>
+     <%= step_form @form_object do |f| %>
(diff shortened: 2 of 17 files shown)


app/services/decisions/outgoings_decision_tree.rb
```diff
@@ -0,0 +1,13 @@
+module Decisions
+  class OutgoingsDecisionTree < BaseDecisionTree
+    def destination
+      case step_name
+      when :outgoings_more_than_income
+        # TODO: link to next step when we have it
```
willmcb, line 6: 👍

app/views/steps/outgoings/outgoings_more_than_income/edit.html.erb
```diff
@@ -0,0 +1,26 @@
+<% title t('.page_title') %>
+<% step_header(path: crime_applications_path) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <span class="govuk-caption-xl"><%= t('steps.outgoings.caption') %></span>
+     <%= step_form @form_object do |f| %>
+        <%= f.govuk_radio_buttons_fieldset :outgoings_more_than_income, legend: { text: t('.outgoings_more_than_income'), tag: 'h1', size: 'xl' } do %>
+          <% @form_object.choices.each_with_index do |choice, index| %>
+            <% if choice == YesNoAnswer::YES %>
```
willmcb, line 10: Don't think we need to the literal comparison like this as you can ask a "choice" what it is directly:

e.g.

`choice.yes?`
>>> true

willmcb (approved):
Small comment around the choice logic in the view that would be good to change, other than this looks great :)

app/views/steps/outgoings/outgoings_more_than_income/edit.html.erb
arthurashman: Roger that. I knew this was how it's set up, hadn't worked in a previous view (which it is now) I think bc I hadn't hooked up the types correctly and had copied the same logic over. All working now, all good - thanks for pointing out

PR #582, opened 2023-12-19
BUGFIX: Fix back links for income/outgoings journey

willmcb:
## Description of change

Fixes the back link for the income and new outgoings journeys. Was just a copy paste error. The `step_header` helper accepts a path as an overide which can be used on pages where the back link just needs to go back to the landing page or some other page when the user isn't in an application journey.

Issue was that we were passing `<% step_header(path: crime_applications_path) %>` on the steps and overriding the default behaviour of utilising the navigation stack to get the back link.

## Link to relevant ticket
n/a

- check back links on the current live journey they should be uneffected.
- start a new application, go past the initial client details page, then use dev tools sto start the income journey.
- go through the income journey, and after you submit each step click the back link, and make sure it takes you back to the last step you were on.

### DIFF
--- a/app/views/steps/income/client_has_dependants/edit.html.erb
+++ b/app/views/steps/income/client_has_dependants/edit.html.erb
@@ -1,5 +1,5 @@
 <% title t('.page_title') %>
-<% step_header(path: crime_applications_path) %>
+<% step_header %>
 
 <div class="govuk-grid-row">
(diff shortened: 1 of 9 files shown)


arthurashman:
Oop, thanks very much. I cast an eye over this in my last ticket and wondered the difference. Should have looked into it properly. This is great, thanks 👍

arthurashman (approved):

PR #583, opened 2023-12-19
Serialize outgoings, with first attrs

arthurashman:
## Description of change
Serialize outgoings, with first attrs

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-227
## Notes for reviewer
outgoings array is hardcoded as this ticket is not for that attribute, but the schema requires something. It will be an empty array if no options are selected (or maybe ['none']

1. `bundle install`
2. Create a new application, fill all details but DO NOT SUBMIT
3. Go to /steps/outgoings/are_clients_outgoings_more_than_income
4. Select 'Yes'
5. Enter some text in the text area
6. Click Save and continue
7. Reopen that application
8. Click Confirm declaration and submit application
9. Make sure submits ok
10. Click View completed application
11. Copy the UUID from the url
12. Run `rails c` in your datastore locally
13. Enter `a = CrimeApplication.find('UUID')`
14. Enter a.submitted_application
15. Make sure application has a means_details>outgoings_details

### DIFF
--- a/app/serializers/submission_serializer/sections/means_details.rb
+++ b/app/serializers/submission_serializer/sections/means_details.rb
@@ -20,6 +20,14 @@ module SubmissionSerializer
                 json.manage_without_income income.manage_without_income
                 json.manage_other_details income.manage_other_details
               end
+
+              json.outgoings_details do
+                # TODO: Update to take array from outgoings payments when we get
+                # there - needs to default to []
+                json.outgoings []
+                json.outgoings_more_than_income outgoings&.outgoings_more_than_income
+                json.how_manage outgoings&.how_manage
+              end
             end
           end
         end
@@ -30,6 +38,10 @@ module SubmissionSerializer
 
       def income
         @income ||= crime_application.income
+      end
+
+      def outgoings
+        @outgoings ||= crime_application.outgoings
       end
     end
   end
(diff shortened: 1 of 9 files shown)


spec/services/datastore/application_submission_spec.rb
```diff
@@ -76,10 +76,16 @@
       types: ['loss_of_liberty'],
       loss_of_liberty_justification: 'Details about loss of liberty.',
     )
+
```
arthurashman, line 79: For consistency

app/serializers/submission_serializer/sections/means_details.rb
```diff
@@ -31,6 +39,10 @@ def to_builder
       def income
         @income ||= crime_application.income
       end
+
+      def outgoings
+        @outgoings ||= crime_application.outgoings
```
mseedat-moj, line 44: Does this method need to default to an empty array something like:

```
@outgoings ||= (crime_application.outgoings || [])
```

Or will `crime_application.outgoings` always be `not nil`

mseedat-moj (approved):
Looks fine - minor comment regarding possible nil values

app/serializers/submission_serializer/sections/means_details.rb
arthurashman: The schema will happily accept it as nil values, so I chose to handle that on each attr with the safe navigation operator. 
My understanding is that if they have gone down the means route, and have got an `income` for which there is a guard clause at the top of this file, then they will have an outgoings

PR #586, opened 2023-12-21
[CRIMAP-295] Re-arrange rehydration of dependants

mseedat-moj:
## Description of change
Refactors `dependants` rehydration which breaks the normal pattern of rehydration due to the split nature of attributes and need to calculate one of the attributes.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-295

## Notes for reviewer
Rehydration of `dependants is complicated by the following:
- `dependants` are very optional because the 'income and outgoings' feature is still WIP
- there are 2 pieces of data - `dependants` and `client_has_dependants`. Schema has no knowledge of `client_has_dependants`.
- in Apply, `client_has_dependants` is stored in the `income` table, generally rehydrated by the `income_details` field from Schema. But `income_details` does not contain `dependants` from which the value can be calculated...
- because in Schema `dependants` is an attribute in `means_details`
- As a result this rigmarole is required.

### Consider:
- Is this a good idea?
- the code in `CrimeApplication.income` which performs some logic
- because `dependants` is optional, the calculated field `income.client_has_dependants` can have `yes/no/nil`
- `Structs::IncomeDetails` contains an `attr_accessor :client_has_dependants` to mirror the Apply database schema.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -47,7 +47,7 @@ gem 'laa-criminal-applications-datastore-api-client',
     require: 'datastore_api'
 
 gem 'laa-criminal-legal-aid-schemas',
-    github: 'ministryofjustice/laa-criminal-legal-aid-schemas', tag: 'v1.0.21'
+    github: 'ministryofjustice/laa-criminal-legal-aid-schemas', tag: 'v1.0.22'
 
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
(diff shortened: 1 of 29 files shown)


arthurashman:
Hey @mseedat-moj  - First thoughts are 100% change the schema to suit out domain needs instead of any jiggery-pokery. I think Tim was working on it ahead of knowing what we needed or what it should look like. Maybe just check in with him on his logic/needs for it to not be in the income.

arthurashman:
@mseedat-moj - I'm going to close this just to tidy up. Hibo did the schema shuffle around and the submission and rehydration over Christmas

PR #589, opened 2023-12-27
[CRIMAPP-286] Serialise and rehydrate dependants

hiboabd:
## Description of change
Fixes the submission and rehydration of applications with dependants data following schema change to move dependants from `means_details` to `income_details`


app/services/datastore/application_rehydration.rb
```diff
@@ -73,11 +73,11 @@ def income
     end
 
     def dependants
-      parent.means_details&.dependants&.map { |struct| Dependant.new(**struct) } || []
+      parent.means_details&.income_details&.dependants&.map { |struct| Dependant.new(**struct) } || []
     end
 
     def client_has_dependants
-      parent.means_details&.dependants&.any? ? YesNoAnswer::YES : YesNoAnswer::NO
+      parent.means_details&.income_details&.dependants&.any? ? YesNoAnswer::YES : YesNoAnswer::NO
```
arthurashman, line 80: I'm not sure I can 100% guarantee this, but if there is a means_details there should always be an income_details, so we wouldn't need the second null safety check

app/services/datastore/application_rehydration.rb
```diff
@@ -73,11 +73,11 @@ def income
     end
 
     def dependants
-      parent.means_details&.dependants&.map { |struct| Dependant.new(**struct) } || []
+      parent.means_details&.income_details&.dependants&.map { |struct| Dependant.new(**struct) } || []
```
arthurashman, line 76: I started looking at this myself before the break, and had extracted this out into `app/services/adapters/structs/income_details.rb` and then removed `dependants: dependants,` from above because this is then a part of the income object in apply as well as the datastore

arthurashman: You were right all along, sorry @hiboabd !

arthurashman (approved):
Nice, thanks so much for sorting 👍

PR #591, opened 2024-01-03
[CRIMAPP-265] Add council tax page

arthurashman:
## Description of change
Add council tax page

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-265

image

1. rails db:migrate
2. Start an application
3. Go to steps/outgoings/does_client_pay_council_tax
4. Check that submitting the form without selecting an option displays error 'Select yes if your client pays Council Tax where they usually live'
5. Check that submitting the form with Yes selected but no amount number input displays error 'Enter how much your client pays yearly'
6. Check that submitting with Yes selected and an amount input submits ok

### DIFF
--- a/app/forms/steps/outgoings/council_tax_form.rb
+++ b/app/forms/steps/outgoings/council_tax_form.rb
@@ -29,9 +29,9 @@ module Steps
       def council_tax_amount_in_pounds
         return unless council_tax_amount
 
-        amount_in_pence = council_tax_amount.dup
+        amount_in_pounds = council_tax_amount.dup / 100.0
 
-        amount_in_pence / 100.0
+        helpers.number_with_precision(amount_in_pounds, precision: 2)
       end
 
       private
@@ -50,6 +50,10 @@ module Steps
 
       def pays_council_tax?
         pays_council_tax&.yes?
+      end
+
+      def helpers
+        ActionController::Base.helpers
       end
     end
   end
--- a/app/forms/steps/outgoings/council_tax_form.rb
+++ b/app/forms/steps/outgoings/council_tax_form.rb
@@ -1,0 +1,40 @@
+module Steps
+  module Outgoings
+    class CouncilTaxForm < Steps::BaseFormObject
+      include Steps::HasOneAssociation
+      has_one_association :outgoings
+
+      attribute :pays_council_tax, :value_object, source: YesNoAnswer
+      attribute :council_tax_amount, :integer
+
+      validates_inclusion_of :pays_council_tax, in: :choices
+
+      validates :council_tax_amount,
+                presence: true,
+                numericality: { only_integer: true },
+                if: -> { pays_council_tax? }
+
+      def choices
+        YesNoAnswer.values
+      end
+
+      private
+
+      def persist!
+        outgoings.update(
+          attributes.merge(attributes_to_reset)
+        )
+      end
+
+      def attributes_to_reset
+        {
+          'council_tax_amount' => (council_tax_amount if pays_council_tax?)
+        }
+      end
+
+      def pays_council_tax?
+        pays_council_tax&.yes?
+      end
+    end
+  end
+end
--- a/app/forms/steps/outgoings/council_tax_form.rb
+++ b/app/forms/steps/outgoings/council_tax_form.rb
@@ -4,23 +4,34 @@ module Steps
       include Steps::HasOneAssociation
       has_one_association :outgoings
 
-      PENCE_REGEXP = /\A\d+(\.\d{0,2})?\z/
-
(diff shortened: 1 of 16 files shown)


app/services/decisions/income_decision_tree.rb
```diff
@@ -24,10 +24,11 @@ def destination
       when :delete_dependant
         edit_dependants
       when :dependants_finished
-        edit('/home', action: :index)
+        # TODO: link to next step when we have it
+        edit(:manage_without_income)
       when :manage_without_income
         # TODO: link to next step when we have it
-        show('/home', action: :index)
```
arthurashman, line 30: Just to keep user testers in flow whilst this is all in flux

config/locales/en/errors.yml
```diff
@@ -323,6 +323,17 @@ en:
...
+          attributes:
+            housing_payment_type:
+              inclusion: Select which payment your client pays where they usually live
+        steps/outgoings/council_tax_form:
+          attributes:
+            pays_council_tax:
+              inclusion: Select yes if your client pays Council Tax where they usually live
+            council_tax_amount:
+              blank: Enter how much your client pays yearly
+              not_a_number: How much they pay yearly must be a number, like 50
```
arthurashman, line 336: I don't know how to actually hit this as a user, especially with the field in the html set as a `govuk_number_field`, _and_ `inputmode: 'numeric'` 
 It does show if I remove `presence: true ` from the `council_tax_amount` validation in the form
 
I also can't hit it in the spec when passing in a string. 

Don't think this is a huge issue as the form UI doesn't allow for non-numeric input

app/services/decisions/income_decision_tree.rb
```diff
@@ -24,7 +24,8 @@ def destination
       when :delete_dependant
         edit_dependants
       when :dependants_finished
-        edit('/home', action: :index)
```
arthurashman, line 27: Just to link everything together

app/views/steps/income/lost_job_in_custody/edit.html.erb
```diff
@@ -3,26 +3,28 @@
```
arthurashman, line 3: Sorry about this one, just fixing indenting!

app/forms/steps/outgoings/council_tax_form.rb
```diff
@@ -0,0 +1,40 @@
...
+    class CouncilTaxForm < Steps::BaseFormObject
+      include Steps::HasOneAssociation
+      has_one_association :outgoings
+
+      attribute :pays_council_tax, :value_object, source: YesNoAnswer
+      attribute :council_tax_amount, :integer
+
+      validates_inclusion_of :pays_council_tax, in: :choices
+
+      validates :council_tax_amount,
```
willmcb, line 12: One thing you can do here is say 'Yes' to paying council tax and enter zero and the form submits, thinking that may be good to have a validation that the value entered needs to be > 0? 🤔

config/locales/en/errors.yml
willmcb: I don't think you can get to it via the UI

app/forms/steps/outgoings/council_tax_form.rb
```diff
@@ -0,0 +1,40 @@
+module Steps
+  module Outgoings
+    class CouncilTaxForm < Steps::BaseFormObject
+      include Steps::HasOneAssociation
+      has_one_association :outgoings
+
+      attribute :pays_council_tax, :value_object, source: YesNoAnswer
+      attribute :council_tax_amount, :integer
```
willmcb, line 8: One thing here is that by having it as an integer we lose pence amounts. Not sure how much of an issue this is but you can input pounds and pence and submit and then when you come back the pence amount after the decimal point is lost.

willmcb (changes requested):
Would like to discuss the comments here if thats okay - they are only minor

app/forms/steps/outgoings/council_tax_form.rb
arthurashman: 👍 Now negative or 0 doesn't pass validation and shows error message

app/forms/steps/outgoings/council_tax_form.rb
arthurashman: 👍 Changed to string here, and in db table. Will convert to pence integer in the serialization before sending to datastore, and back again on rehydration

arthurashman: Storing now as council_tax_amount in pence

willmcb (approved):
Have rerun a regression and tested the decimal issue and it all works great :)

PR #593, opened 2024-01-04
Show DWP bypass in Dev post case type move

willmcb:
## Description of change
Adds DWP bypass back in post the case type move.

Does a check on charge as this is the next mandatory data point we can use to remove it when it is not relevant.

No strong feelings either way - guess it could get more crowded if we add more buttons etc.

- DWP bypass button should be back now when you start the case.
- it will disappear on the offences page as its not relevant after the DWP page of the journey


arthurashman (commented):
👍 thank you!

arthurashman (approved):

PR #599, opened 2024-01-10
Add is application means tested page

arthurashman:
## Description of change
Add Yes/No question at start of new application journey to allow for routing of non-means tested applications. This PR is just for the page, not for the routing changes. 

Only routing currently done is to skip partner question if select no to means tested question, next page is the tasklist. Partner values are not relevant for non-means tested applications. If selecting yes to means test question, next page is has partner. 

This is behind the means-journey feature flag for now, but can be removed once the routing logic had been implemented. 

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-313

## Notes for reviewer

Subsequent PR for submitting to datastore and rehydrating
Subsequent PR for updating routing based on a non-means tested journey

### After changes:
image

1. Start a new application
2. First page should be means test. Check design/copy, check errors for when none selected. Check submits. 
3. Select Yes
4. Click save and continue
5. Ensure next page is has_partner
6. Go back
7. Select No
8. Click save and continue
9. Ensure next page is tasklist

10. Shut down server
11. Turn off means_journey feature flag locally, restart server
12. Start a new application
13. Ensure first page is has_partner


willmcb:
Looks like the back link on the partner page has to be updated. As this was the first page it meant the only place to go back was the index page but now, it needs to go to means and the means question needs to go to the index.

This link also needs to be feature flagged.

The routing looks good 
- yes --> partner 👍 
- no  --> tasklist 👍 

Feature Flag set to false sets application back to how it currently appears.

willmcb (changes requested):
See comment :)

willmcb (approved):
👍

PR #601, opened 2024-01-11
PSE Spike: with added pse fields

hiboabd:
## Description of change
PSE spike with added PSE fields, requires changes to datastore and schema
PRs: https://github.com/ministryofjustice/laa-criminal-applications-datastore/pull/199 
https://github.com/ministryofjustice/laa-criminal-legal-aid-schemas/pull/76


app/presenters/summary/html_presenter.rb
```diff
@@ -8,24 +8,33 @@ def initialize(crime_application:)
 
     # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
```
timpeat, line 9: use similar approach as adapter.

PR #602, opened 2024-01-12
[CRIMAPP-313] Update non means tested flow

arthurashman:
## Description of change
If an application is not means tested, it:

- skips case type page
- has an optional NINO
- skips benefit type and benefit check

Much of the code assumes that case type will be present, so a lot of this PR is to allow for it to be nil

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-313
https://dsdmoj.atlassian.net/browse/CRIMAPP-320
https://dsdmoj.atlassian.net/browse/CRIMAPP-321

### DIFF
--- a/app/forms/steps/client/has_nino_form.rb
+++ b/app/forms/steps/client/has_nino_form.rb
@@ -7,13 +7,21 @@ module Steps
       NINO_REGEXP = /\A(?!BG)(?!GB)(?!NK)(?!KN)(?!TN)(?!NT)(?!ZZ)[A-CEGHJ-PR-TW-Z][A-CEGHJ-NPR-TW-Z][0-9]{6}([A-DFM])\z/
 
       attribute :nino, :string
-      validates :nino, format: { with: NINO_REGEXP }
+      validate :validate_nino
 
       def nino=(str)
         super(str.upcase.delete(' ')) if str
       end
 
       private
+
+      def validate_nino
+        if nino.blank?
+          crime_application.not_means_tested? ? return : errors.add(:nino, :blank)
+        end
+
+        errors.add(:nino, :invalid) unless NINO_REGEXP.match?(nino)
+      end
 
       def persist!
         return true unless changed?
--- a/app/models/concerns/passportable.rb
+++ b/app/models/concerns/passportable.rb
@@ -2,7 +2,8 @@ module Passportable
   extend ActiveSupport::Concern
 
   delegate :age_passported?,
-           :benefit_check_passported?, to: :means_passporter
+           :benefit_check_passported?,
+           :not_means_tested?, to: :means_passporter
 
   def means_passported?
     means_passporter.passported?
--- a/app/presenters/summary/components/base_answer.rb
+++ b/app/presenters/summary/components/base_answer.rb
@@ -16,7 +16,7 @@ module Summary
       end
 
       def show?
-        show.presence || value?
+        show.nil? ? value? : show
       end
 
       def value?
--- a/app/presenters/summary/sections/case_details.rb
+++ b/app/presenters/summary/sections/case_details.rb
@@ -10,7 +10,6 @@ module Summary
         [
           Components::ValueAnswer.new(
             :case_type, kase.case_type,
-            show: crime_application.means_passport.exclude?('on_not_means_tested'),
             change_path: edit_steps_client_case_type_path
           ),
 
--- a/app/presenters/summary/sections/case_details.rb
+++ b/app/presenters/summary/sections/case_details.rb
@@ -10,6 +10,7 @@ module Summary
         [
           Components::ValueAnswer.new(
             :case_type, kase.case_type,
+            show: crime_application.means_passport.exclude?('on_not_means_tested'),
             change_path: edit_steps_client_case_type_path
           ),
 
--- a/app/presenters/summary/sections/client_details.rb
+++ b/app/presenters/summary/sections/client_details.rb
@@ -30,7 +30,7 @@ module Summary
 
           Components::FreeTextAnswer.new(
             :nino, applicant.nino,
-            change_path: edit_steps_client_has_nino_path
+            change_path: edit_steps_client_has_nino_path, show: true
           ),
 
           Components::ValueAnswer.new(
--- a/app/presenters/summary/sections/overview.rb
+++ b/app/presenters/summary/sections/overview.rb
@@ -2,36 +2,50 @@ module Summary
   module Sections
     class Overview < Sections::BaseSection
       def show?
-        !crime_application.in_progress? && super
+        crime_application.case.present? && super
       end
 
-      # rubocop:disable Metrics/MethodLength
+      # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
       def answers
-        [
-          Components::FreeTextAnswer.new(
-            :reference, crime_application.reference.to_s,
-          ),
+        relevant_answers =
+          [
+            Components::FreeTextAnswer.new(
+              :reference, crime_application.reference.to_s,
+            ),
 
-          Components::DateAnswer.new(
-            :date_stamp, crime_application.date_stamp,
-            i18n_opts: { format: :datetime },
-          ),
+            Components::ValueAnswer.new(
+              :means_tested, crime_application.is_means_tested,
+              change_path: edit_steps_client_is_means_tested_path
+            ),
 
-          Components::DateAnswer.new(
-            :date_submitted, crime_application.submitted_at,
-            i18n_opts: { format: :datetime },
-          ),
+            Components::DateAnswer.new(
+              :date_stamp, crime_application.date_stamp,
+              i18n_opts: { format: :datetime },
+            ),
+          ]
 
-          Components::FreeTextAnswer.new(
-            :provider_email, provider_details.provider_email,
-          ),
+        unless crime_application.in_progress?
+          completed_answers =
+            [
+              Components::DateAnswer.new(
+                :date_submitted, crime_application.submitted_at,
+                i18n_opts: { format: :datetime },
+              ),
 
-          Components::FreeTextAnswer.new(
-            :office_code, provider_details.office_code,
-          ),
-        ].select(&:show?)
+              Components::FreeTextAnswer.new(
+                :provider_email, provider_details.provider_email,
+              ),
+
+              Components::FreeTextAnswer.new(
+                :office_code, provider_details.office_code,
+              ),
+            ]
+          (relevant_answers += completed_answers).select(&:show?)
+        end
+
+        relevant_answers
       end
-      # rubocop:enable Metrics/MethodLength
+      # rubocop:enable Metrics/MethodLength, Metrics/AbcSize
 
       private
 
--- a/app/presenters/summary/sections/overview.rb
+++ b/app/presenters/summary/sections/overview.rb
@@ -13,16 +13,18 @@ module Summary
               :reference, crime_application.reference.to_s,
             ),
 
-            Components::ValueAnswer.new(
-              :means_tested, crime_application.is_means_tested,
-              change_path: edit_steps_client_is_means_tested_path
-            ),
-
             Components::DateAnswer.new(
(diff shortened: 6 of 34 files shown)


app/forms/steps/client/has_nino_form.rb
```diff
@@ -7,14 +7,22 @@ class HasNinoForm < Steps::BaseFormObject
...
 
       def nino=(str)
         super(str.upcase.delete(' ')) if str
       end
 
       private
 
+      def validate_nino
+        if nino.blank?
+          crime_application.not_means_tested? ? return : errors.add(:nino, :blank)
```
arthurashman, line 20: NINO remains mandatory for means tested cases as we still need it to run Benefit Checker, but it can be optional if the application is not means tested

app/models/concerns/passportable.rb
```diff
@@ -2,7 +2,8 @@ module Passportable
   extend ActiveSupport::Concern
 
   delegate :age_passported?,
-           :benefit_check_passported?, to: :means_passporter
+           :benefit_check_passported?,
+           :not_means_tested?, to: :means_passporter
```
arthurashman, line 6: not_means_tested was easiest to think of in terms of passporting, and is actually very similar to u18

app/presenters/summary/components/base_answer.rb
```diff
@@ -16,7 +16,7 @@ def initialize(question, value, *args)
       end
 
       def show?
-        show.presence || value?
```
arthurashman, line 19: `presence` doesn't work if `show` is [false](url), as `false.present? `resolves to `false`, so `show?` didn't get set as `false`, but as `value?`

app/presenters/summary/sections/case_details.rb
```diff
@@ -10,6 +10,7 @@ def answers
         [
           Components::ValueAnswer.new(
             :case_type, kase.case_type,
+            show: crime_application.means_passport.exclude?('on_not_means_tested'),
```
arthurashman, line 13: Provider won't have been asked case_type if the application is not means tested, so we don't want to show this row.

app/presenters/summary/sections/client_details.rb
```diff
@@ -30,7 +30,7 @@ def answers
 
           Components::FreeTextAnswer.new(
             :nino, applicant.nino,
-            change_path: edit_steps_client_has_nino_path
+            change_path: edit_steps_client_has_nino_path, show: true
```
arthurashman, line 33: As NINO is now sometimes optional, we want to always show this row, and populate it with None if nothing was entered

app/presenters/summary/sections/overview.rb
```diff
@@ -2,36 +2,50 @@ module Summary
   module Sections
     class Overview < Sections::BaseSection
       def show?
-        !crime_application.in_progress? && super
+        crime_application.case.present? && super
```
arthurashman, line 5: The Overview section was only shown on completed applications, however this section was the place it made most sense to show the user what they selected for 'Is this application means tested?'

It made sense to me to only show it if there was a case, because by that point the user will have got the answer to this, and a datestamp (which I have also decided there is no reason not to show on an in progress application as it is already shown on the tasklist sidebar)

app/presenters/summary/sections/overview.rb
```diff
@@ -2,36 +2,50 @@ module Summary
...
+              change_path: edit_steps_client_is_means_tested_path
+            ),
+
+            Components::DateAnswer.new(
+              :date_stamp, crime_application.date_stamp,
+              i18n_opts: { format: :datetime },
+            ),
+          ]
+
+        unless crime_application.in_progress?
```
arthurashman, line 27: These 3 answers are only available on completed applications

app/services/adapters/structs/income_details.rb
```diff
@@ -6,6 +6,7 @@ def employment_status
         employment_type || []
       end
 
+      # `client_has_dependants` is not part of Schema, requires calculation
```
arthurashman, line 9: Just moving this comment to where the code got moved to

app/services/datastore/application_rehydration.rb
```diff
@@ -36,11 +37,22 @@ def split_case?
       parent.return_details.reason.inquiry.split_case?
     end
 
+    # `is_means_tested` is not part of Schema, requires calculation
+    def means_tested
+      parent.means_passport.include?('on_not_means_tested') ? YesNoAnswer::NO : YesNoAnswer::YES
+    end
+
+    def client_has_partner
+      return if not_means_tested?
```
arthurashman, line 46: client_has_partner will not have been asked, so will be `nil` . It's not in the datastore yet, so needs handling here

app/services/datastore/application_rehydration.rb
```diff
@@ -79,5 +90,11 @@ def outgoings
 
       Outgoings.new(parent.outgoings.serializable_hash)
     end
+
+    def not_means_tested?
+      return true if means_tested.value == :no
+
+      false
```
arthurashman, line 97: For historic applications `means_tested` will be `nil` but as they have all been passported applications, none would be 'not means tested' - so this needs to be set to false if means_tested is nil - it will be an always asked, mandatory question moving forward

app/services/decisions/client_decision_tree.rb
```diff
@@ -55,6 +55,16 @@ def after_has_partner
       end
     end
 
+    def after_client_details
+      if DateStamper.new(form_object.crime_application, 'no_case_type').call
+        edit(:date_stamp)
+      elsif form_object.crime_application.not_means_tested?
```
arthurashman, line 61: If it already has a datestamp but is not means tested, it is routed differently (if no datestamp yet, it would go to datestamp page and routing onward from there is the same as here)

app/presenters/summary/sections/case_details.rb
arthurashman: Do as a helper method

app/presenters/summary/sections/overview.rb
```diff
@@ -2,36 +2,52 @@ module Summary
...
+        relevant_answers =
+          [
+            Components::FreeTextAnswer.new(
+              :reference, crime_application.reference.to_s,
+            ),
+
+            Components::DateAnswer.new(
+              :date_stamp, crime_application.date_stamp,
+              i18n_opts: { format: :datetime },
+            ),
```
arthurashman, line 19: Delete bc not always present

app/presenters/summary/sections/overview.rb
```diff
@@ -2,36 +2,52 @@ module Summary
...
+          ]
+
+        if FeatureFlags.means_journey.enabled? && crime_application.is_means_tested
```
arthurashman, line 22: test is_means_tested

app/presenters/summary/sections/overview.rb
willmcb: Might be worth wrapping this in a method so that it is clear why it is here. Its not that obvious that the case being pressent means that we should show overview details - unless you are familar with the flow.

```ruby
def show_overview_details?
    crime_application.case.present?
end
```

app/services/adapters/structs/crime_application.rb
```diff
@@ -4,6 +4,12 @@ module Adapters
   module Structs
     class CrimeApplication < LaaCrimeSchemas::Structs::CrimeApplication
       include TypeOfApplication
+      # `is_means_tested` is not part of Schema, requires calculation
+      # rubocop:disable Naming/PredicateName
+      def is_means_tested
```
willmcb, line 9: what aboud just `means_tested?` I think it means the same as `is_means_tested?` and we could lose the rubocop opt out

willmcb, line 9: What about `means_tested?` I think it means the same as `is_means_tested` and we could lose the rubocop optout

app/services/datastore/application_rehydration.rb
willmcb: same here could be `means_tested?`

app/services/date_stamper.rb
```diff
@@ -22,8 +22,9 @@ def call
     # -- If case type is “date stampable” then we use the date stamp value
     # -- If case is non “date stampable” we use the submission date as the date
     #    stamp.
+    return false if @crime_app.date_stamp.present?
 
-    if CaseType.new(@case_type).date_stampable? && @crime_app.date_stamp.nil?
+    if @crime_app.not_means_tested? || CaseType.new(@case_type).date_stampable?
```
willmcb, line 27: How comes we don't need the nil check as well?

app/services/passporting/means_passporter.rb
```diff
@@ -34,6 +41,12 @@ def passport_types_collection
 
     private
 
+    def app_not_means_tested?
+      return true if crime_application.is_means_tested == 'no'
```
willmcb, line 45: To me `is_means_tested` sounds like it should be a predicate method, its implying a boolean outcome so should probably end in a `?` in line with the ruby convention. It would mean in situations like this we don't have to the comparision to 'no' as it would return `true`/`false`

app/views/steps/client/has_nino/edit.html.erb
```diff
@@ -6,9 +6,11 @@
     <%= govuk_error_summary(@form_object) %>
 
     <%= step_form @form_object do |f| %>
+      <% heading = current_crime_application.is_means_tested == 'no' ? '.optional_nino' : '.nino'%>
```
willmcb, line 9: making it a predicate would also elminate the comparison here

spec/presenters/summary/sections/overview_spec.rb
```diff
@@ -42,36 +42,75 @@
...
   describe '#answers' do
     let(:answers) { subject.answers }
 
-    it 'has the correct rows' do
-      expect(answers.count).to eq(5)
+    context 'when the application is completed' do
+      it 'has the correct rows' do
+        expect(answers.count).to eq(6)
+
+        answer = answers[0]
```
willmcb, line 57: We have that pattern elsewhere that utilises a method to eliminate some of the duplication

app/presenters/summary/sections/case_details.rb
arthurashman: Since the show method now works properly, this isn't needed - as `show` resolves to false if the value is `nil`

app/services/adapters/structs/crime_application.rb
arthurashman: Well the thing that rubocop annoyed me with here is that it isn't actually a Predicate, as it doesn't return a boolean. This is to populate the YesNoAnswer on rehydration

app/services/date_stamper.rb
arthurashman: Bc the nil check is covered on line 25

I wanted to save the cognitive overhead of brackets

arthurashman: It was:
If it's a case type that should get a datestamp, AND doesn't already have a date stamp, give it a datestamp

Now:
If it already has a datestamp return false
THEN
If it's a non means tested app OR  a case type that should get a datestamp, give it a datestamp

app/services/passporting/means_passporter.rb
arthurashman: That's just the name I gave the value on the form, it made logical choice given the question wording, but I appreciate it breaks convention (and brain) so could change it. What do you think? - I was trying to follow the 'has_partner' pattern

https://staging.apply-for-criminal-legal-aid.service.justice.gov.uk/applications/68ce4681-3712-4f0e-add7-571a4557a919/steps/client/is_application_means_tested

app/services/datastore/application_rehydration.rb
arthurashman: means_tested method is rebuilding the attribute called `is_means_tested` on the `crime_application` which is a YesNoAnswer

no_means_tested? is a predicate helper method used here to skip a local rehydration of partner that is not stored in the db and needs to be nil in this scenario

willmcb (approved):
👍

PR #604, opened 2024-01-17
Apply: Add Has the case concluded page

ripan:
## Description of change

Add "Has the case concluded?" page behind feature flag `case_concluded_page`

Add db columns
-   cases.has_case_concluded(string)
-   cases.date_case_concluded(date)

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-310

## Notes for reviewer
**Currently "Has the case concluded?" page is behind a feature flag for `local` and `staging`**

Screenshot 2024-01-19 at 15 36 10


app/presenters/summary/components/base_answer.rb
```diff
@@ -16,7 +16,7 @@ def initialize(question, value, *args)
       end
 
       def show?
-        show.presence || value?
+        show.nil? ? value? : show
```
willmcb, line 19: 👍

app/presenters/summary/sections/case_details.rb
```diff
@@ -44,6 +55,10 @@ def answers
       def kase
         @kase ||= crime_application.case
       end
+
+      def case_concluded?
+        kase.has_case_concluded == 'yes' && !kase.date_case_concluded.nil?
```
willmcb, line 60: If this is a ValueObject I think we can just use `has_case_concluded.yes?` rather than doing the direct comparison with the string.

config/settings.yml
```diff
@@ -27,6 +27,10 @@ feature_flags:
     local: true
     staging: true
     production: false
+  case_concluded_page:
```
willmcb, line 30: I think as this is part of the means journey this needs to be under the 'means_journey' flag rather than its own flag for the page 👍

willmcb (changes requested):
Code works well and you are awaiting the other pages so that the flow issue is fixed. It should be under the means_journey flag though not its own page flag

Theres also one tiny comment to address on using predicate methods for the ValueObject 👍

willmcb (approved):
Nice 👍

PR #621, opened 2024-01-29
Fix/add guard clause for case concluded remanded

arthurashman:
## Description of change
With means_journey FeatureFlag off in local, the CYA page was breaking as the app does not have the case concluded and remanded in custody routes defined. Put these behind guard clauses to make sure they don't get called whilst means_journey FeatureFlag is off, so that this change can be deployed to prod. 

To test:

1. You can see that in main, if you switch the means_journey feature flag to false for local (and restart your local server), that the CYA page blows up as the case concluded and client remanded routes aren't defined.
2. Checkout this branch
3. Complete a means tested application, Selecting Yes and providing dates for Case concluded and Client remanded pages
4. Ensure on the CYA page, those 4 rows show in the Case section, and the change links go to the correct pages
5. Ensure that the application submits
6. Switch the means_journey feature flag to false for local and restart your local server
7. Complete an application - you should not see the means tested q, or Case concluded and Client remanded pages
8. Ensure the CYA page renders, Ensure those 4 rows are not there
9. Ensure that the application submits

### DIFF
--- a/app/controllers/developer_tools/crime_applications_controller.rb
+++ b/app/controllers/developer_tools/crime_applications_controller.rb
@@ -29,10 +29,8 @@ module DeveloperTools
       find_or_create_applicant
 
       crime_application.update(
-        is_means_tested: YesNoAnswer::YES,
         client_has_partner: YesNoAnswer::NO,
         navigation_stack: [
-          edit_steps_client_is_means_tested_path(crime_application),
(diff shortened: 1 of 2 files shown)


timpeat (approved):
Looks good to me. To prevent this sort of issue in the future, I think we'd be better off feature flagging in the relevant controllers rather than the routes.

willmcb (approved):

PR #626, opened 2024-02-02
Fix/validation for missing case type

arthurashman:
## Description of change
In applications where provider initially sets non-means tested to true, then returns and changes to false, there will be no case type recorded. We need to add to the validator to catch this scenario.

### Screenshot after changes:
image

Create an application that is not means tested
When reaching the CYA page, click the change link for the means tested No answer - change it to Yes. 
Click save and continue
Select No for partner
Click save and continue
Click Confirm declaration and submit application
Click Save and submit to see errors

### DIFF
--- a/app/services/passporting/ioj_passporter.rb
+++ b/app/services/passporting/ioj_passporter.rb
@@ -50,7 +50,7 @@ module Passporting
     end
 
     def appeal_case_type?
-      return false if crime_application.not_means_tested?
+      return false if kase.case_type.nil?
 
       CaseType.new(kase.case_type).appeal?
     end
--- a/app/validators/application_fulfilment_validator.rb
+++ b/app/validators/application_fulfilment_validator.rb
@@ -12,7 +12,7 @@ class ApplicationFulfilmentValidator < BaseFulfilmentValidator
       ]
     end
 
-    if record.case.case_type.nil? && record.is_means_tested == 'yes'
+    if record.is_means_tested == 'yes' && record.case.case_type.nil?
       errors << [
         :base, :case_type_missing, { change_path: edit_steps_client_case_type_path }
       ]
--- a/app/validators/application_fulfilment_validator.rb
+++ b/app/validators/application_fulfilment_validator.rb
@@ -1,12 +1,20 @@
 class ApplicationFulfilmentValidator < BaseFulfilmentValidator
   private
 
-  def perform_validations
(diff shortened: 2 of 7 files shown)


config/locales/en/helpers.yml
```diff
@@ -28,7 +28,7 @@ en:
       steps_provider_select_office_form:
         office_code: Select your office account number
       steps_client_is_means_tested_form:
-        is_means_tested: Is this application means tested?
+        is_means_tested:  Is this application subject to the usual means or passported test?
```
mseedat-moj, line 31: Extra space introduced at start of line

config/locales/en/steps.yml
```diff
@@ -19,7 +19,7 @@ en:
     client:
       is_means_tested:
         edit:
-          page_title: Is this application means tested?
+          page_title:  Is this application subject to the usual means or passported test?
```
mseedat-moj, line 22: Extra space at start of line

app/validators/application_fulfilment_validator.rb
```diff
@@ -28,6 +28,12 @@ def perform_validations
       ]
     end
 
+    if record.case.case_type.nil? && record.is_means_tested == 'yes'
```
mseedat-moj, line 31: Recommend switching the conditions around:

`if record.is_means_tested == 'yes' && record.case.case_type.nil? ...` so that there is no need to hit the ActiveRecord cache or worse the database

mseedat-moj (changes requested):
Looks good, some minor changes required.

app/services/passporting/ioj_passporter.rb
```diff
@@ -50,7 +50,7 @@ def offences
     end
 
     def appeal_case_type?
-      return false if crime_application.not_means_tested?
+      return false if kase.case_type.nil?
```
mseedat-moj, line 53: Can this just be `kase.case_type` is the nil necessary?

mseedat-moj (approved):
Looks good. Minor suggestion not sure if my suggestion is actually correct tbh

app/services/passporting/ioj_passporter.rb
mseedat-moj: Ignore my comment didn't notice the method is a Boolean method

mseedat-moj (approved):

PR #633, opened 2024-02-09
[CRIMAPP-457] Prevent pse apps from being submitted without evidence

hiboabd:
## Description of change
Adds validation to the review/submit page to prevent users from submitting a pse application without uploading evidence 

## Link to relevant ticket
[CRIMAPP-457](https://dsdmoj.atlassian.net/browse/CRIMAPP-457)

## Notes for reviewer
I attempted to add a validation message to the upload screen itself but the error was shown when evidence was deleted meaning when an application only has one piece of evidence and that is deleted, the validation checking there is at least one piece of supporting evidence fails and then the error is shown. This is not simple to change so this approach was implemented instead as it is technically simpler

### After changes:
Screenshot 2024-02-09 at 15 32 43

1. Start a pse application
2. Click save and continue on the add supporting evidence screen 
3. Click whatever option on the adding more information and save and continue 
4. Review the application and save and continue 
5. Attempt to submit the application and you should see a validation error like the one in the screenshot 
6. Click the change link and ensure it takes you back to the evidence upload screen 

Play around with uploading and deleting files
Also ensure regular applications do not show this validation as users are not required to submit evidence (for now)

[CRIMAPP-457]: https://dsdmoj.atlassian.net/browse/CRIMAPP-457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/crime_application.rb
+++ b/app/models/crime_application.rb
@@ -37,4 +37,5 @@ class CrimeApplication < ApplicationRecord
   # Before submitting the application we run a final
   # validation to ensure some key details are fulfilled
   validates_with ApplicationFulfilmentValidator, on: :submission, unless: :post_submission_evidence?
+  validates_with PseFulfilmentValidator, on: :submission, if: :post_submission_evidence?
 end
--- a/app/validators/application_fulfilment_validator.rb
+++ b/app/validators/application_fulfilment_validator.rb
@@ -1,24 +1,6 @@
-class ApplicationFulfilmentValidator < ActiveModel::Validator
-  include Routing
-
-  attr_reader :record
-
-  def validate(record)
-    @record = record
-
-    errors = perform_validations
-    add_errors(errors)
-  end
-
-  # Used by the `Routing` module to build the urls
-  def default_url_options
-    { id: record }
-  end
-
+class ApplicationFulfilmentValidator < BaseFulfilmentValidator
   private
 
-  # More validations can be added here
-  # Errors, when more than one, will maintain the order
   def perform_validations
     errors = []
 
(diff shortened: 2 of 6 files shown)


app/models/crime_application.rb
```diff
@@ -37,4 +37,5 @@ class CrimeApplication < ApplicationRecord
   # Before submitting the application we run a final
   # validation to ensure some key details are fulfilled
   validates_with ApplicationFulfilmentValidator, on: :submission, unless: :post_submission_evidence?
+  validates_with PseFulfilmentValidator, on: :submission, if: :post_submission_evidence?
```
timpeat, line 40: 👍

app/validators/application_fulfilment_validator.rb
```diff
@@ -1,24 +1,6 @@
...
-  def default_url_options
-    { id: record }
-  end
-
+class ApplicationFulfilmentValidator < BaseFulfilmentValidator
   private
 
-  # More validations can be added here
-  # Errors, when more than one, will maintain the order
   def perform_validations
```
ripan, line 4: @hiboabd Just a minor suggestion. What do you think of changing method name from `perform_validations` to `validation_errors` as this method returns collection of errors ?

hiboabd: Thanks, this got merged so too late to do now 😄 it's doing both so both names are accurate. I'm happy to leave as is for now

PR #637, opened 2024-02-13
CRIMAPP-489 - Add validations

ripan:
## Description of change

State of case related forms should be maintained with valid data
- Conditional Validation added while submitting a form using "Save and come back later" button

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-489

Screenshot 2024-02-13 at 15 53 27
Screenshot 2024-02-13 at 16 46 39

### Before changes:
Case DB state after clicking "Save and come back later" when "Yes" is selected
-  is_preorder_work_claimed: "yes",
-  preorder_work_date: nil,
-  preorder_work_details: nil>

Review the application page
Do you intend to claim pre-order work? yes 

### After changes:
Case DB state after clicking "Save and come back later" when "Yes" is selected
-  is_preorder_work_claimed: "no",
-  preorder_work_date: nil,
-  preorder_work_details: nil>

Review the application page
Do you intend to claim pre-order work? no

Selecting 'Yes' for any or all of the following questions should save the associated dates/details. If any of the date/details is missing the answer should be saved as 'No'.

- Has the case concluded?
- Do you intend to claim pre-order work?
- Has a court remanded your client in custody?

### DIFF
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -23,15 +23,15 @@ class Case < ApplicationRecord
   private
 
   def case_concluded?
-    has_case_concluded == Constants::YES
+    has_case_concluded == YesNoAnswer::YES.to_s
   end
 
   def client_remanded?
-    is_client_remanded == Constants::YES
+    is_client_remanded == YesNoAnswer::YES.to_s
   end
 
   def preorder_work_claimed?
-    is_preorder_work_claimed == Constants::YES
+    is_preorder_work_claimed == YesNoAnswer::YES.to_s
   end
 
   def initialise_dates(charge)
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -12,6 +12,10 @@ class Case < ApplicationRecord
     end
   end
 
+  validates :date_case_concluded, presence: true, if: -> { has_case_concluded == 'yes' }
+  validates :date_client_remanded, presence: true, if: -> { is_client_remanded == 'yes' }
+  validates :preorder_work_date, :preorder_work_details, presence: true, if: -> { is_preorder_work_claimed == 'yes' }
+
   composed_of :hearing_court, class_name: 'Court',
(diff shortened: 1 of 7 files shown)


app/presenters/tasks/case_details.rb
```diff
@@ -20,7 +20,7 @@ def in_progress?
     def completed?
       crime_application.case.values_at(
         :hearing_court_name, :hearing_date
-      ).all?(&:present?)
+      ).all?(&:present?) && crime_application.case.valid?
```
mseedat-moj, line 23: Could the new validation `crime_application.case.valid` be added as a guard clause to make it easier to follow the rule?

app/models/case.rb
```diff
@@ -12,6 +12,10 @@ def complete
     end
   end
 
+  validates :date_case_concluded, presence: true, if: -> { has_case_concluded == 'yes' }
+  validates :date_client_remanded, presence: true, if: -> { is_client_remanded == 'yes' }
+  validates :preorder_work_date, :preorder_work_details, presence: true, if: -> { is_preorder_work_claimed == 'yes' }
+
```
mseedat-moj, line 18: Not sure what best practice is - whether to use the constant for the yes value or ok to just compare to 'yes'?

mseedat-moj (commented):
Some suggestions

app/models/case.rb
ripan: We need to declare global constants somewhere. So i added new file. Thanks for pointing it out. 👏

app/presenters/tasks/case_details.rb
ripan: @mseedat-moj I removed this additional validation for time being to support old applications.  For new applications Case model is validating the  object before saving. 
I think definition of completed? needs to be revised

PR #640, opened 2024-02-16
[CRIMAPP-281] Limit dependants to 50

mseedat-moj:
## Description of change
Limits number of dependants a user can add to 50. Arbitrary number, seems sensible.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-281

## Notes for reviewer
- Added the 'fifty first' translation string because of a quirk in the way this 50 limit works:
- When the user presses 'Add new dependant' that creates a new row in the `dependants` database table, but wit a `NULL` age value.
- If the user then presses 'Add new dependant' to save the age for entry number 50 (which updates the database correctly), yet another row is generated in the database for entry number 51, with a `NULL` age, however a validation error is also shown
- Therefore the simplest solution is to allow Dependant entry number 51 to show on screen, show the error message and hope the user does not ever have 50 dependants.

Screenshot 2024-02-16 at 10 04 58

Create an application and then navigate to the dependants form e.g.:

https://https://staging.review-criminal-legal-aid.service.justice.gov.uk/applications/<CRIME APP ID>/steps/income/dependants

### DIFF
--- a/config/locales/en/errors.yml
+++ b/config/locales/en/errors.yml
@@ -377,7 +377,7 @@ en:
         steps/income/dependants_form:
           attributes:
             dependants:
-              too_long: Must have fewer than 50 dependants
+              too_long: Must have 50 or fewer dependants
         steps/income/dependant_fieldset_form:
           summary:
             age:
--- a/config/locales/en/errors.yml
+++ b/config/locales/en/errors.yml
@@ -374,6 +374,10 @@ en:
           attributes:
             client_has_dependants:
               inclusion: Select yes if your client has any dependants
+        steps/income/dependants_form:
+          attributes:
+            dependants:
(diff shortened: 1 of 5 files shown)


config/locales/en/errors.yml
```diff
@@ -374,6 +374,10 @@ en:
           attributes:
             client_has_dependants:
               inclusion: Select yes if your client has any dependants
+        steps/income/dependants_form:
+          attributes:
+            dependants:
+              too_long: Must have fewer than 50 dependants
```
arthurashman, line 380: Does this need to be 51? Not sure, just checking! - i.e. they can actually add 50, but not 51 right?

arthurashman (approved):
Just the one question re 50/51 in the error message

PR #641, opened 2024-02-16
Connect income journey to flow

arthurashman:
## Description of change

Connect income journey to follow on from benefit type question

## Notes for reviewer
Following on from a suggestion from Justin, looking at the code and the flow, this makes complete sense to have the employment and means questions come directly after the user has said they do not get a passporting benefit. 

This feels right logically from a user journey perspective, but also makes the conditionality of entering that flow really easy based on the answer to the passporting benefit type question. 

N.B. This messes up the order as appears on the tasklist and would need attention there (if we keep tasklist!)

1. Set all the feature flags to production mode, and do a regression where you submit and application for an applicant passported on means, send it back, re-submit, and Mark it as ready for MAAT
2. With flags in production mode, answer 'None' to passporting benefit question and check that you get the eforms exit page
3. Turn on the means assessment features, ensure that answering 'None' to passporting benefit, and 'Not employed' to employment status enters the means journey.
4. Submit an application with means info, send it back, re-submit and then mark it ready for MAAT


mseedat-moj (approved):
👍

willmcb (approved):
👍

PR #650, opened 2024-02-19
CRIMAPP-536 board and lodgings payments

RoseSAK:


app/models/outgoing_payment.rb
```diff
@@ -0,0 +1,5 @@
+class OutgoingPayment < ApplicationRecord
```
arthurashman, line 1: Can we rename all of these OutgoingsPayment?

app/models/crime_application.rb
```diff
@@ -15,6 +15,8 @@ class CrimeApplication < ApplicationRecord
   has_many :documents, dependent: :destroy
   has_many :income_payments, dependent: :destroy
   accepts_nested_attributes_for :income_payments, allow_destroy: true
+  has_many :outgoing_payments, dependent: :destroy
```
arthurashman, line 18: Can we rename all of these outgoings_payment?

app/value_objects/outgoing_payment_type.rb
```diff
@@ -0,0 +1,11 @@
+class OutgoingPaymentType < ValueObject
```
arthurashman, line 1: Can we rename all of these OutgoingsPayment?

db/migrate/20240208114218_create_outgoing_payments.rb
```diff
@@ -0,0 +1,15 @@
+class CreateOutgoingPayments < ActiveRecord::Migration[7.0]
```
arthurashman, line 1: OutgoingsPayments

app/controllers/steps/outgoings/board_and_lodgings_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+        )
+      end
+
+      def update
+        update_and_advance(BoardAndLodgingsForm, as: :board_and_lodgings)
+      end
+
+      def additional_permitted_params
+        []
+      end
```
mseedat-moj, line 16: If this method is returning an empty array the method can be deleted

app/forms/steps/outgoings/board_and_lodgings_form.rb
```diff
@@ -0,0 +1,53 @@
...
+          form.send("#{attr_name}=", value)
+        end
+        form
+      end
+
+      private
+
+      def amount
+        amount = board_amount.to_f - food_amount.to_f
+        amount.to_s
```
mseedat-moj, line 41: Is this right - the amount is being returned (and saved?) as a string?

app/forms/steps/outgoings/board_and_lodgings_form.rb
```diff
@@ -0,0 +1,53 @@
...
+      end
+
+      def persist!
+        outgoings_payment.update!(
+          amount: amount,
+          frequency: frequency,
+          details: attributes.except('frequency')
```
mseedat-moj, line 48: Would prefer the attributes to be saved to be explicitly listed (allowlist) rather than a denylist.

mseedat-moj (commented):
Some minor comments whilst it's WIP

app/forms/steps/outgoings/board_and_lodgings_form.rb
RoseSAK: The board_amount and food_amount when cast to Pence are strings (not sure why) so in this method that's why I'm converting to floats and back to a string to be consistent - maybe the conversion back to string is not needed though

app/forms/steps/outgoings/board_and_lodgings_form.rb
RoseSAK: Okay good point, will update this!

app/forms/steps/outgoings/board_and_lodgings_form.rb
RoseSAK: Update: `to_s` is not needed so have removed!

app/forms/steps/outgoings/board_and_lodgings_form.rb
```diff
@@ -0,0 +1,52 @@
...
+      validates :food_amount, presence: true
+      validates :frequency, inclusion: { in: PaymentFrequencyType.values }
+      validates :payee_name, presence: true
+      validates :payee_relationship_to_client, presence: true
+
+      def outgoings_payment
+        @outgoings_payment ||= crime_application.outgoings_payments.find_by(
+          payment_type: OutgoingsPaymentType::BOARD_AND_LODGING.to_s
+        )
+      end
```
mseedat-moj, line 20: Is this method definitely needed? I think it is only used by the `persist` method now?

app/services/decisions/outgoings_decision_tree.rb
```diff
@@ -17,5 +19,15 @@ def destination # rubocop:disable Metrics/MethodLength
...
       end
     end
+
+    private
+
+    def after_housing_payment_type
+      if form_object.housing_payment_type.value == :board_and_lodging
+        edit(:board_and_lodgings)
+      else
+        edit(:council_tax)
```
mseedat-moj, line 29: The `board_and_lodging` and `board_and_lodgings` is causing a lot of confusion - can the route name alias `board_and_lodgings` be the same as the type value `board_and_lodging` or is that too difficult to alter?

db/schema.rb
```diff
@@ -222,9 +222,9 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.string "payment_type", null: false
-    t.integer "amount", null: false
-    t.string "frequency", null: false
-    t.jsonb "metadata", default: {}, null: false
+    t.integer "amount"
+    t.string "frequency"
+    t.jsonb "metadata", default: {}
```
mseedat-moj, line 227: This schema change to remove the `not null` property will need a dedicated migration because the existing staging/production database already enforces null: false. Otherwise your user journey will fail on staging/production

mseedat-moj (changes requested):

mseedat-moj (approved):
Top stuff

PR #652, opened 2024-02-19
Remove more information feature flag

hiboabd:
## Description of change
Remove more information feature flag

### DIFF
--- a/app/presenters/summary/sections/more_information.rb
+++ b/app/presenters/summary/sections/more_information.rb
@@ -1,10 +1,6 @@
 module Summary
   module Sections
     class MoreInformation < Sections::BaseSection
-      def show?
-        crime_application.case.present? && super
-      end
-
       def answers
         [
           Components::FreeTextAnswer.new(
--- a/app/presenters/summary/sections/more_information.rb
+++ b/app/presenters/summary/sections/more_information.rb
@@ -2,7 +2,7 @@ module Summary
   module Sections
     class MoreInformation < Sections::BaseSection
       def show?
-        FeatureFlags.more_information.enabled?
+        crime_application.case.present? && super
       end
 
       def answers
--- a/app/presenters/tasks/more_information.rb
+++ b/app/presenters/tasks/more_information.rb
@@ -5,7 +5,7 @@ module Tasks
     end
 
     def not_applicable?
-      !FeatureFlags.more_information.enabled?
+      false
     end
 
     def can_start?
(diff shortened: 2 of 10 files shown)


app/presenters/summary/sections/more_information.rb
```diff
@@ -1,10 +1,6 @@
 module Summary
   module Sections
     class MoreInformation < Sections::BaseSection
-      def show?
-        FeatureFlags.more_information.enabled?
-      end
```
hiboabd, line 6: Removed as I believe this will always display? Not sure however so correct me if wrong!

app/presenters/tasks/more_information.rb
```diff
@@ -5,7 +5,7 @@ def path
     end
 
     def not_applicable?
-      !FeatureFlags.more_information.enabled?
+      false
     end
```
mseedat-moj, line 9: Can this method be deleted entirely if it is always false?

mseedat-moj (changes requested):

app/presenters/tasks/more_information.rb
arthurashman: not_applicable detaults to true in BaseTask, so needs to be implemented and overridden here

app/presenters/summary/sections/more_information.rb
arthurashman: Yep. Section always shown, and content defaulted to 'None' if nothing entered

app/services/decisions/case_decision_tree.rb
```diff
@@ -128,8 +128,6 @@ def blank_date_required?
     end
 
     def submission_root_step
```
arthurashman, line 130: No longer need this helper method. Can just put `edit('/steps/submission/more_information')` back into the `def after_ioj` helper

arthurashman, line 130: Likewise here, we can put `edit('/steps/submission/more_information')` under `when :upload_finished` directly in the decision tree

app/services/decisions/evidence_decision_tree.rb
```diff
@@ -14,11 +14,7 @@ def destination
     private
 
     def submission_root_step
```
arthurashman, line 16: Likewise here, we can put `edit('/steps/submission/more_information')` under `when :upload_finished` directly in the decision tree

arthurashman (approved):
Just some non-blocking tidying of helper methods in the decision trees. 

Otherwise looks good. Thank you! 👍

app/services/decisions/evidence_decision_tree.rb
hiboabd: 🤦‍♀️ corrected!

willmcb (approved):
👍

arthurashman (approved):

mseedat-moj (approved):
Nice work

PR #663, opened 2024-02-23
Pence type can handle nil and negative values

timpeat:
## Description of change
Prior to this change, `nil` amounts were being serialised as `0`. This change fixes that.
It also specs Pence to make sure that it can handle negative numbers.

### Before changes:
Screenshot 2024-02-23 at 15 29 19

### After changes:
Screenshot 2024-02-23 at 15 29 36

Load a form with a pence attribute. Do not fill it in and save the page. Confirm that the pence attribute is not set to 0.00.

### DIFF
--- a/app/attributes/type/pence.rb
+++ b/app/attributes/type/pence.rb
@@ -10,11 +10,13 @@ module Type
     def deserialize(value)
       return if value.blank?
 
-      super.to_f / 100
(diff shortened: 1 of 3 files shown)


arthurashman (approved):
Looks great, works great. 

Thanks for sorting. Nice to have the input fields blank!

PR #676, opened 2024-02-28
[CRIMAPP-555] Add which other savings page

hiboabd:
## Description of change
Adds ‘Which other savings does your client have inside or outside the UK?’ screen, adding validation and linking to current savings flow

## Link to relevant ticket
[CRIMAPP-555](https://dsdmoj.atlassian.net/browse/CRIMAPP-555)

### After changes:
Screenshot 2024-02-29 at 09 03 19

Screenshot 2024-02-29 at 09 29 23

Pull down this branch
Start an application 
Go to `/steps/capital/which_savings_does_client_have`
Follow the flow to add a savings account
When on the savings summary page, select yes to add another saving
This should route you to the ‘Which other savings does your client have inside or outside the UK?’ screen
Click save and continue and check the validation error is in place 
Select a saving account option and select save and continue and confirm the validation error is not shown and that you are taken to the page to fill in the saving account details

[CRIMAPP-555]: https://dsdmoj.atlassian.net/browse/CRIMAPP-555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/views/steps/shared/_saving_type_form.html.erb
```diff
@@ -0,0 +1,30 @@
+<% if show_none_option %>
```
timpeat, line 1: I think it would be better to duplicate the form in the each edit view, rather than create a partial with lots of conditionals.

hiboabd: Thanks tim i've updated

app/forms/steps/capital/other_saving_type_form.rb
```diff
@@ -0,0 +1,11 @@
+module Steps
+  module Capital
+    class OtherSavingTypeForm < SavingTypeForm
+      private
+
+      def persist!
+        @saving = crime_application.savings.create(saving_type:)
+      end
```
hiboabd, line 8: Note this differs to the savings type form behaviour. On the other saving type form, we will always create a new saving rather than directing a user to an existing incomplete saving for the saving type selected.

timpeat (approved):
🎬

PR #679, opened 2024-02-28
Fix back button routing

arthurashman:
## Description of change
Fix back button routing
The path argument is optional for the step_header function, and only needs passing if you do not want the user to go back to the previous page they were on as recorded in the navigation stack.
Passing crime_applications_path means that if the user hits the back button they are taken to the tasklist.


hiboabd (approved):
Good to know!

PR #680, opened 2024-02-28
CRIMAPP-571 Allow means application through validator

arthurashman:
## Description of change
Allow means application through validator 

Now that we have a route through the application that does not run a DWP check, the end submission validator needs to allow through applications that do not have a means passport - this PR adds a light touch to check that the means journey was started 

## Link to relevant ticket
We will in time revisit the way this validator works and possibly the need for it at all, once a variety of different routes through would result in a valid application. 

Create an application, clicking 'My client receives none of these passporting benefits', then go through the means journey, and case details steps. ensure that the application is allowed to be submitted.

### DIFF
--- a/app/validators/application_fulfilment_validator.rb
+++ b/app/validators/application_fulfilment_validator.rb
@@ -32,6 +32,6 @@ class ApplicationFulfilmentValidator < BaseFulfilmentValidator
   end
 
   def means_record_present?
-    record.income.present? && record.income.employment_status.include?('not_working')
+    record.income.present? && record.income&.employment_status&.include?('not_working')
   end
 end
--- a/app/validators/application_fulfilment_validator.rb
+++ b/app/validators/application_fulfilment_validator.rb
@@ -3,10 +3,10 @@ class ApplicationFulfilmentValidator < BaseFulfilmentValidator
 
   # More validations can be added here
   # Errors, when more than one, will maintain the order
-  def perform_validations # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+  def perform_validations # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metrics/CyclomaticComplexity
     errors = []
 
-    unless Passporting::MeansPassporter.new(record).call || evidence_present?
+    unless Passporting::MeansPassporter.new(record).call || evidence_present? || means_record_present?
       errors << [
         :means_passport, :blank, { change_path: edit_steps_client_details_path }
       ]
@@ -29,5 +29,9 @@ class ApplicationFulfilmentValidator < BaseFulfilmentValidator
 
   def ioj_present?
     record.ioj.present? && record.ioj.types.any?
+  end
+
+  def means_record_present?
+    record.income.present? && record.income.employment_status.include?('not_working')
   end
 end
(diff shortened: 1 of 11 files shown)


app/validators/application_fulfilment_validator.rb
```diff
@@ -3,10 +3,10 @@ class ApplicationFulfilmentValidator < BaseFulfilmentValidator
 
   # More validations can be added here
   # Errors, when more than one, will maintain the order
-  def perform_validations # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+  def perform_validations # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metrics/CyclomaticComplexity
     errors = []
 
-    unless Passporting::MeansPassporter.new(record).call || evidence_present?
+    unless Passporting::MeansPassporter.new(record).call || evidence_present? || means_record_present?
```
mseedat-moj, line 9: In `Passporting::MeansPassporter.new()...` the `.call` method has the following code:

```      means_passport = []
      means_passport << MeansPassportType::ON_NOT_MEANS_TESTED if app_not_means_tested?
      means_passport << MeansPassportType::ON_AGE_UNDER18      if applicant_under18?
      means_passport << MeansPassportType::ON_BENEFIT_CHECK    if benefit_check_passed?

      crime_application.update(means_passport:)
```

I think this is saving the means passporting decision making check? Does this mean skipping means passporting using `means_record_present?` will mean the reason is not persisted to database? Also should `means_record_present?` and `evidence_present?` be first in the if statement to prevent unnecessary object initialisation? Or does the order matter here through side-effects in the other 2 conditions?

app/validators/application_fulfilment_validator.rb
```diff
@@ -30,4 +30,8 @@ def perform_validations # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
   def ioj_present?
     record.ioj.present? && record.ioj.types.any?
   end
+
+  def means_record_present?
+    record.income.present? && record.income.employment_status.include?('not_working')
```
mseedat-moj, line 35: Does `income.employment_status` need a `nil` check as well? The schema at the moment doesn't enforce its presence but it is validated for correctness in `employment_status_form.rb`

mseedat-moj (commented):
Seems fine - some comments to ensure valid data check

app/validators/application_fulfilment_validator.rb
arthurashman: I can do to be double safe. But it is the first step in income, so if record.income.present? is true, the user has landed on the employment_status page and it's an empty array. I get your point though - I'll put it in in case the first page changes/the user can somehow skip p1 of income

app/validators/application_fulfilment_validator.rb
arthurashman: So the concept is that on validation we rerun the passporter in order to set the passport status based on the values the application has at the point of submission. 
i.e. applicant goes through the form with details that gains them a means passport, then return and change one of those values (e.g. NINO) and dip out and back to the submission page. When they try to save and submit, the passporter is rerun against the new NINO and returns 'No' - so the original means_passport gets scrapped, and the validator checks if there is otherwise evidence (currently only available for evidencing ppt benefit) then if no evidence, check that the application has means details (therefore not needing a ppt benefit)

I'm a little unsure what you mean by 'skipping' means passporting. If the means passporter was never run, then there would be no means passport. It being run again now would resolve to the same state. 

TBH, I've not put a huge amount of thought into this (order etc) because it needs revisiting soon anyway. The whole concept will quickly be nonsensical

mseedat-moj (approved):
Discussed with @arthurashman - may require additional refactoring going forward.

PR #742, opened 2024-04-04
Fix app flow when passporting benefit is not selected

hiboabd:
## Description of change
Discovered during qa testing, when no passporting benefit is selected then applicant is not being routed to means assessment and they are not being correctly routed to the ioj screens 

Makes the following fixes: 
- Reverts to routing to the ioj screens following the court hearing details 
- Routes to means assessment following ioj screens if the applicant does not have a passporting benefit or does not have benefit evidence 
- Following an undetermined dwp check, routes to case details if instead of the benefit exit screen if user does not have benefit evidence

## Link to relevant ticket
[CRIMAPP-693](https://dsdmoj.atlassian.net/browse/CRIMAPP-693)

Test the scenarios above and in the ticket 

[CRIMAPP-693]: https://dsdmoj.atlassian.net/browse/CRIMAPP-693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/decisions/client_decision_tree.rb
+++ b/app/services/decisions/client_decision_tree.rb
@@ -25,10 +25,8 @@ module Decisions
         after_benefit_type
       when :retry_benefit_check
         determine_dwp_result_page
-      when :benefit_check_result
-        after_dwp_check
-      when :has_benefit_evidence
-        after_has_benefit_evidence
+      when :benefit_check_result, :has_benefit_evidence
+        edit('/steps/case/urn')
       when :cannot_check_benefit_status
         after_cannot_check_benefit_status
       else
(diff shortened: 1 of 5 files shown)


app/services/decisions/client_decision_tree.rb
```diff
@@ -133,18 +131,6 @@ def determine_dwp_result_page
...
 
-    def after_dwp_check
-      edit('/steps/case/urn')
-    end
-
-    def after_has_benefit_evidence
-      if form_object.has_benefit_evidence.yes?
-        edit('/steps/case/urn')
-      else
-        show(:evidence_exit)
```
arthurashman, line 144: We're not going to forget to update the copy on the 'No' option on this page right? There is that other ticket for it

arthurashman (approved):
Thanks @hiboabd !

app/services/decisions/client_decision_tree.rb
hiboabd: Yes there is but i just added it quickly

Screenshot 2024-04-04 at 12 26 20

PR #750, opened 2024-04-08
[CRIMAPP-662] Add cannot check with dwp screen

hiboabd:
## Description of change
Add new cannot check dwp status screen 
Removes and replaces previous retry benefit check screen (see screenshot below)

## Link to relevant ticket
[CRIMAPP-662](https://dsdmoj.atlassian.net/browse/CRIMAPP-662)

### Before changes:
Screenshot 2024-04-08 at 15 19 07

### After changes:
Screenshot 2024-04-08 at 16 11 22

Note you will need to set the env variable `BC_USE_DEV_MOCK` to `false` to simulate the dwp api being down 
Test the following scenario with the feature flag turned on and then turned off  

1. Start an application 
2. Enter a nino
3. Select a benefit type 
4. You should see the screen `Sorry we cannot check your client's info with DWP`
5. Click save and continue and you should be routed to the screen `Do you have evidence your client gets ____?`
6. Go back and click save and come back later and you should be routed to the task list 

[CRIMAPP-662]: https://dsdmoj.atlassian.net/browse/CRIMAPP-662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/decisions/client_decision_tree.rb
```diff
@@ -133,7 +131,7 @@ def determine_dwp_result_page
       DWP::UpdateBenefitCheckResultService.call(applicant)
 
       if applicant.passporting_benefit.nil?
-        edit(:retry_benefit_check)
+        edit('steps/dwp/cannot_check_dwp_status')
```
ripan, line 134: @hiboabd What do you think of using  **:cannot_check_dwp_status** instead of string path

app/services/decisions/dwp_decision_tree.rb
```diff
@@ -6,13 +6,19 @@ def destination
...
+        after_cannot_check_dwp_status
       else
         raise InvalidStep, "Invalid step '#{step_name}'"
       end
     end
 
     private
 
+    def after_cannot_check_dwp_status
+      edit('steps/client/has_benefit_evidence')
```
ripan, line 19: @hiboabd What do you think of using  **:cannot_check_dwp_status** instead of string path

ripan (approved):

app/services/decisions/client_decision_tree.rb
hiboabd: `cannot_check_dwp_status` is in the dwp scope so i can't do that here 😩

app/services/decisions/dwp_decision_tree.rb
hiboabd: The opposite problem for this one `has_benefit_evidence` is in the client scope 😩

PR #772, opened 2024-04-18
Fix bug in client decision tree 

RoseSAK:
## Description of change

Add guard clause to `entered_appeal_reference_number?` to prevent  throwing an error when case type is nil

## Link to relevant ticket

[Sentry alert](https://ministryofjustice.sentry.io/issues/5216134617/?alert_rule_id=14332616&alert_type=issue&notification_uuid=dc0ac186-f05b-4a47-9dc9-dfcd66094590&project=6587263&referrer=slack)

### DIFF
--- a/app/services/decisions/client_decision_tree.rb
+++ b/app/services/decisions/client_decision_tree.rb
@@ -183,6 +183,8 @@ module Decisions
 
     def entered_appeal_reference_number?
       kase = current_crime_application.case
+      return false if kase&.case_type.nil?
+
       case_type = CaseType.new(kase.case_type)
       case_type&.appeal? && kase.appeal_reference_number.present?
     end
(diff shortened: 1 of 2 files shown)


app/services/decisions/client_decision_tree.rb
```diff
@@ -183,6 +183,8 @@ def after_cannot_check_dwp_status
 
     def entered_appeal_reference_number?
       kase = current_crime_application.case
+      return false if kase&.case_type.nil?
```
arthurashman, line 186: I'm not convinced we need the null safety check. Have you seen that we do, or just being cautious?

RoseSAK: Just being super cautious - I wrongly assumed there was no scenario in which case_type wouldn't be there, so don't want to make the same assumptions about case!

hiboabd (approved):

PR #773, opened 2024-04-18
[CRIMAPP-649] Introduce evidence rules definition framework

mseedat-moj:
## Description of change
Allow individual rules to be defined and then executed to show output (a prompt)

## Link to relevant ticket

## Notes for reviewer
This is the initial framework introducing various concepts such as a `Rule` a `Runner` a `Prompt` and the `Rule Generator`. Only 1 rule is implemented in this PR - `NationalInsuranceProof`

Create a new rule using the generator:
```
rails g evidence:rule --key reference_to_spreadsheet_entry --class SomeNewRule
```

This generator creates a corresponding spec.

The general system design is:

> The user is presented with the results of a `Prompt`. A `Prompt` invokes the ruleset `Runner` which generates a `Ruleset`. A `Ruleset` is a collection of `Rule` definitions. A `Rule` definition file will define 1, 2 or 3 predicates - one for each `persona`. Personas are `:client`, `:partner` and `:other`. Each `Rule` is also associated withe a `group`. A group can be `:income`, `:outgoings`, `:capital` or `:none`. Each predicate will evaluate to true or false. The combination of `group`, `persona` and corresponding `predicate` generates `sentences` to display to the user, and persist to the database.

#### What is a key?
Each `Rule` is associated with a `key`. Each `key` can have 1 or more Rule definitions - usually there is just 1, but over time there may be new Rule versions created for the same key. The `key` is a reference to the original rule defined in the 'Evidence Triggers' spreadsheet, and allows an audit trail between implemented code and business analysis documentation.

#### With some fake rules
Screenshot 2024-04-18 at 05 41 37

#### With a real rule to check if National Insurance proof is required
Screenshot 2024-04-18 at 05 42 28

#### When persisted ruleset is executed using an archived rule after resubmission (note: fake rules)
Screenshot 2024-04-18 at 12 48 56

#### When no rules are triggered

Screenshot 2024-04-24 at 06 18 51

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,7 @@ gem 'clamby', '1.6.10', require: false
 # Exceptions notifications
 gem 'sentry-rails'
 gem 'sentry-ruby'
+gem 'stackprof'
 
 # The original asset pipeline for Rails [https://github.com/rails/sprockets-rails]
 gem 'sprockets-rails'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -435,6 +435,7 @@ GEM
       actionpack (>= 5.2)
       activesupport (>= 5.2)
       sprockets (>= 3.0.0)
+    stackprof (0.2.26)
     stringio (3.1.0)
     thor (1.3.1)
     timeout (0.4.1)
@@ -511,6 +512,7 @@ DEPENDENCIES
   sentry-ruby
   simplecov
   sprockets-rails
+  stackprof
   uk_postcode
   web-console
   webmock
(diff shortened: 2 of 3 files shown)


app/services/evidence/rules/20240418015755_national_insurance_proof.rb
```diff
@@ -0,0 +1,15 @@
+module Evidence
+  module Rules
+    class NationalInsuranceProof < Rule
+      include Evidence::RuleDsl
+
+      key :national_insurance_32
```
hiboabd, line 6: what's the significance of 32?

app/services/evidence/rule_dsl.rb
```diff
@@ -0,0 +1,72 @@
+module Evidence
+  module RuleDsl
```
hiboabd, line 2: Dsl?

timpeat (commented):
Looks Great @mseedat-moj !

timpeat (approved):

PR #779, opened 2024-04-18
CRIMAPP-747 appeal with no changes CYA bug

RoseSAK:
## Description of change

Appeal no changes case types skip over passporting steps, and most means questions (although bizarrely do ask employment type - I'd expect this is an error on eForms we can enquire about later) 

As such we want to skip the Income CYA (because if we skip income payments, income benefits, dependants we don't want to show the 'none' answers for those sections, and so it will only show the remaining answers for employment type and lost job and date)

On final CYA, we don't want to show income payments, income benefits, dependants 'none' answers because the user wasn;'t  asked them/

No passport, no means also needed modification to get through the CYA and submission validator

## Link to relevant ticket

[CRIMAPP-747](https://dsdmoj.atlassian.net/browse/CRIMAPP-747)

[CRIMAPP-747]: https://dsdmoj.atlassian.net/browse/CRIMAPP-747?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/presenters/summary/html_presenter.rb
+++ b/app/presenters/summary/html_presenter.rb
@@ -1,6 +1,5 @@
 module Summary
-  # rubocop:disable Metrics/ClassLength
-  class HtmlPresenter
+  class HtmlPresenter # rubocop:disable Metrics/ClassLength
     attr_reader :crime_application
 
     delegate :application_type, to: :crime_application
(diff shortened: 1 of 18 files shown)


app/presenters/summary/html_presenter.rb
```diff
@@ -43,6 +43,22 @@ class HtmlPresenter
         more_information
         legal_representative_details
       ],
+      appeal_with_no_changes: %i[
```
timpeat, line 46: 👍

app/presenters/summary/html_presenter.rb
```diff
@@ -72,27 +88,36 @@ def initialize(crime_application:)
...
+
+    private
+
+    def build_sections(relevant_sections)
+      SECTIONS.fetch(relevant_sections).map do |section|
         Sections.const_get(section.to_s.camelize).new(crime_application)
       end.select(&:show?)
     end
+
+    def appeal_no_changes?
```
timpeat, line 118: Could this live in "TypeOfApplication"

timpeat (approved):

app/presenters/summary/html_presenter.rb
arthurashman: That would be the ideal place for it to be, it just didn't conceptually feel like a different kind of application - does it fit with your model of being a different kind of application @timpeat 

And if so where does that stop (i.e. full vs initial means assessment for example)

PR #806, opened 2024-04-25
Give non-means tested its own feature flag

arthurashman:


spec/requests/dashboard_spec.rb
```diff
@@ -27,7 +27,7 @@
 
     context 'when means journey feature flag is not enabled' do
```
hiboabd, line 28: when non means journey feature flag is not enabled

app/services/passporting/means_passporter.rb
```diff
@@ -5,7 +5,11 @@ def call
       return true if appeal_no_changes?
 
       means_passport = []
-      means_passport << MeansPassportType::ON_NOT_MEANS_TESTED if app_not_means_tested?
+
+      if app_not_means_tested? && FeatureFlags.non_means_tested.enabled?
+        means_passport << MeansPassportType::ON_NOT_MEANS_TESTED
+      end
```
hiboabd, line 11: I think the feature flag could be a guard clause in the `app_not_means_tested` method, would be easier to follow

app/services/datastore/application_rehydration.rb
```diff
@@ -51,6 +51,8 @@ def split_case?
 
     # `is_means_tested` is not part of Schema, requires calculation
     def means_tested
+      return nil unless FeatureFlags.non_means_tested.enabled?
```
hiboabd, line 54: Why `nil` and not true?

hiboabd (approved):
LGTM just a couple questions/comments

PR #808, opened 2024-04-25
Fix/remove eforms exit for means journey

arthurashman:
## Description of change
If user fails the DWP check, then answer Yes when asked if the DWP records are correct, the app shows the eForms exit page (on the previous assumption that they couldn't continue the application as we didn't have means pages). If means pages are available, we can allow the user to continue


hiboabd (approved):
👍

PR #816, opened 2024-04-26
[CRIMAPP-649] Allow evidence file download

mseedat-moj:
## Description of change
Allows evidence file uploaded by the user to be downloaded by the user, including immediately after upload

## Link to relevant ticket

## Notes for reviewer
- To get the URL to show immediately after upload requires the document URL to be submitted to the frontend AJAX request, so introduced the virtual attribute `Document#url`
- Ensures the document download only occurs if the document belongs to the current document

Go to /steps/evidence/upload and add document

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -44,10 +44,8 @@ class DocumentsController < ApplicationController
   private
 
   def set_document
-    @document = Document.find(params[:document_id])
-
-    return if current_crime_application.documents.find { |document| document.id == @document.id }
-
+    @document = current_crime_application.documents.find(params[:document_id])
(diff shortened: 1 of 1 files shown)


app/controllers/documents_controller.rb
```diff
@@ -25,12 +28,29 @@ def create
...
+
+    redirect_to(presign_download.url, allow_other_host: true)
+  end
 
   def destroy; end
 
   private
 
+  def set_document
+    @document = Document.find(params[:document_id])
```
timpeat, line 47: Can we do the document find after the current_crime_application check (to prevent leaking information in request times).

timpeat (approved):

PR #819, opened 2024-04-29
[CRIMAPP-786] Add guard to check if income/outgoings present

hiboabd:
## Description of change
Fixes error on completed application page for applications that did not go through means assessment (e.g. nino = yes, passporting benefit = yes, has benefit evidence = yes). 

Adds a guard to prevent checking attributes on the income/outgoings object when they are not present


timpeat (approved):
🚀

PR #822, opened 2024-04-29
[CRIMAPP-649] Add restraint freezing order rule

hiboabd:
## Description of change
Adds rule to request evidence for having assets in restraint or freezing order

## Link to relevant ticket
### After changes:

Screenshot 2024-04-29 at 16 04 57


app/services/evidence/rules/20240429122128_restraint_or_freezing_order.rb
```diff
@@ -0,0 +1,20 @@
...
+  module Rules
+    class RestraintOrFreezingOrder < Rule
+      include Evidence::RuleDsl
+
+      key :restraint_freezing_order_31
+      group :none
+
+      client do |crime_application|
+        (crime_application.capital&.has_frozen_income_or_assets == 'yes' ||
+          crime_application.income&.has_frozen_income_or_assets == 'yes') || false
```
timpeat, line 11: What's the || false for?

timpeat: @hiboabd oh I see. They can both return nil. Forgive me.

timpeat (approved):

app/services/evidence/rules/20240429122128_restraint_or_freezing_order.rb
hiboabd: @mseedat-moj it does error if `nil` is returned but i don't think my condition will ever return `nil`. Happy to keep as a precaution however

PR #833, opened 2024-05-01
CRIMAPP-775 require next hearing date for complete

timpeat:
## Description of change

Case details are only shown as complete on the task list when complete.

## Link to relevant ticket

[CRIMAPP-775](https://dsdmoj.atlassian.net/browse/CRIMAPP-775)

## Notes for reviewer

Case details did not fully understand when it was complete. This PR uses the same pattern as applied to the means sections to ensure we understand when CaseDetails are complete.

See ticket for acceptance

[CRIMAPP-775]: https://dsdmoj.atlassian.net/browse/CRIMAPP-775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -16,54 +16,15 @@ class Case < ApplicationRecord
               mapping: %i[hearing_court_name name],
               constructor: :find_by_name, allow_nil: true
 
-  validates_with CaseDetails::AnswersValidator, on: [:check_answers, :submission]
+  validate on: :submission do
+    CaseDetails::AnswersValidator.new(self).validate
+  end
 
   def complete?
-    case_hearing_attributes_present? &&
-      case_concluded_attributes_present? &&
-      client_remanded_attributes_present? &&
-      preorder_work_attributes_present? &&
-      all_codefendants_complete? &&
-      all_charges_complete?
+    valid?(:submission)
   end
 
   private
-
-  def case_hearing_attributes_present?
-    values_at(:hearing_court_name, :hearing_date).all?(&:present?)
-  end
-
-  def all_codefendants_complete?
-    codefendants.all?(&:complete?)
-  end
-
-  def all_charges_complete?
-    charges.all?(&:complete?)
-  end
-
-  def case_concluded_attributes_present?
-    if has_case_concluded.nil? || has_case_concluded == YesNoAnswer::NO.to_s
-      true
-    elsif has_case_concluded == YesNoAnswer::YES.to_s
-      values_at(:date_case_concluded).all?(&:present?)
-    end
-  end
-
-  def preorder_work_attributes_present?
-    if is_preorder_work_claimed.nil? || is_preorder_work_claimed == YesNoAnswer::NO.to_s
-      true
-    elsif is_preorder_work_claimed == YesNoAnswer::YES.to_s
-      values_at(:preorder_work_date, :preorder_work_details).all?(&:present?)
-    end
-  end
-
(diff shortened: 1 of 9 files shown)


app/models/case.rb
```diff
@@ -16,55 +16,16 @@ def complete
               mapping: %i[hearing_court_name name],
               constructor: :find_by_name, allow_nil: true
 
-  validates_with CaseDetails::AnswersValidator, on: [:check_answers, :submission]
+  validate on: :submission do
+    CaseDetails::AnswersValidator.new(self).validate
+  end
```
hiboabd, line 21: Much better

app/validators/case_details/answers_validator.rb
```diff
@@ -1,7 +1,88 @@
...
+      errors.add(:first_court_hearing, :blank) unless first_court_hearing_complete?
+
+      errors.add :base, :incomplete_records unless errors.empty?
+    end
+
+    delegate :not_means_tested?, to: :crime_application
+
+    def case_concluded_complete?
+      return false if kase.has_case_concluded.blank?
+      return true unless kase.has_case_concluded == 'yes'
```
hiboabd, line 37: Should the condition be `has_case_concluded == 'no'`

hiboabd, line 37: This condition is a little different to the `preorder_work_complete` and `client_remanded_complete` below. The logic is correct but perhaps they should be the same for consistency?

```suggestion
      return true if kase.has_case_concluded == 'no'
```

app/validators/case_details/answers_validator.rb
```diff
@@ -1,7 +1,88 @@
...
+        :preorder_work_date,
+        :preorder_work_details
+      ).all?(&:present?)
+    end
+
+    def client_remanded_complete?
+      return false if kase.is_client_remanded.blank?
+      return true if kase.is_client_remanded == 'no'
+
+      kase.values_at(:date_client_remanded).all?(&:present?)
```
hiboabd, line 57: is the `.all?` needed? 
```suggestion
      kase.date_client_remanded.present?
```

timpeat: Ug! Sorry!

spec/validators/case_details/answers_validator_spec.rb
```diff
@@ -0,0 +1,403 @@
...
+
+        expect(errors).to receive(:add).with(:base, :incomplete_records)
+
+        subject.validate
+      end
+
+      context 'when application is not means tested' do
+        let(:not_means_tested) { true }
+
+        it 'adds errors for all failed validations' do
```
hiboabd, line 71: Is it worth adding `no_to` assertions for the ones that shouldn't be checked? That way if there's a regression the test would fail

hiboabd (approved):
Working for me locally, happy for this to go in once changes have been made

spec/validators/case_details/answers_validator_spec.rb
timpeat: It blows up on unexpected message if any error not listed is added, so not necessary.

app/validators/case_details/answers_validator.rb
timpeat: Thanks because the pre-order is dependent on the case concluded question and can be blank. (Where as the others can never be blank and complete.)

PR #839, opened 2024-05-08
CRIMAPP-859 Update employment status page

RoseSAK:
## Description of change

Remove three checkbox options and update content.

I haven't changed the routing yet (in the Figma if you select `Employed` then you continue down the `Income assessment` route) this will be updated later on when the rest of the Employed journey is actually available. For the time being I think it makes sense to keep the routing the same as it is currently i.e. if you select 'Employed or self-employed' then you are redirected to the 'Please use eForms' page.

## Link to relevant ticket

[CRIMAPP-859](https://dsdmoj.atlassian.net/browse/CRIMAPP-859)

### Before changes:
Screenshot 2024-05-08 at 17 50 32

### After changes:
Screenshot 2024-05-09 at 15 31 55

Navigate to the `Employed Status` page.  Check that the content, checkboxes and error messages are correct.

[CRIMAPP-859]: https://dsdmoj.atlassian.net/browse/CRIMAPP-859?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/value_objects/employment_status.rb
+++ b/app/value_objects/employment_status.rb
@@ -2,9 +2,6 @@ class EmploymentStatus < ValueObject
   VALUES = [
     EMPLOYED = new(:employed),
     SELF_EMPLOYED = new(:self_employed),
-    BUSINESS_PARTNERSHIP = new(:business_partnership),
-    DIRECTOR = new(:director),
-    SHAREHOLDER = new(:shareholder),
     NOT_WORKING = new(:not_working),
   ].freeze
 end
--- a/app/forms/steps/income/employment_status_form.rb
+++ b/app/forms/steps/income/employment_status_form.rb
@@ -29,7 +29,7 @@ module Steps
       private
 
       def validate_statuses
-        return unless employment_status.empty? || (employment_status - EmploymentStatus.values.map(&:to_s)).any?
+        return unless employment_status.empty? || (employment_status - choices.map(&:to_s)).any?
 
         errors.add(:employment_status,
                    :invalid)
--- a/app/services/evidence/rules/20240426012343_self_employed.rb
+++ b/app/services/evidence/rules/20240426012343_self_employed.rb
@@ -9,8 +9,6 @@ module Evidence
       client do |crime_application|
         self_employed_status = [
           EmploymentStatus::SELF_EMPLOYED.to_s,
-          EmploymentStatus::BUSINESS_PARTNERSHIP.to_s,
-          EmploymentStatus::SHAREHOLDER.to_s,
         ]
 
         self_employed_status.intersect?(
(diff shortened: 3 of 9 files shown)


app/value_objects/employment_status.rb
```diff
@@ -7,4 +7,10 @@ class EmploymentStatus < ValueObject
     SHAREHOLDER = new(:shareholder),
     NOT_WORKING = new(:not_working),
   ].freeze
+
+  INITIAL_STATUSES = [
+    EmploymentStatus::EMPLOYED,
+    EmploymentStatus::SELF_EMPLOYED,
+    EmploymentStatus::NOT_WORKING,
+  ].freeze
```
RoseSAK, line 15: I have named this `INITIAL_STATUSES` indicating that the other statuses might be selected/used later on i.e. in the self-employed journey but not totally happy with it and definitely open to better suggestions for a name

ripan:
@RoseSAK Minor: "After changes" screenshot needs to be updated in the PR description. It still shows "My client is not working" 😄

app/value_objects/employment_status.rb
ripan: @RoseSAK What do you think of removing BUSINESS_PARTNERSHIP, DIRECTOR and SHAREHOLDER from employment_status values array and create separate  self_employed_status (app/value_objects/self_employed_status.rb) value objects ?

RoseSAK:
> @RoseSAK Minor: "After changes" screenshot needs to be updated in the PR description. It still shows "My client is not working" 😄

@ripan that is really odd as I had updated it - just took a new screenshot and it's appearing correctly :) it must have been an old screenshot!

app/value_objects/employment_status.rb
RoseSAK: Yeah that would make sense, shall we leave that for the self-employed journey though? Not sure if just removing without replacing would break any existing in progress applications if someone has selected one of the options which are removed. Although they are just redirected to use eforms anyway right?

ripan: I don't think it will break anything. Lets go with removing BUSINESS_PARTNERSHIP, DIRECTOR and SHAREHOLDER from values array

arthurashman (approved):

app/value_objects/employment_status.rb
arthurashman: It probably would break in progress apps that had those selected - I've checked data and noone has one of those 3 so it's fine (right this sec)

ripan (approved):

PR #841, opened 2024-05-09
Fix tasklist issues for u18 and appeal no changes apps

hiboabd:
## Description of change

Fixes issue with task list for appeal no changes and u18 apps where the validator was marking the benefit type as incomplete when it was not asked for these application types


timpeat (approved):
🚀

PR #843, opened 2024-05-10
Require appeal details to be complete when applicable

timpeat:
## Description of change

Require appeal details to be complete where applicable.

## Link to relevant ticket
[CRIMAPP-828](https://dsdmoj.atlassian.net/browse/CRIMAPP-828)

## Notes for reviewer

Appeal details were not defined by the answers validator. This PR creates an appeal_details answers validator which is included in the ClientDetails::AnswersValidator. It also introduces a method  #appliable?, which could be added to all answer validators and used to determined if a given set of answers should be submitted as part of the final submission and could also be used by the Task list items.


hiboabd (approved):
Working for me locally 👍

PR #848, opened 2024-05-13
[CRIMAPP-840] Add unemployed partner attributes

hiboabd:
## Description of change
Adds attributes required for unemployed partner journey

## Link to relevant ticket
[CRIMAPP-840](https://dsdmoj.atlassian.net/browse/CRIMAPP-840)

Could test an application all the way to resubmission to check it hasn't broken anything

[CRIMAPP-840]: https://dsdmoj.atlassian.net/browse/CRIMAPP-840?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,7 @@ gem 'laa-criminal-applications-datastore-api-client',
 
 gem 'laa-criminal-legal-aid-schemas',
     github: 'ministryofjustice/laa-criminal-legal-aid-schemas',
-    ref: 'f3d14ada5d83b8c11a5a22740ebdccd74a887f1b'
+    tag: 'v1.1.0'
 
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,7 @@ gem 'laa-criminal-applications-datastore-api-client',
 
 gem 'laa-criminal-legal-aid-schemas',
     github: 'ministryofjustice/laa-criminal-legal-aid-schemas',
-    ref: 'ab0821284dc9dac96409748ced28698748766397'
+    ref: '666b41ac653915e6af2c832116633aa4763b767a'
 
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,7 @@ gem 'laa-criminal-applications-datastore-api-client',
 
 gem 'laa-criminal-legal-aid-schemas',
     github: 'ministryofjustice/laa-criminal-legal-aid-schemas',
-    ref: '666b41ac653915e6af2c832116633aa4763b767a'
+    ref: 'f3d14ada5d83b8c11a5a22740ebdccd74a887f1b'
 
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,7 @@ gem 'laa-criminal-applications-datastore-api-client',
 
 gem 'laa-criminal-legal-aid-schemas',
     github: 'ministryofjustice/laa-criminal-legal-aid-schemas',
-    tag: 'v1.0.75'
+    ref: 'ab0821284dc9dac96409748ced28698748766397'
 
 group :development, :test do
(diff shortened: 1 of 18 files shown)


Gemfile
```diff
@@ -50,7 +50,7 @@ gem 'laa-criminal-applications-datastore-api-client',
 
 gem 'laa-criminal-legal-aid-schemas',
     github: 'ministryofjustice/laa-criminal-legal-aid-schemas',
-    tag: 'v1.0.75'
+    ref: 'f3d14ada5d83b8c11a5a22740ebdccd74a887f1b'
```
hiboabd, line 53: Need to update this once schema changes are in

app/models/payment.rb
```diff
@@ -4,6 +4,7 @@ class Payment < ApplicationRecord
   attribute :amount, :pence
   attribute :payment_type
   attribute :frequency, :value_object, source: PaymentFrequencyType
+  attribute :ownership_type
```
hiboabd, line 7: do i need to add `:value_object, source: OwnershipType?` here

app/services/adapters/structs/income_details.rb
```diff
@@ -6,13 +6,17 @@ def employment_status
...
+
       def serializable_hash(options = {})
         super(
           options.merge(
-            methods: [:employment_status],
+            methods: [:employment_status, :partner_employment_status],
             # `employment_type` is the name for employment_status
             # in the datastore, we don't use it
-            except: [:employment_type, :dependants, :income_payments, :income_benefits]
+            except: [:employment_type, :partner_employment_type, :dependants, :income_payments, :income_benefits]
```
hiboabd, line 19: Not sure if i should just be naming this `partner_employment_status` in the schema?

app/value_objects/property_relationship_type.rb
```diff
@@ -1,4 +1,4 @@
-class RelationshipType < ValueObject
+class PropertyRelationshipType < ValueObject
```
hiboabd, line 1: Renamed as we will add a `RelationshipType` as part of the partner work

app/models/payment.rb
timpeat: If you want it casting correctly you do, but it does not make that much difference unless your asking .partner? question etc.

app/services/adapters/structs/income_details.rb
timpeat: The type was used in the datastore to determine which means assessment is required, rather than as a record of all employment relations. Personally, I think we should change the datastore to store all values as an array, similar to how it is done on Apply. We can then compute the work stream and value for the MAAT API within the datastore.

However, I don't think this change needs to be included in your current pull request.

app/services/datastore/application_rehydration.rb
```diff
@@ -70,7 +70,10 @@ def date_stamp
     end
 
     def applicant
```
timpeat, line 72: Lets not forget to do this!

timpeat (approved):
@hiboabd and @mseedat-moj I'm happy with the proposed data model here if it work for you both.

db/migrate/20240510150029_add_new_attributes.rb
```diff
@@ -0,0 +1,39 @@
...
+    add_column :capitals, :partner_trust_fund_amount_held, :bigint
+    add_column :capitals, :partner_has_premium_bonds, :string
+    add_column :capitals, :partner_premium_bonds_total_value, :bigint
+    add_column :capitals, :partner_premium_bonds_holder_number, :string
+
+    # Reset People/CrimeApplication index to include type column for STI
+    remove_index :people, column: [:crime_application_id], name: 'index_people_on_crime_application_id'
+    add_index :people, [:type, :crime_application_id], unique: true
+  end
+end
```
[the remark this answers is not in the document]
hiboabd: @mseedat-moj the idea is to default `has_partner` to 'no' and then we will remove the `client_has_partner` attribute from the `crime_applications` table so there will be no partner attributes in the `crime_applications` table

If there is no partner, the `has_partner`, `relationship_status` and `separation_date` fields will be set. I think this is better than having some partner related fields in the `crime_applications` table and some in the `partner_details` table but let me know if you object

app/services/adapters/structs/income_details.rb
arthurashman: I'd be keen to consolidate employment_type and employment_status if we don't think we need 2 different attributes. It will simplify this point a lot

PR #854, opened 2024-05-15
[CRIMAPP-889] Add partner benefit type screen

hiboabd:
## Description of change
Adds screen to select partners benefit type 
Also updates routing of related screens

## Link to relevant ticket
[CRIMAPP-889](https://dsdmoj.atlassian.net/browse/CRIMAPP-889)

### After changes:
Screenshot 2024-05-15 at 14 01 09
Screenshot 2024-05-15 at 14 01 17
Screenshot 2024-05-15 at 14 01 48

To test the screen itself 

1. Start an application and fill in client details and navigate to `/steps/dwp/partner_benefit_type `
2. Verify the validation is in place when no option is selected and when no jsa date is inputted 
3. Select an option and select save and continue and verify you are able to continue to another screen 

To test the routing 

1. Start an application and select yes to having a partner. 
2. You will be shown the partner exit screen but you can use the url to navigate to the task list to begin the application (`/edit`)
3. Create a over 18 client, summary only application where the client does not have a nino or a passporting benefit 
4. After selecting no benefit on the client passporting benefit screen, you should be routed to the partner passporting benefit screen 
5. Select a benefit for the partner and you should be navigated to the dwp check 
6. Select none for the partner and you should be routed to the case urn screen 

[CRIMAPP-889]: https://dsdmoj.atlassian.net/browse/CRIMAPP-889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -16,4 +16,8 @@ class Person < ApplicationRecord
   def correspondence_address?
     correspondence_address&.address_line_one.present?
   end
+
+  def has_passporting_benefit?
+    BenefitType.passporting.map(&:value).include?(benefit_type&.to_sym)
+  end
 end
--- a/app/services/decisions/dwp_decision_tree.rb
+++ b/app/services/decisions/dwp_decision_tree.rb
@@ -1,10 +1,13 @@
 module Decisions
+  # rubocop:disable Metrics/ClassLength
   class DWPDecisionTree < BaseDecisionTree
     # rubocop:disable Metrics/MethodLength, Metrics/CyclomaticComplexity
     def destination
       case step_name
       when :benefit_type
         after_benefit_type
+      when :partner_benefit_type
+        after_partner_benefit_type
       when :benefit_check_result
         edit('/steps/case/urn')
       when :has_benefit_evidence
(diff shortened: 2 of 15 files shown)


app/models/person.rb
```diff
@@ -16,4 +16,8 @@ def home_address?
   def correspondence_address?
     correspondence_address&.address_line_one.present?
   end
+
+  def has_passporting_benefit?
+    BenefitType.passporting.map(&:value).include?(benefit_type&.to_sym)
+  end
```
hiboabd, line 22: Moved from applicant to person as both applicant and partner will use this method

config/locales/en/errors.yml
```diff
@@ -148,6 +148,24 @@ en:
               invalid_year: Enter a valid year
               year_too_early: Date your client last attended an appointment with their JSA work coach is too far in the past
               future_not_allowed: Date your client last attended an appointment with their JSA work coach must be today or in the past
+        steps/dwp/partner_benefit_type_form:
```
hiboabd, line 151: This will probably require a helper but I'm deferring refactoring in this pr as I think the solution for how best to refactor this to avoid duplication will become clearer when we start implementing the shared partner client forms.

app/models/person.rb
timpeat: It would be good to update the method in TypeOfMeansAssessent to use this method and be partner aware.

app/services/decisions/dwp_decision_tree.rb
```diff
@@ -46,19 +49,37 @@ def after_confirm_details
...
+      benefit_check_routing(applicant)
+    end
+
+    def after_partner_benefit_type
+      benefit_check_routing(partner)
+    end
+
+    def benefit_check_routing(person)
       if form_object.benefit_type.none?
         if FeatureFlags.means_journey.enabled?
```
timpeat, line 63: I'd really like us to remove this Means feature flag. I certainly wouldn't want to trust that turning it off would work. We would have to do a massive amount of regression testing, etc.

app/services/decisions/dwp_decision_tree.rb
```diff
@@ -46,19 +49,37 @@ def after_confirm_details
...
       else
-        determine_dwp_result_page
+        determine_dwp_result_page(person)
       end
     end
 
+    def partner_benefit_type_required?
+      # TODO: refactor to use has_partner in partner_details table instead of client_has_partner
+
+      form_object.benefit_type.none? &&
```
timpeat, line 78: I think it would be good to do the task list side of this next. I suspect quite a bit of this logic will need to be shared with the answers validator.

timpeat (approved):
I think this is a good first step. Be good to move onto the task list side of this next as I think that will drive out some of the shared logic we need.

app/services/decisions/dwp_decision_tree.rb
hiboabd: Me too!

app/models/person.rb
hiboabd: Will do this in the next ticket i've picked up or with the task list updates

PR #865, opened 2024-05-20
CRIMAPP-917 Nino not required when passported on age

timpeat:
## Description of change

Fix bug where Nino was incorrectly being required when benefit_type nil.
Also fixes a regression where applicants without NINO and not in custody could access the declaration page from the task list.

## Link to relevant ticket
[CRIMAPP-917](https://dsdmoj.atlassian.net/browse/CRIMAPP-917)

See ticket


arthurashman (approved):
Code looks good. I have not tested on local though

PR #868, opened 2024-05-21
CRIMAPP-869 - Remove payment association from employments

ripan:
## Description of change
Remove payment association from employments

Reason:
`PG::UniqueViolation: ERROR: duplicate key value violates unique constraint "index_payments_crime_application_id_and_payment_type"  is stopping me to use payments table for employments`

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-869


app/views/steps/income/client/employment_details/edit.html.erb
```diff
@@ -7,14 +7,12 @@
     <%= govuk_error_summary(@form_object) %>
 
     <h1 class="govuk-heading-xl"><%=t '.heading' %></h1>
-
+    <% @form_object.before_or_after_tax = @form_object.record.before_or_after_tax['value'] if @form_object.record.before_or_after_tax %>
```
RoseSAK, line 10: Personally I'd prefer to have this logic in the Form class rather than the view but probably just a preference thing!

RoseSAK (approved):
Looks good! 💯

app/views/steps/income/client/employment_details/edit.html.erb
ripan: 👍 Agree. Even i am liking this solution. Will fix it with another PR

PR #874, opened 2024-05-22
CRIMAPP-937 review task locked until other tasks fulfilled

timpeat:
## Description of change

The 'Review the application’ task cannot be started until all required sections have been completed.
Also, only mark 'more information' task as completed if more information has been provided when answered 'yes'.

## Link to relevant ticket

[CRIMAPP-937](https://dsdmoj.atlassian.net/browse/CRIMAPP-937)

## Notes for reviewer

Previously, if an application was Means Passported or if the means assessment had started, a provider could navigate to the ‘Review the application’ page from the Task List.

However, UCD has confirmed that this is not the desired behavior. The ‘Review the application’ task should be marked as ‘cannot yet start’ until all required sections have been completed. This means that all prior sections, except for ‘Supporting evidence,’ must be marked as either ‘complete’ or ‘not applicable’.

Check the ‘Review the application’ task on the task list after each task in the process of completing a Means Passported application.

Confirm that the ‘Review the application’ task is marked as ‘in progress’ only after the ‘More information’ section is complete.

Go back and change the details so that the application is no longer Means Passported.

Confirm that the ‘Review the application’ task is marked as ‘cannot yet start’ until the means assessment is completed.

Repeat the same process for PSE, ensuring that only the ‘Supporting evidence’ and ‘Additional information’ sections are required.

[CRIMAPP-937]: https://dsdmoj.atlassian.net/browse/CRIMAPP-937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/presenters/tasks/capital_assessment.rb
+++ b/app/presenters/tasks/capital_assessment.rb
@@ -22,6 +22,12 @@ module Tasks
       capital.present?
     end
 
+    # Capital task is only complete when both the AnswersValidator
+    # AND the ConfirmationValidators are complete
+    def completed?
+      super && capital.complete?
+    end
+
     private
 
     def validator
--- a/app/presenters/tasks/declaration.rb
+++ b/app/presenters/tasks/declaration.rb
@@ -9,11 +9,7 @@ module Tasks
     end
 
     def can_start?
-      if crime_application.pse?
-        fulfilled?(EvidenceUpload)
-      else
-        fulfilled?(CaseDetails) && crime_application.valid?(:submission)
-      end
+      fulfilled?(Review)
     end
 
     def in_progress?
--- a/app/models/crime_application.rb
+++ b/app/models/crime_application.rb
@@ -74,7 +74,8 @@ class CrimeApplication < ApplicationRecord
   validates_with PseFulfilmentValidator, on: :submission, if: :post_submission_evidence?
 
   validate on: :client_details do
-    ::ClientDetails::AnswersValidator.new(self).validate
+    ::ClientDetails::AnswersValidator.new(record: self, crime_application: self)
+                                     .validate
   end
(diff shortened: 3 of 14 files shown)


app/presenters/tasks/capital_assessment.rb
```diff
@@ -22,6 +22,12 @@ def in_progress?
       capital.present?
     end
 
+    # Capital task is only complete when both the AnswersValidator
```
timpeat, line 25: This was resulting in the task list incorrectly showing the capital section as complete when the confirmation question had not been added.

app/presenters/tasks/declaration.rb
```diff
@@ -9,11 +9,7 @@ def not_applicable?
     end
 
     def can_start?
-      if crime_application.pse?
-        fulfilled?(EvidenceUpload)
-      else
-        fulfilled?(CaseDetails) && crime_application.valid?(:submission)
-      end
+      fulfilled?(Review)
```
timpeat, line 12: The review task now has a good understanding of which tasks are applicable and complete. So we can defer to that task for the declaration page.

app/presenters/tasks/capital_assessment.rb
hiboabd: If we don't save the confirmation question value, won't this be unreliable to use?

timpeat: https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/blob/main/db/schema.rb#L49

hiboabd (approved):
👍

PR #889, opened 2024-06-03
Update partner routing

hiboabd:
## Description of change

- Routes to the task list when an application started if the partner journey is enabled and routes to the does client have partner page if the partner journey is not enabled
- Removes partner journey feature flags that are no longer required 
- Sets non means tested ff to false on staging to allow for regression testing on staging and prepare for deployment

Toggle the partner journey feature flag and test the routing is as expected for different application types (means assessed application, non means assessed application, u18 application, appeal no changes etc..)


timpeat (approved):
Thank you :)

PR #894, opened 2024-06-04
Temp fix for sentry error when rehydrating partner fields

hiboabd:
## Description of change
Fixes sentry error when rehydrating an application: https://mojdt.slack.com/archives/C056U956ESH/p1717495014170419

### DIFF
--- a/app/services/datastore/application_rehydration.rb
+++ b/app/services/datastore/application_rehydration.rb
@@ -87,6 +87,7 @@ module Datastore
     end
 
     def partner
+      return nil unless FeatureFlags.partner_journey.enabled?
       return nil unless parent.partner && parent.applicant.has_partner == 'yes'
 
       attributes = parent.partner.serializable_hash.except!(*PartnerDetail.fields)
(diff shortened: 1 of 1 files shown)


app/services/datastore/application_rehydration.rb
```diff
@@ -95,11 +96,13 @@ def partner
...
     # NOTE: Actual partner_detail fields are mixed between the Applicant and Partner Structs
     def partner_detail
+      return nil unless FeatureFlags.partner_journey.enabled?
+
       fields_from_applicant = %w[has_partner relationship_to_partner relationship_status separation_date]
       fields_from_partner = %w[involvement_in_case conflict_of_interest has_same_address_as_client]
 
       from_applicant = parent.applicant.serializable_hash.slice(*fields_from_applicant)
-      from_partner = parent.partner.serializable_hash.slice(*fields_from_partner)
+      from_partner = parent.partner.serializable_hash.slice(*fields_from_partner) if parent.partner
```
hiboabd, line 105: I haven't been able to recreate this locally. The error appears to be originating from this line. I think the fix would be to check if partner is present but I have also put the partner rehydration methods behind a feature flag so we can test this properly.

RoseSAK (approved):

PR #897, opened 2024-06-04
[CRIMAPP-890] Unemployed partner dwp journey

hiboabd:
## Description of change

- Routing and content changes for unemployed partner dwp journey 
- Moves the `confirm_dwp_result` attribute to the person table to follow pattern of other dwp attributes
- Also covers rehydration of partner dwp attributes

Note: cannot be merged until schema changes are merged 

## Link to relevant ticket
[CRIMAPP-890](https://dsdmoj.atlassian.net/browse/CRIMAPP-890)
[CRIMAPP-895](https://dsdmoj.atlassian.net/browse/CRIMAPP-895)

### After changes:

Screenshot 2024-06-06 at 12 34 12
Screenshot 2024-06-06 at 12 34 29
Screenshot 2024-06-06 at 12 34 37
Screenshot 2024-06-06 at 12 35 02
Screenshot 2024-06-06 at 12 35 13
Screenshot 2024-06-06 at 12 35 42

Test the scenarios shown [here in figma](https://www.figma.com/design/thkvnDBbQbHlSqzt1TE3lk/Crime-Apply?node-id=13705-65710&t=QLp9kyj63Y6lJAIG-0), requires schema changes in this pr 

[CRIMAPP-890]: https://dsdmoj.atlassian.net/browse/CRIMAPP-890?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CRIMAPP-895]: https://dsdmoj.atlassian.net/browse/CRIMAPP-895?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/forms/steps/dwp/benefit_type_form.rb
```diff
@@ -34,7 +34,7 @@ def persist!
           attributes.merge(attributes_to_reset)
         )
 
-        crime_application.update(confirm_dwp_result: nil)
+        crime_application.partner&.update(attributes_to_reset)
```
hiboabd, line 37: If a user was to go back to the client benefit type page and select a passporting benefit when they had already entered partner benefit details, it would need to be reset

app/forms/steps/shared/no_op_form.rb
```diff
@@ -4,7 +4,8 @@ class NoOpForm < Steps::BaseFormObject
       # NOTE: steps using this form do not persist anything to DB.
       # It is only used to advance in the decision tree.
       # Normally used through the `Steps::NoOpAdvanceStep` concern.
-      #
+      include Steps::ApplicantOrPartnerExclusive
```
hiboabd, line 7: Not sure about this including this in this shared form. There are two dwp screens that do not have their own form as they don't persist anything (the cannot check dwp status and confirmed benefit check result). They use this form along with two other forms in our codebase (4 in total). So overall doesn't impact a lot of forms but this could increase in the future...

app/models/crime_application.rb
```diff
@@ -123,4 +123,10 @@ def applicant_18_or_over_at_date_stamp?
     age >= 18
   end
   # rubocop:enable Metrics/AbcSize
+
+  def benefit_check_recipient
+    return partner if partner.present? && partner.has_passporting_benefit?
+
+    applicant
```
hiboabd, line 130: The logic for this might need refactoring, I think it will be clearer when working on the cya page and tasklist

app/services/dwp/benefit_check_status_service.rb
```diff
@@ -2,35 +2,53 @@ module DWP
...
-      @applicant = applicant
+      @person = person
     end
 
     def self.call(*args)
       new(*args).call
     end
 
     def call
+      # TODO: determine when status should be calculated
```
hiboabd, line 15: I think this will also be clearer when working on the summary component

db/migrate/20240606080623_move_confirm_dwp_result_to_person.rb
```diff
@@ -0,0 +1,6 @@
+class MoveConfirmDWPResultToPerson < ActiveRecord::Migration[7.0]
+  def change
+    remove_column :crime_applications, :confirm_dwp_result
```
hiboabd, line 3: I think this is safe to remove as we do not pass this to review and so only draft apps may have had this value set

app/controllers/steps/dwp_step_controller.rb
```diff
@@ -5,5 +5,11 @@ class DWPStepController < BaseStepController
     def decision_tree_class
       Decisions::DWPDecisionTree
     end
+
+    def benefit_check_on_partner?
+      return true if current_crime_application.partner&.has_passporting_benefit?
```
timpeat, line 10: @hiboabd is the DWP check run when the partner has a contrary interest? If not, I think we should add an `include_partner_in_means_assessment?` guard here as well. Even if it is possible to use `has_passporting_benefit` to infer this, I think the code should explicitly use that guard here.

app/forms/concerns/steps/applicant_or_partner_exclusive.rb
```diff
@@ -0,0 +1,27 @@
+module Steps
+  module ApplicantOrPartnerExclusive
```
timpeat, line 2: 👍

app/forms/steps/shared/no_op_form.rb
timpeat: I think we need to find an alternative to including this here. Happy to pair.

app/models/concerns/type_of_means_assessment.rb
```diff
@@ -47,30 +48,30 @@ def evidence_of_passporting_means_forthcoming?
   # blocked until the NINO or benefit evidence is provided.
   def nino_forthcoming?
     return false unless has_passporting_benefit?
-    return false unless applicant.has_nino == 'no'
-    return false if applicant.will_enter_nino == 'yes'
+    return false unless benefit_check_recipient.has_nino == 'no'
+    return false if benefit_check_recipient.will_enter_nino == 'yes'
 
-    applicant.will_enter_nino == 'no' || kase.is_client_remanded == 'yes'
+    benefit_check_recipient.will_enter_nino == 'no' || kase.is_client_remanded == 'yes'
```
timpeat, line 54: I'm wondering if it would be better to have a dedicated object for the benefit check questions instead of referring to a person.

This way, we can ask questions like benefit_check.nino_forthcoming?. The logic to understand who the check is on could also live there instead of within the crime application.

app/models/crime_application.rb
timpeat: Could this live in TypeOfMeansAssessment for no? (Although I do think we should also explore a separate object to model the benefit check which would understand who the check was being run against, where there was forthcoming information etc.)

timpeat (commented):
Looking really good @hiboabd :) .

PR #905, opened 2024-06-06
[CRIMAPP-958] Add partner paid higher tax

arthurashman:
## Description of change
Add partner paid higher tax
Serialize and rehydrate
Add to CYAs and validator

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-958

## Notes for reviewer
We might need a holistic way (on submission like when means answers no longer relevant) to not submit any partner only qs


app/services/decisions/outgoings_decision_tree.rb
```diff
@@ -45,5 +55,12 @@ def after_answers
...
       form_object.crime_application
     end
+
+    def partner_relevant
+      return if partner_detail.nil?
+
+      partner_detail.involvement_in_case == 'none' ||
+        (partner_detail.involvement_in_case == 'codefendant' && partner_detail.conflict_of_interest == 'no')
+    end
   end
```
hiboabd, line 65: The `include_partner_in_means_assessment?` method in the Type of means assessment module would take care of this check for you

app/validators/outgoings_assessment/answers_validator.rb
```diff
@@ -74,11 +75,31 @@ def income_tax_rate_complete?
       record.income_tax_rate_above_threshold.present?
     end
 
+    def partner_income_tax_rate_complete?
+      return true unless FeatureFlags.partner_journey.enabled?
+      return true unless partner_relevant
```
hiboabd, line 80: Same here

spec/services/decisions/outgoings_decision_tree_spec.rb
```diff
@@ -124,6 +128,41 @@
...
+
+      context 'when partner is codefendant with a conflict' do
+        let(:involvement_in_case) { 'codefendant' }
+        let(:conflict_of_interest) { 'yes' }
+
+        it { is_expected.to have_destination(:outgoings_more_than_income, :edit, id: crime_application) }
+      end
+    end
+
+    context 'and there is a relvant partner' do
```
hiboabd, line 153: relevant

spec/validators/outgoings_assessment/answers_validator_spec.rb
```diff
@@ -312,6 +341,30 @@
...
+      before do
+        allow(record).to receive(:partner_income_tax_rate_above_threshold).and_return(true)
+      end
+
+      it 'returns true' do
+        expect(subject.partner_income_tax_rate_complete?).to be(true)
+      end
+    end
+
+    context 'when partner_income_tax_rate_above_threshold is nil' do
```
hiboabd, line 357: Would it be possible for this value to be nil if the partner was part of the means assessment?

config/locales/en/summary.yml
```diff
@@ -535,15 +535,19 @@ en:
...
         answers:
           <<: *YESNO
       outgoings_more_than_income:
-        question: Are your client’s outgoings more than their income?
+        question: Outgoings more than income?
         answers:
           <<: *YESNO
       how_manage:
-        question: How does your client manage if their outgoings are more than their income?
+        question: How they cover outgoings that are more than their income
```
hiboabd, line 550: Could the text have remained the same as before just with `does your client` changed to `do they`

app/services/decisions/outgoings_decision_tree.rb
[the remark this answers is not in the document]
arthurashman: Thank you! I was looking for that!

config/locales/en/summary.yml
arthurashman: This is from the figma - so I just did as I was told! 😂

spec/validators/outgoings_assessment/answers_validator_spec.rb
arthurashman: Thinking save and come back later I guess

config/locales/en/summary.yml
hiboabd: Ah my bad haha

PR #918, opened 2024-06-11
[CRIMAPP-894] Unemployed partner form fixes 

hiboabd:
## Description of change
Some fixes for a couple of things I've noticed when submitting applications 
- Subject translation fix for the other investment type page 
- The national savings certificate form was displaying three ownership types when it should not display applicant and partner
- The text `confirmed client details correct?` was specific to client so updated to make more generic


timpeat (approved):
How did I miss that?

PR #927, opened 2024-06-13
[CRIMAPP-1044] Fix outgoings summary title

mseedat-moj:
## Description of change
Fix outgoings summary title

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1044
Revision as a result of this PR: https://dsdmoj.atlassian.net/browse/CRIMAPP-1082 (not yet implemented)

## Notes for reviewer
Should fix any other similar bugs
### Before changes:
Screenshot 2024-06-12 at 15 16 59-20240612-141705

### After changes:

Screenshot 2024-06-13 at 10 52 25
Screenshot 2024-06-13 at 10 54 33

Complete outgoings

### DIFF
--- a/app/services/evidence/rules/20240425220057_salaried_employee.rb
+++ b/app/services/evidence/rules/20240425220057_salaried_employee.rb
@@ -16,9 +16,14 @@ module Evidence
         end
       end
 
-      # TODO: Awaiting partner implementation
-      partner do |_crime_application|
-        false
+      partner do |crime_application|
+        if MeansStatus.include_partner?(crime_application) && crime_application.income
+          [EmploymentStatus::EMPLOYED.to_s].intersect?(
+            Array.wrap(crime_application.income.partner_employment_status)
+          )
+        else
+          false
+        end
       end
     end
   end
--- a/app/services/evidence/rules/20240426002316_self_assessed.rb
+++ b/app/services/evidence/rules/20240426002316_self_assessed.rb
@@ -7,13 +7,11 @@ module Evidence
       group :income
 
       client do |crime_application|
-        if crime_application.outgoings
-          (crime_application.outgoings.income_tax_rate_above_threshold == 'yes') || false
-        else
-          false
-        end
+        crime_application.outgoings&.income_tax_rate_above_threshold == 'yes' ||
+          crime_application.income&.applicant_self_assessment_tax_bill == 'yes'
       end
 
+      # TODO: awaiting partner_self_assessment_tax_bill
       partner do |crime_application|
         if MeansStatus.include_partner?(crime_application) && crime_application.outgoings
           (crime_application.outgoings.partner_income_tax_rate_above_threshold == 'yes') || false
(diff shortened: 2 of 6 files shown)


app/presenters/summary/sections/base_section.rb
```diff
@@ -29,7 +29,7 @@ def title
         I18n.t(
           name,
           scope: 'summary.sections',
-          subject: I18n.t('summary.dictionary.subject', subject_type:),
+          subject: I18n.t("summary.dictionary.subjects.#{subject_type}"),
```
timpeat, line 32: @mseedat-moj is this change necessary? This is handled here: 
https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/blob/main/config/locales/en.rb

and is used elsewhere I think.

timpeat: @mseedat-moj forgive me, I see this is for the summary dictionary.

config/locales/en/summary.yml
```diff
@@ -66,13 +66,13 @@ en:
...
       income_payments_details: Payments your client gets
       income_benefits_details: Benefits your client gets
-      other_income_details: 
+      other_income_details:
         one: "Other sources of income"
         other: "Other sources of income: client and partner"
       housing_payments: Housing Payments
-      outgoings_payments_details: 
-        one: Payments %{subject} makes
-        other: Payments %{subject} make
```
timpeat, line 75: This will incorrectly produce the title "Payments your partner makes" I think. Given this has a different subject (__your__ client, __their__ partner) to other titles (partner, client). I'd be inclined to overwrite the title method in the child class with something that uses the same subject as the forms. I.e.

def title
  I18n.t(
          name,
          scope: 'summary.sections',
          subject: I18n.t("dictionary.subjects.#{subject_type}")
        ) 
        end

timpeat: Also. I don't think think this text will every be plural.

ripan (approved):

app/presenters/summary/sections/base_section.rb
arthurashman, line 32: I'm not sure subject_type is working on the outgoings page. The page title knows that it is the client and partner, but the section title doesn't
image

arthurashman: I think we want ownership too, not subject looking at the dictionary

config/locales/en/summary.yml
arthurashman: Pluralisation is to avoid:
`Payments your client and their partner **make**`

timpeat: Yes @arthurashman but in this context it will only ever be client OR partner. No?

config/locales/en/summary.yml
```diff
@@ -66,13 +66,13 @@ en:
...
+      outgoings_payments_details:
+        one: Payments your %{subject} makes
+        other: Payments your %{subject} make
```
arthurashman, line 75: If we switch to ownership from the dictionary, you need to cut these 'your's

config/locales/en/summary.yml
arthurashman: Nope, for outgoings the payments page is either `Client` or `Client and Partner`

PR #953, opened 2024-06-18
Only display partner 40% tax if there is an included partner

arthurashman:
## Description of change
When feature flags are set to production values, outgoings answer playback view cause issues with viewing and submitting an application.

Only display partner 40% tax if there is an included partner

Set feature flags to prod values (particularly partner journey)
Restart Apply server
Submit an application (try some varieties)

### DIFF
--- a/app/presenters/summary/sections/other_outgoings_details.rb
+++ b/app/presenters/summary/sections/other_outgoings_details.rb
@@ -1,21 +1,31 @@
 module Summary
   module Sections
     class OtherOutgoingsDetails < Sections::BaseSection
+      include TypeOfMeansAssessment
+
       def show?
         outgoings.present? && super
       end
 
-      # rubocop:disable Metrics/MethodLength
+      # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
       def answers
-        [
+        answers = [
           Components::ValueAnswer.new(
             :income_tax_rate_above_threshold, outgoings.income_tax_rate_above_threshold,
             change_path: edit_steps_outgoings_income_tax_rate_path
           ),
-          Components::ValueAnswer.new(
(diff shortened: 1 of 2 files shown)


app/presenters/summary/sections/other_outgoings_details.rb
```diff
@@ -1,21 +1,31 @@
...
             :income_tax_rate_above_threshold, outgoings.income_tax_rate_above_threshold,
             change_path: edit_steps_outgoings_income_tax_rate_path
           ),
-          Components::ValueAnswer.new(
-            :partner_income_tax_rate_above_threshold, outgoings.partner_income_tax_rate_above_threshold,
-            change_path: edit_steps_outgoings_partner_income_tax_rate_path
-          ),
+        ]
+
+        if FeatureFlags.partner_journey.enabled? && include_partner_in_means_assessment?
```
hiboabd, line 19: I'd prefer if this was based on whether the attribute `partner_income_tax_rate_above_threshold` is present rather than if the partner is included in the assessment

arthurashman: Hmm hmm, just wondering whether what I've got will cover off if they answer it then go back and change the partner status? I'm not super bothered either way so long as the partner means are reset effectively

hiboabd: True, I don't think we are resetting these attributes actually 😅 ok good with me!

hiboabd (approved):

app/presenters/summary/sections/other_outgoings_details.rb
arthurashman: TY! Can take another look later

PR #954, opened 2024-06-19
Allow u18 to pass partner validator

arthurashman:
## Description of change
Allow u18 to pass partner validator

Client details DoB 01/01/2013
Complete rest of application
Confirm no errors on CYA
Confirm can submit

### DIFF
--- a/app/models/concerns/type_of_means_assessment.rb
+++ b/app/models/concerns/type_of_means_assessment.rb
@@ -125,4 +125,8 @@ module TypeOfMeansAssessment
   def income_above_threshold?
     !income_below_threshold?
   end
+
+  def client_has_under18_passport?
+    crime_application.means_passport.include?('on_age_under18')
+  end
 end
--- a/app/validators/sections_completeness_validator.rb
+++ b/app/validators/sections_completeness_validator.rb
@@ -31,6 +31,7 @@ class SectionsCompletenessValidator
 
   def partner_detail_complete?
     return true unless FeatureFlags.partner_journey.enabled?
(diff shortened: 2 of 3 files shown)


app/models/concerns/type_of_means_assessment.rb
```diff
@@ -125,4 +125,8 @@ def income_below_threshold?
   def income_above_threshold?
     !income_below_threshold?
   end
+
+  def client_has_under18_passport?
+    crime_application.means_passport.include?('on_age_under18')
```
arthurashman, line 130: Is this the best measure of whether or not they will have been asked the Partner questions?

If the client is u18 on original application, but over 18 on resubmission, we honour this original passport which is why I'm using it rather than a u18 value, or DoB calculation. 

Can anyone think of any edge cases where this passport either isn't there, or isn't accurate?

hiboabd: I can't think of any edge cases. You could also check whether the applicant is under 18 in case it wouldn't be present?

Looking at the other answers validators we are also not very consistent in how we handle this. We should agree on a approach and clean it up. This doesn't have to block this pr though which I think is good to go in but we could do with a spec on this method

arthurashman: @hiboabd I started looking at writing a test for this and it seemed a bit pointless, as it would just be me setting a pssport as ['on_age_under18'] then pulling it back out again so I decided not to write it. These other little helpers in this file are test free too

arthurashman:
Old and fixed subsequently in another PR

PR #957, opened 2024-06-19
CRIMAPP-1007 Add partner other work benefits page

RoseSAK:
## Description of change

Add page to add any other work benefits for the partner.

## Link to relevant ticket

[CRIMAPP-1007](https://dsdmoj.atlassian.net/browse/CRIMAPP-1007)

### After changes:

Screenshot 2024-06-19 at 16 16 52

[CRIMAPP-1007]: https://dsdmoj.atlassian.net/browse/CRIMAPP-1007?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/decisions/income_decision_tree.rb
+++ b/app/services/decisions/income_decision_tree.rb
@@ -39,10 +39,11 @@ module Decisions
       when :client_self_assessment_tax_bill
         edit(:other_work_benefits)
       when :partner_self_assessment_tax_bill
-        # TODO: Implement next step for partner income payments
-        edit('/steps/income/income_payments')
+        edit(:other_work_benefits)
       when :client_other_work_benefits
         edit('/steps/income/income_payments')
+      when :partner_other_work_benefits
+        edit('/steps/income/income_payments_partner')
       when :income_payments
         edit(:income_benefits)
       when :income_benefits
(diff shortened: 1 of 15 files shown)


app/services/decisions/income_decision_tree.rb
```diff
@@ -40,9 +40,11 @@ def destination # rubocop:disable Metrics/PerceivedComplexity
         edit(:other_work_benefits)
       when :partner_self_assessment_tax_bill
         # TODO: Implement next step for partner income payments
```
hiboabd, line 42: Can this todo be removed? 🤔

config/locales/en/errors.yml
```diff
@@ -532,6 +532,12 @@ en:
               inclusion: Select yes if your client receives any other benefits from work
             amount:
               not_a_number: Enter the total yearly value of their other work benefits
+        steps/income/partner/other_work_benefits_form:
+          attributes:
+            partner_other_work_benefit_received:
+              inclusion: Select yes if your client's partner receives any other benefits from work
```
hiboabd, line 538: I think this being different from the question text which uses `the partner` is a bit awkward. Would be good to check if it can be changed to match the question text

hiboabd (commented):
LGTM just some small comments

config/locales/en/errors.yml
RoseSAK: yes makes sense, have changed it to `Select yes if they receive any other benefits from work`

PR #959, opened 2024-06-19
[CRIMAPP-1031] Force partner benefit_type reset

mseedat-moj:
## Description of change
Resets partner benefit_type when the client selects 'none' for 'Does your client get one of these passporting benefits?'

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1031

## Notes for reviewer
Added a reset and transaction - not sure if this is the best way of resetting values in the partner when the client answers with particular values.


arthurashman (approved):
Nice, ty 👍

PR #974, opened 2024-06-26
[CRIMAPP-1044/1083/1118] Hardcode output for subject_type when applicant

mseedat-moj:
Results in

## Description of change
Hard-codes the `applicant` for summary panels that will only show the client/applicant outgoings

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1118
https://dsdmoj.atlassian.net/browse/CRIMAPP-1083
https://dsdmoj.atlassian.net/browse/CRIMAPP-1044

## Notes for reviewer
Assumes the output must be for the applicant or applicant/partner only. Partner has their own output panels

### DIFF
--- a/app/forms/steps/income/client/employment_income_form.rb
+++ b/app/forms/steps/income/client/employment_income_form.rb
@@ -17,7 +17,7 @@ module Steps
         end
 
         def self.build(crime_application)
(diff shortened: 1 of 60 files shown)


timpeat (approved):
Looks good. You could use 'return super unless incl....' as well

PR #1020, opened 2024-07-06
Remove has partner feature flag

arthurashman:
## Description of change
Remove has partner feature flag

Because the partner question moved in flow it makes the routing if partner on vs off conditionals really tricky to maintain - especially once combined with non-means.

### DIFF
--- a/app/services/decisions/client_decision_tree.rb
+++ b/app/services/decisions/client_decision_tree.rb
@@ -37,13 +37,19 @@ module Decisions
 
     private
 
-    # TODO: New partner details journey will cause the existing `has_partner` question
-    # to move further along the client journey rather than at the beginning
+    def after_client_details
+      if DateStamper.new(form_object.crime_application).call
+        edit(:date_stamp)
+      elsif form_object.crime_application.not_means_tested?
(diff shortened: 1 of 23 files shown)


app/services/decisions/client_decision_tree.rb
```diff
@@ -133,37 +129,22 @@ def start_address_journey(address_class)
...
     end
-
-    # TODO: Consider whether !crime_application.age_passported? is more appropriate?
-    def start_partner_journey?
-      return false unless FeatureFlags.partner_journey.enabled?
-
-      form_object.client_has_partner.yes? &&
-        !current_crime_application.not_means_tested? &&
-        !applicant.under18?
-    end
```
timpeat, line 166: What's the reason we no long care about u18 etc? Is it because we would not get to this point now the position of the has_partner question has moved?

timpeat (approved):
Nice :)

app/services/decisions/client_decision_tree.rb
arthurashman: That's right, has partner question won't be shown to u18 or non-means applicants

PR #1022, opened 2024-07-08
Guard against there being no income object

arthurashman:
## Description of change
Guard against there being no income object


timpeat (commented):
@arthurashman What is the scenario that causes this to be required?

timpeat (approved):

PR #1035, opened 2024-07-12
[CRIMAPP-1663] Add new CIFC routes and related pages

mseedat-moj:
## Description of change
Introduces Change In Financial Services application type and related questions.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-533

## Notes for reviewer

1. First question `Are you making a new application or telling us about a change in financial circumstances?` sits outside of the `steps` form system
2. Prior to this change, when the user clicked on `Make a new application`, a new empty Crime Application was immediately persisted. This implementation changes behaviour so that clicking on `Start an application` delays Crime Application creation until after the user has selected an option.
3. As a result the `IsCifcForm` sits outside of the normal `steps` system, which has consequence for how the translation keys are generated
4. The CrimeApplication.application_type field remains immutable hence this delayed Crime Application creation
5. Added new `pre_cifc_xxxx` fields to store the values entered during the first 2 questions
6. Skips IOJ Passport screen if the application is a Change in Financial Circumstances
7. Prevents user from accessing the existing Appeal Financial Circumstances and Appeal Reference Number forms if they are accessed directly via URL
8. Text for the submission page, the tasklist etc requires further consideration
9. Updated shared specs for future re-use as well as in this PR
10. The updated `CrimeApplicationController#update` method deliberately leaves the code in-line rather than breaking up into methods to make it clearer what is happening. It will also require refactoring once the featureflag is removed.

### New button
Screenshot 2024-07-16 at 13 58 50

### First question
Screenshot 2024-07-16 at 22 14 55

### Second question
Screenshot 2024-07-16 at 22 15 05

### Task list
Screenshot 2024-07-16 at 19 17 53

### Submission
Screenshot 2024-07-16 at 19 17 46


app/services/decisions/case_decision_tree.rb
```diff
@@ -95,7 +95,7 @@ def after_hearing_details
     end
 
     def ioj_or_passported
-      if Passporting::IojPassporter.new(crime_application).call
+      if change_in_financial_circumstances? || Passporting::IojPassporter.new(crime_application).call
```
timpeat, line 98: Should we add CIFC as an IoJ passport?

app/services/decisions/case_decision_tree.rb
```diff
@@ -135,6 +135,12 @@ def blank_date_required?
     def crime_application
       form_object.crime_application
     end
+
+    def change_in_financial_circumstances?
```
timpeat, line 139: Can this be added to the TypeOfApplication module

app/services/decisions/case_decision_tree.rb
arthurashman: I think this is a good idea @timpeat - let's prioritise as a refactor, agree with @mseedat-moj that this could kick up a load of unknowns

app/models/partner.rb
```diff
@@ -11,6 +11,15 @@ class Partner < Person
     through: :crime_application
   )
 
+  # :nocov:
```
ripan, line 14: @mseedat-moj Does this change belong to this PR?

app/controllers/completed_applications_controller.rb
```diff
@@ -18,9 +18,10 @@ def show
 
   def recreate
     usn = current_crime_application.reference
+    application_type = current_crime_application['application_type']
```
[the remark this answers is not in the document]
hiboabd: Let's put a comment for now and address after this pr has gone in

hiboabd:
Added row to display maat id or usn for cifc apps
Screenshot 2024-07-19 at 15 58 37
Screenshot 2024-07-19 at 15 58 15

app/controllers/crime_applications_controller.rb
```diff
@@ -8,15 +8,38 @@ def index
...
+        new_application_params[:is_cifc]
+      )
+
+      return render(:new, flash:) unless @form_object.valid?
+
+      attrs[:application_type] = ApplicationType::CHANGE_IN_FINANCIAL_CIRCUMSTANCES if cifc?
+    end
+
+    initialize_crime_application(attrs) do |crime_application|
+      if FeatureFlags.cifc_journey.enabled? && cifc?
```
timpeat, line 38: @form_object.is_cifc rather than params?

app/controllers/crime_applications_controller.rb
```diff
@@ -41,7 +64,15 @@ def sorting_params
...
 
+  def new_application_params
+    params.fetch(:start_is_cifc_form, {}).permit(:is_cifc)
+  end
+
   def log_context
     LogContext.new(current_provider: current_provider, ip_address: request.remote_ip)
   end
+
+  def cifc?
```
timpeat, line 75: Can we use @form_object.is_cifc rather than params?

timpeat (approved):

app/controllers/crime_applications_controller.rb
hiboabd: @mseedat-moj perhaps something to refactor on monday

PR #1038, opened 2024-07-14
CRIMAPP-1209 - Skip 'manage with no income' for self employed

ripan:
## Description of change
Skip 'manage with no income' for self employed

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1209

### DIFF
--- a/app/services/decisions/income_decision_tree.rb
+++ b/app/services/decisions/income_decision_tree.rb
@@ -288,7 +288,7 @@ module Decisions
     def determine_showing_no_income_page
       if include_partner_in_means_assessment?
         edit(:partner_employment_status)
-      elsif crime_application.income&.all_income_over_zero?
+      elsif crime_application.income&.all_income_over_zero? || crime_application.income&.client_self_employed?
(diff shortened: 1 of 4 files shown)


app/services/decisions/income_decision_tree.rb
```diff
@@ -288,7 +288,7 @@ def edit_dependants(add_blank: false)
     def determine_showing_no_income_page
       if include_partner_in_means_assessment?
         edit(:partner_employment_status)
-      elsif crime_application.income&.all_income_over_zero?
+      elsif crime_application.income&.all_income_over_zero? || crime_application.income&.client_self_employed?
```
arthurashman, line 291: We probably want to check that the user has an income from their self employment, not just that they are self employed. Not certain -  I'm checking what eForms does now

arthurashman: Seems like they're not asked 👍

arthurashman (approved):
👍

PR #1039, opened 2024-07-15
CRIMAPP-1111 Only show needed capital answers.

timpeat:
## Description of change

Only show Capital asset answers when full capital required.

## Link to relevant ticket

[CRIMAPP-1111](https://dsdmoj.atlassian.net/browse/CRIMAPP-1111)

## Notes for reviewer

This follows a similar pattern to Income, where the Capital model is responsible for knowing when to include/omit answers.

There is a similar issue for evidence prompt which is part of a separate ticket: [CRIMAPP-1215](https://dsdmoj.atlassian.net/browse/CRIMAPP-1215)

[CRIMAPP-1111]: https://dsdmoj.atlassian.net/browse/CRIMAPP-1111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CRIMAPP-1215]: https://dsdmoj.atlassian.net/browse/CRIMAPP-1215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/capital.rb
+++ b/app/models/capital.rb
@@ -2,17 +2,13 @@ class Capital < ApplicationRecord
   include MeansOwnershipScope
 
   belongs_to :crime_application
+
   attribute :premium_bonds_total_value, :pence
   attribute :partner_premium_bonds_total_value, :pence
   attribute :trust_fund_amount_held, :pence
   attribute :trust_fund_yearly_dividend, :pence
   attribute :partner_trust_fund_amount_held, :pence
   attribute :partner_trust_fund_yearly_dividend, :pence
-
-  has_many :savings, through: :crime_application
-  has_many :investments, through: :crime_application
-  has_many :national_savings_certificates, through: :crime_application
-  has_many :properties, through: :crime_application
 
   validate on: :submission do
     answers_validator.validate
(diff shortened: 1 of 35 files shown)


app/models/capital.rb
```diff
@@ -27,6 +23,54 @@ def complete?
...
+  def savings
+    return [] unless requires_full_capital?
+
+    crime_application.savings.where(ownership_type: ownership_types)
+  end
+
+  def properties
+    return [] unless requires_full_capital?
+
+    crime_application.properties
```
hiboabd, line 35: does this require `.where(ownership_type: ownership_types)`

spec/support/shared_examples/a_capital_records_section.rb
```diff
@@ -0,0 +1,119 @@
+RSpec.shared_examples 'a capital records section' do |_options|
```
hiboabd, line 1: 🚀

hiboabd (approved):
LGTM

app/models/capital.rb
timpeat: We could investigate the percentage owned values to determine if we should include properties not owned by the applicant. For now, we rely on the applicant to remove any entries added when a partner was included in the means.

PR #838, opened 2024-05-08
Fix task list for returned applications

timpeat:
## Description of change
Returned applications can become passported.
We consider will_enter_nino complete if we already know applicant in custody.

## Link to relevant ticket
[CRIMAPP-841](https://dsdmoj.atlassian.net/browse/CRIMAPP-841)
[CRIMAPP-820](https://dsdmoj.atlassian.net/browse/CRIMAPP-820)

## How to manually test the feature

See ticket

[CRIMAPP-841]: https://dsdmoj.atlassian.net/browse/CRIMAPP-841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CRIMAPP-820]: https://dsdmoj.atlassian.net/browse/CRIMAPP-820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1042, opened 2024-07-15
[CRIMAPP-1195] Recalculate means passport value for resubmitted applications

hiboabd:
## Description of change
Refactors the means passporter to always calculate an apps passport status regardless of whether the app is resubmitted (with the exception of under 18 applications which need to retain the previous status). This is to prevent inaccuracies in presenting the passporting status when an application is resubmitted and the means passport value has not been updated on resubmission

## Link to relevant ticket
[CRIMAPP-1195](https://dsdmoj.atlassian.net/browse/CRIMAPP-1195)

## Notes for reviewer
The only means passport value that needs to be retained on resubmission is the under 18 status. I did take a look at the benefit check passport but that is recalculated ([see pr](https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/pull/838)) 

Requires [datastore](https://github.com/ministryofjustice/laa-criminal-applications-datastore/pull/291) and [review](https://github.com/ministryofjustice/laa-review-criminal-legal-aid/pull/649) pr's to test locally 

[CRIMAPP-1195]: https://dsdmoj.atlassian.net/browse/CRIMAPP-1195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/passporting/means_passporter.rb
+++ b/app/services/passporting/means_passporter.rb
@@ -3,11 +3,9 @@ module Passporting
     include TypeOfMeansAssessment
 
     def call
-      return passported? if resubmission?
-
       means_passport = []
       means_passport << MeansPassportType::ON_NOT_MEANS_TESTED if app_not_means_tested?
-      means_passport << MeansPassportType::ON_AGE_UNDER18      if applicant_under18?
+      means_passport << MeansPassportType::ON_AGE_UNDER18      if age_passported?
       means_passport << MeansPassportType::ON_BENEFIT_CHECK    if benefit_check_passed?
 
       crime_application.update(means_passport:)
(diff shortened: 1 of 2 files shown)


timpeat (changes requested):
Great change. We will need to update the method used to determine `applicant_under_18?` from `applicant.under18` to` !applicant.over_18_at_date_stamp?` before this can be merged, as its is important that an applicant U18 at time of submission is still considered U18 on return even if they are now over 18.

timpeat:
> Great change. We will need to update the method used to determine `applicant_under_18?` from `applicant.under18` to` !applicant.over_18_at_date_stamp?` before this can be merged, as its is important that an applicant U18 at time of submission is still considered U18 on return even if they are now over 18.

@hiboabd has pointed out that the age_passported? method has a guard to exclude resubmitted applications. Agree that this should cary on using the u18 method as this will maintain existing behaviour.

timpeat (approved):

PR #1046, opened 2024-07-16
Evidence upload task list behaviour amends

hiboabd:
## Description of change

Fixes behaviour with the evidence upload task list following work to validate evidence upload 
Previously the evidence upload task will display as complete immediately as there are no evidence prompts yet 
However, we don't really know if an application's evidence is complete, only whether they have uploaded at least one piece of evidence. It may also give unrealistic view to providers that they don't have to upload any more evidence. So it was decided it is best to keep the status as in progress rather than ever setting the task to complete

The evidence form, review application page and declaration page will ensure apps are only submitted if they have fulfilled the validation criteria 

Implemented behaviour:

- Evidence upload status will = not started if there is no evidence uploaded
- Evidence upload status will = in progress if at least one piece of evidence has been uploaded 
- Evidence upload status will never be set to completed as we do not know if evidence is complete (so arguably more accurate)

There was also a bug when a user clicks the evidence upload link immediately after starting an application as there is no applicant yet and the evidence prompts are expecting a person to be available to assess whether evidence is required. So the evidence upload section cannot be started if an applicant is not present 

### Before changes:
**Currently evidence upload section is always displayed as completed**
Screenshot 2024-07-16 at 13 14 29

### After changes:

**When there is no evidence uploaded**
Screenshot 2024-07-16 at 13 12 39

**When there is evidence uploaded**
Screenshot 2024-07-16 at 13 12 56

### DIFF
--- a/app/presenters/tasks/evidence_upload.rb
+++ b/app/presenters/tasks/evidence_upload.rb
@@ -5,11 +5,15 @@ module Tasks
     end
 
     def can_start?
-      true
+      crime_application.applicant.present?
     end
 
     def in_progress?
-      true
+      crime_application.documents.any?
+    end
+
+    def completed?
+      false
     end
 
     private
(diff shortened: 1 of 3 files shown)


timpeat (approved):

arthurashman (approved):
Nice, thanks @hiboabd !

PR #1053, opened 2024-07-17
CRIMAPP-1234 Ensure no partner declaration reason is persisted

arthurashman:
## Description of change
Ensure no partner declaration reason is persisted
## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1234

### DIFF
--- a/app/serializers/submission_serializer/sections/provider_details.rb
+++ b/app/serializers/submission_serializer/sections/provider_details.rb
@@ -6,7 +6,7 @@ module SubmissionSerializer
           json.provider_details do
             json.office_code crime_application.office_code
             json.provider_email crime_application.provider_email
-            json.legal_rep_has_partner_declaration rep_has_partner_declaration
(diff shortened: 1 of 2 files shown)


app/serializers/submission_serializer/sections/provider_details.rb
```diff
@@ -16,14 +16,6 @@ def to_builder # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
           end
         end
       end
-
-      private
-
-      def rep_has_partner_declaration
-        return unless crime_application.legal_rep_has_partner_declaration
-
-        crime_application.legal_rep_has_partner_declaration['value']
```
arthurashman, line 25: @timpeat Can you remember if I had a good argument for having done this? It seems to work for me without now and I've worried myself that I can't find the journey that required it

timpeat: I wonder if it was because the value object attribute was previously not set correctly? Looks like it is now so can go?

timpeat (approved):

PR #1069, opened 2024-07-22
CRIMAPP-1235 Add validation for national savings certificate format

RoseSAK:
## Description of change

Add validation to prevent user entering strings that end with '-' to avoid modsec errors. It is still possible to enter strings that contain '-' just not those ending in dashes. 

## Link to relevant ticket

[CRIMAPP-1235](https://dsdmoj.atlassian.net/browse/CRIMAPP-1235)

Screenshot 2024-07-22 at 12 46 24

[CRIMAPP-1235]: https://dsdmoj.atlassian.net/browse/CRIMAPP-1235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/forms/steps/capital/national_savings_certificates_form.rb
+++ b/app/forms/steps/capital/national_savings_certificates_form.rb
@@ -5,13 +5,18 @@ module Steps
       include OwnershipConfirmation
       include ApplicantOrPartner
 
+      # match any alphanumeric string that does not contain any special characters
+      # this is to avoid modsec errors, specifically strings ending with - (dash)
+      NSC_REGEXP = /\A[\w\s]+\z/
+
       delegate :national_savings_certificate_type, to: :record
 
       attribute :holder_number, :string
       attribute :certificate_number, :string
       attribute :value, :pence
 
-      validates :holder_number, :certificate_number, :value, presence: true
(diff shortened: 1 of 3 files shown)


app/forms/steps/capital/national_savings_certificates_form.rb
```diff
@@ -5,13 +5,17 @@ class NationalSavingsCertificatesForm < Steps::BaseFormObject
       include OwnershipConfirmation
       include ApplicantOrPartner
 
+      # matches any alphanumeric string that does not end with '-' to avoid modsec errors
+      NSC_REGEXP = /.*(?<!-)\z/
```
ripan, line 9: @RoseSAK Should we only invalidate '-' or any other special character?. 
If we know format of 'holder_number'  then we can validate according to the format. May be allow only characters and numbers only.

RoseSAK: Hmm yes good point, I will check with Chris what the format would likely be

ripan (approved):

PR #1077, opened 2024-07-24
CRIMAPP-1105: Show assets section incomplete if owners information is missing

ripan:
## Description of change
Fix definition of 'incomplete' for property

- check presence of owners details if property has owners
- Sum of ownership percentage should be 0 (Applicant + Partner(if present) + Owners(if present) )

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1105

### Before changes:
b7b21516-896e-465d-b5dd-d479c5aa4b05

### After changes:
Screenshot 2024-07-25 at 09 55 04

### DIFF
--- a/app/models/property.rb
+++ b/app/models/property.rb
@@ -65,6 +65,10 @@ class Property < ApplicationRecord
   end
 
   def property_owners_complete?
-    property_owners.all?(&:complete?)
+    return true if has_other_owners == YesNoAnswer::NO.to_s || has_other_owners.nil?
+
+    (has_other_owners == YesNoAnswer::YES.to_s) &&
+      property_owners.present? &&
+      property_owners.all?(&:complete?)
   end
 end
--- a/app/models/property.rb
+++ b/app/models/property.rb
@@ -53,7 +53,7 @@ class Property < ApplicationRecord
     return false unless property_type
 
     values_at(*REQUIRED_ATTRIBUTES[property_type.to_sym]).all?(&:present?) &&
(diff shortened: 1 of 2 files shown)


app/models/property.rb
```diff
@@ -65,6 +65,19 @@ def address_complete?
   end
 
   def property_owners_complete?
-    property_owners.all?(&:complete?)
+    return true if has_other_owners == YesNoAnswer::NO.to_s || has_other_owners.nil?
```
hiboabd, line 68: Is this guard needed? feels like line 70 is sufficient?

app/models/property.rb
```diff
@@ -65,6 +65,19 @@ def address_complete?
...
+
+    (has_other_owners == YesNoAnswer::YES.to_s) &&
+      property_owners.present? &&
+      property_owners.all?(&:complete?)
+  end
+
+  def property_ownership_valid?
```
hiboabd, line 75: I think this logic is used in the property owners validator, would be good to see if there is a way to avoid duplicating the logic but don't worry if it's not straightforward

ripan: I tried but its not a straightforward thing. 😢

app/models/property.rb
ripan: Just trying to save adding query when not required :)

app/models/property.rb
hiboabd: I thought so  😢

hiboabd (approved):
👍

PR #1087, opened 2024-07-26
[CRIMAPP-1251] Office code removal request

hiboabd:
## Description of change
Removes an office code as per requested. I also had to check if a couple of office codes were active and all were with the exception of one office code which was in both lists for which i have left a comment below

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1251


config/inactive_offices.yml
```diff
@@ -238,7 +238,6 @@ production:
     - 2M241U
     - 2M247A
     - 2P619D
-    - 2Q248M
```
hiboabd, line 241: The ticket asks to check if this office code is active. It was in both the active and inactive office code list and was in both lists when the office codes [were being updated](https://dsdmoj.atlassian.net/browse/CRIMAPP-244). My assumption now is that we want it to be active based on the current ticket, so i have removed from the inactive list but we can also wait for confirmation from fleur

arthurashman: Agree, that's my understanding from the ticket too

PR #1090, opened 2024-07-30
[CRIMAPP-1243] Prevent saving attributes as empty string

hiboabd:
## Description of change
Fixes a bug where when users clicked save and come back later on a form (typically a radio button form), an empty string was saved which then threw an error when submitting the application as the value does not conform to the schema. 

This has been fixed by adding config to the form builder to not include hidden input fields that have been populated with a blank value when rendering radio buttons. 

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1243

## Notes for reviewer
Note this won't fix any existing applications that have the bug, it will just prevent new draft applications from getting into this state 

Start an application and navigate to a form with radio buttons e.g. case type. Select save and come back later and verify that the attribute is still `nil` and not an empty string. You may also want to check the validation is still working as expected.

### DIFF
--- a/app/forms/steps/capital/other_investment_type_form.rb
+++ b/app/forms/steps/capital/other_investment_type_form.rb
@@ -16,8 +16,6 @@ module Steps
       private
 
       def persist!
-        return true if investment_type == ''
-
         @investment = crime_application.investments.create!(investment_type:)
       end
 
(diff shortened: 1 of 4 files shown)


arthurashman (approved):
Works for me 🙌  Thank you!

PR #1098, opened 2024-08-02
[CRIMAPP-1096] Property ownership must be less than 100% if there are other owners

hiboabd:
## Description of change
When validating percentage ownership amounts for a property with other owners, the percentage(s) entered are no longer allowed to equal 100% as there are more ownership details to be entered. 

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1096

### After changes:
Screenshot 2024-08-02 at 11 42 52

Create a property (with or without a partner) and enter a total % ownership that is equal to 100%. Then select yes for other owners and click save and continue. You should see the error message attached above. Then click no for other owners and you should be able to progress. Test for all property types (residential, land and commercial)


ripan (approved):
LGTM 👍

PR #1099, opened 2024-08-02
Check for presence of outgoings before calculating evidence triggers

hiboabd:
## Description of change
Fixes bug where the evidence upload page breaks when there are no outgoings because the presence of outgoings is not being checked properly

Start an application and go through the income assessment section answering no to every question. You should be taken to the evidence upload page rather than the outgoings flow. Confirm the evidence upload page loads rather than breaking


hiboabd:
Was fixed in: https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/commit/1e3463787829010718e710ad898582d77f03c3ef

PR #1109, opened 2024-08-07
[CRIMAPP-936] Remove legacy client_has_partner fields and update forms

mseedat-moj:
## Description of change
Removes the `client_has_partner` field and refactors corresponding code

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-936

## Notes for reviewer
- Note schema change allows rollback completely

N/A

### DIFF
--- a/app/controllers/steps/circumstances/pre_cifc_reason_controller.rb
+++ b/app/controllers/steps/circumstances/pre_cifc_reason_controller.rb
@@ -1,0 +1,15 @@
+module Steps
+  module Circumstances
+    class PreCifcReasonController < Steps::CircumstancesStepController
+      def edit
+        @form_object = PreCifcReasonForm.build(
+          current_crime_application
+        )
+      end
+
+      def update
+        update_and_advance(PreCifcReasonForm, as: :pre_cifc_reason)
+      end
+    end
+  end
+end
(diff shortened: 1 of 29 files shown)


app/forms/steps/client/has_partner_form.rb
```diff
@@ -1,49 +1,46 @@
...
 
         ::CrimeApplication.transaction do
-          reset!
-
-          crime_application.update!(
-            attributes
-          )
+          if has_partner.no?
+            reset!
+          else
```
hiboabd, line 27: Do we want to explicit here and check that has_partner is yes as someone can click save and come back

hiboabd (commented):
The partner screen is automatically populated with no when i'm going through an application 🤔
Screenshot 2024-08-08 at 16 39 07

app/forms/steps/client/has_partner_form.rb
hiboabd: @mseedat-moj i think it's better to be explicit here

hiboabd: Ah never mind, line 22 covers it

hiboabd (approved):

PR #1118, opened 2024-08-13
CRIMAPP-1181 Add ARC number to NINO page

RoseSAK:
## Description of change

Add a third radio option to the `has_nino` page which allows the provider to enter an ARC number for the client or partner
Also updates the routing to skip the passporting benefit section if the client or partner has an arc. The validation and task list logic is also updated to accompany this change
For now, will be hidden behind a feature flag to enable QA testing without blocking deployments

## Link to relevant ticket

[CRIMAPP-1181](https://dsdmoj.atlassian.net/browse/CRIMAPP-1181)

### Before changes:
Screenshot 2024-08-13 at 14 59 30

### After changes:
Screenshot 2024-08-13 at 14 58 58

Screenshot 2024-08-13 at 15 00 13
Screenshot 2024-08-13 at 15 00 03

Requires the datastore and review PRs to test locally

[CRIMAPP-1181]: https://dsdmoj.atlassian.net/browse/CRIMAPP-1181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/presenters/summary/sections/passporting_benefit_check_partner.rb
```diff
@@ -2,7 +2,7 @@ module Summary
   module Sections
     class PassportingBenefitCheckPartner < Sections::PassportingBenefitCheck
       def show?
-        include_partner_in_means_assessment?
+        super && include_partner_in_means_assessment?
```
hiboabd, line 5: As this is now an optional question (won't be asked if the partner has an arc), this needs to check if the partner has a benefit type selected

spec/forms/steps/client/nino_form_spec.rb
```diff
@@ -214,6 +217,61 @@
             end
           end
         end
+
+        context 'when `has_nino` answer is no but they have an arc number' do
+          let(:has_nino) { HasNinoType::ARC.to_s }
+          let(:arc) { 'ABC12/345678/A' }
```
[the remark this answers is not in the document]
hiboabd: This is correct, `has_nino` is `arc` if there is an arc number

app/presenters/summary/sections/passporting_benefit_check_partner.rb
hiboabd: Ignore the above, the section is meant to be displayed whether there is a benefit type or not (see designs). Instead I have amended the logic to not display the section if the partner has an arc

app/value_objects/has_nino_type.rb
```diff
@@ -0,0 +1,7 @@
+class HasNinoType < ValueObject
```
timpeat, line 1: I don't think `HasNino` should be converted into a type. It should only have the values 'yes', 'no', and 'nil'. Adding 'arc' as a value for `has_nino` makes sense only from the perspective of the form, not our underlying data model. Would it be possible to add both `has_nino` and `has_arc` if we need to store these values for rehydration? I realise it's a bit more work to update both values from the form, but I'm concerned that combining them into a single value just because they appear on the same form ties the data model too closely to a design that might change.

timpeat (changes requested):

timpeat (approved):

PR #1137, opened 2024-08-22
[CRIMAPP-1257] Cifc is non means tested journey update

mseedat-moj:
## Description of change
Skips 'is this application means tested' for cifc application

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1257

## Notes for reviewer
- note the serializer hard-codes the answer for `is_means_tested` to 'yes' for cifc applications rather than ommitting it.
- removes `non_menas_tested` feature flag


config/locales/en/errors.yml
```diff
@@ -86,6 +86,7 @@ en:
           attributes:
             is_means_tested:
               inclusion: Select yes if this application is means tested
+              cifc: Change in Financial Circumstances application must be means tested
```
arthurashman, line 89: Has this been provided by content? Or otherwise checked/approved?

arthurashman (approved):
Unlikely that a user would see the error message so approving, but just a Q if that copy has been provided/checked

PR #1138, opened 2024-08-22
Add system spec for completed age passported application

timpeat:
## Description of change

Add system spec for submitting an age passported application

## Notes for reviewer

This update adds a system spec for the U18 passported application as part of our effort to expand system specs to cover all general means and application types. Initially, the focus will be on confirming the happy path.

This spec takes the application from the initial creation up to datastore submission. To achieve this, a schema-compliant application must be generated (covering steps such as submission and review), and the fulfilment validation must be satisfied. This process, therefore, executes the serialisation and deserialisation of a datastore application for this scenario.

Next up will be the non-means-tested application.

### DIFF
--- a/spec/system/applying/passported_on_age_spec.rb
+++ b/spec/system/applying/passported_on_age_spec.rb
@@ -1,0 +1,105 @@
+require 'rails_helper'
+
+RSpec.describe 'Apply for Criminal Leagal Aid when under 18 years old' do
+  describe 'Submitting a completed application' do
+    before do
+      visit root_path
+      click_button('Start now')
+
+      # steps/provider/select_office
+      choose('2A555X')
+      # prevent requests to the datastore for counters for tab headings on the next page
+      allow_any_instance_of(Datastore::ApplicationCounters).to receive_messages(returned_count: 0)
+      save_and_continue
+
+      # applications
+      click_link('Start an application')
+      choose('New application')
+      save_and_continue
+
+      # edit (Task list)
+      click_link('Client details')
+
+      # steps/client/details
+      fill_in('First name', with: 'Jo')
+      fill_in('Last name', with: 'Bloggs')
+      fill_date('Date of birth', with: 17.years.ago.to_date)
+      save_and_continue
+
+      # steps/client/is_application_means_tested
+      save_and_continue
+
+      # steps/client/case_type
+      choose('Indictable')
+      save_and_continue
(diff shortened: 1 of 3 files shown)


hiboabd (commented):
LGTM and looks like it was simple to set up too. Only question from me is whether this will be part of the build pipeline? Also would this eliminate the need for the selenium repo? And do we need to add something to the README?

And general thoughts is that it would be good to agree on what journeys we want the system specs to cover (is it all potential journeys?).

spec/system/applying/passported_on_age_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+      # steps/client/contact_details
+      choose('Provider’s office')
+      save_and_continue
+
+      # steps/client/urn
+      save_and_continue
+
+      # steps/case/has_the_case_concluded
+      choose_answer('Has the case concluded?', 'No')
+      save_and_continue
```
hiboabd, line 49: Would be good to add a shared context or example to reuse some of the parts of the form that would be repeated in a lot of specs (e.g. case details)

spec/system/applying/passported_on_age_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+      select('Derby Crown Court', from: 'Court name')
+      fill_date('Date of next hearing', with: 1.week.from_now.to_date)
+      choose_answer('Did this court also hear the first hearing?', 'Yes')
+      save_and_continue
+
+      # steps/case/ioj_passport
+      save_and_continue
+
+      # steps/evidence/upload
+      save_and_continue
```
hiboabd, line 78: Will the validation on this screen be a problem when uploading evidence is mandatory for an application?

spec/system/applying/passported_on_age_spec.rb
timpeat: Potentially, I'd rather extract those patterns when we get to that point.

timpeat:
> LGTM and looks like it was simple to set up too. Only question from me is whether this will be part of the build pipeline? Also would this eliminate the need for the selenium repo? And do we need to add something to the README?
> 
> And general thoughts is that it would be good to agree on what journeys we want the system specs to cover (is it all potential journeys?).

I think we need all these things, but I'd rather, add a couple of other variants before we start documenting / refactoring.

PR #1153, opened 2024-09-02
[CRIMAPP-835] Add IoJ answer validator

timpeat:
## Description of change

Validate Interests of Justice answers on Submission Review

## Link to relevant ticket

https://dsdmoj.atlassian.net/browse/CRIMAPP-835

## Notes for reviewer

Adds an answer validator for the IoJ task and adds it to the section completeness validator.

Confirm that an error message is shown when IoJ are required but have not been given (or have only been partially given). See ticket for details.

### DIFF
--- a/app/presenters/tasks/ioj.rb
+++ b/app/presenters/tasks/ioj.rb
@@ -8,29 +8,18 @@ module Tasks
       end
     end
 
-    def not_applicable?
-      return true if crime_application.cifc?
-
-      false
-    end
-
     def can_start?
       fulfilled?(CaseDetails)
     end
 
     def in_progress?
-      crime_application.ioj_passport.any? || ioj.present?
+      ioj.present?
     end
 
-    def completed?
-      return true if crime_application.ioj_passported?
+    private
 
-      ioj.present? && ioj.types.any? && ioj_attributes_present?
-    end
-
-    def ioj_attributes_present?
-      required_attributes = ioj.types.map { |type| "#{type}_justification" }
-      ioj.values_at(*required_attributes).all?(&:present?)
+    def validator
+      @validator ||= InterestsOfJustice::AnswersValidator.new(crime_application)
     end
   end
 end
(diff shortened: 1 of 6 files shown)


app/validators/interests_of_justice/answers_validator.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    def validate
+      return if complete?
+
+      errors.add(:ioj, :incomplete)
+      errors.add(:base, :incomplete_records)
+    end
+
+    def complete?
+      crime_application.ioj_passported? || ioj_answers_complete?
```
arthurashman, line 23: Wondering whether `passported` is more of a not-applicable scenario - could it confuse a user to update to `complete` when they did not complete anything?

timpeat: Ideally, we want the task list's 'non-applicable' and the answer validator's 'applicable?' to be inverses of each other. For this to happen, 'applicable?' in the validator should function as the inverse of 'not_applicable?'.

For example, in a "change in financial circumstances" application, the 'Interests of Justice' section is not applicable. However, for an under-18 applicant, the IoJ is applicable to the application but passported, so it is automatically completed.

I think we're at the point where we can start defining  what 'applicable?' means in the context of an answer validator, and we now have enough of a pattern to consider a refactor and see if we can standardise each interface and clearly define their roles. Does that make sense?

app/presenters/tasks/ioj.rb
```diff
@@ -8,29 +8,18 @@ def path
...
     def can_start?
       fulfilled?(CaseDetails)
     end
 
     def in_progress?
-      crime_application.ioj_passport.any? || ioj.present?
+      ioj.present?
     end
 
-    def completed?
```
arthurashman, line 25: I think removing this has meant that the tasklist badge doesn't show complete anymore

app/validators/interests_of_justice/answers_validator.rb
arthurashman: Yep, that makes sense.

PR #1157, opened 2024-09-03
Fix ARC validation (bug from QA)

RoseSAK:
## Description of change

Allow lower case letters and upcase on save


hiboabd (approved):
Thanks!

PR #1158, opened 2024-09-03
CRIMAPP-801 Prevent setting more than one home address

hiboabd:
## Description of change
Adds validation to property form to prevent a property being declared as the client's home address if another property has already been declared as the client's home address

This is required for MAAT as MAAT will only display 1 of the assets that are declared as the home address, each subsequent asset declared as the home address will not display in MAAT. 

## Link to relevant ticket

### After changes:

Screenshot 2024-09-03 at 13 45 39

### DIFF
--- a/app/services/evidence/rules/20240425120827_council_tax_payments.rb
+++ b/app/services/evidence/rules/20240425120827_council_tax_payments.rb
@@ -15,7 +15,6 @@ module Evidence
         payment.present? && payment.prorated_monthly.to_f > THRESHOLD
       end
 
-      # TODO: Awaiting partner implementation
       partner do |_crime_application|
         false
       end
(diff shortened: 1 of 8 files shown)


timpeat (approved):

app/services/evidence/rules/20240425120827_council_tax_payments.rb
```diff
@@ -15,7 +15,6 @@ class CouncilTaxPayments < Rule
         payment.present? && payment.prorated_monthly.to_f > THRESHOLD
       end
 
-      # TODO: Awaiting partner implementation
```
arthurashman, line 18: I think we need to update the rules though? I'll write a ticket. But just for ref when deleting outdated comments, make sure the action has been taken (which I don't think it has 😬 )

hiboabd: It's a joint client and partner question so I don't think there is any further action

PR #1161, opened 2024-09-05
[CRIMAPP-1289] Add explicit 'not provided' for house_type

mseedat-moj:
## Description of change
Forces blank string `house_type` field to be output in the Capital Assesment summary table with 'not provided'.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1289
## Notes for reviewer
- This bug just does not manifest in localhost but does so in staging and presumably production. The change makes the 'absence' answer explicit. 
- Also updates the 'Bedrooms' field to show 'Not provided' rather than 'None' because it makes more sense to do so.
### Before changes:
If `house_type` is saved with a blank string (``) then localhost will throw an exception:
Screenshot 2024-09-05 at 13 53 45

But if the `house_type` is `nil` then the word 'None' is output:
Screenshot 2024-09-05 at 13 54 44

### After changes:
Screenshot 2024-09-05 at 13 49 39

Follow instructions to reproduce in ticket

### DIFF
--- a/app/presenters/summary/components/property.rb
+++ b/app/presenters/summary/components/property.rb
@@ -136,7 +136,7 @@ module Summary
         if property.house_type == ::Property::OTHER_HOUSE_TYPE
           property.other_house_type
         else
-          I18n.t(property.house_type, scope: [:summary, :sections, :property, :house_type])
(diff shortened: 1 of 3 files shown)


arthurashman (approved):
Let's do it!

PR #1163, opened 2024-09-09
[LAA-APPLY-FOR-CRIMINAL-LEGAL-AID-3X] fix validation issues

EdwinKruglov:
## Description of change
Ensure that numericality validation messages on the Income Payment form have all the necessary information to construct an error message.

## Link to relevant ticket
[LAA-APPLY-FOR-CRIMINAL-LEGAL-AID-3X](https://ministryofjustice.sentry.io/issues/5819702006/?alert_rule_id=15141078&alert_type=issue&notification_uuid=7c31a4d3-5779-4928-a581-03462cdf34a4&project=4507142444023808)

## Notes for reviewer
I've implemented a custom numericality validation method in order to have more control over the error messaging, i.e. to be able to provide the `payment_type` value to the error strings. The need for this comes from content changes that require more dynamic field-level error messaging - e.g. "Maintenance payments amount must be greater than 0" instead of "Amount must be greater than 0". Previously, these "fuller" error messages were only shown in the error summary box.

Initially I wanted to extend the default `NumericalityValidator` which does all the validation we could want, but I couldn't find a good way to extend it just so I could alter the error messaging.

What's also becoming clear is that we end up constructing error messages both at the attribute level and at the summary level when in most (all?) cases the summary error messages will be the same, just repeated above the form. Unless there are cases where summary errors need to differ from field-level errors, this seems like duplicated effort.

### DIFF
--- a/app/forms/concerns/steps/field_validation.rb
+++ b/app/forms/concerns/steps/field_validation.rb
@@ -1,0 +1,34 @@
+module Steps
+  module FieldValidation
+    extend ActiveSupport::Concern
+
+    # rubocop:disable Metrics/MethodLength
+    def numericality(attribute, options = {})
+      errors = []
+      value =
+        begin
+          Kernel.Float(send(attribute))
+        rescue ArgumentError, TypeError
+          nil
+        end
+
+      return errors << :not_a_number if value.nil?
+
+      options.each do |option, arg|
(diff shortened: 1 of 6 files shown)


hiboabd (commented):
I wonder if this is something we need to push back on if it’s causing bloating and duplication? Not sure if there have been conversations already though..

PR #1164, opened 2024-09-09
[CRIMAPP-801] Remove save and come back later button from property form

hiboabd:
## Description of change
Removes the save and come back later button from the property form pages (residential, land, commercial) in order to stop users from saving more than one property as the client's home address. This was addressed in a previous pr but during qa, it was discovered that this could still occur via the save and come back later button as the save and come back later functionality currently skips validations so the error is not picked up. 

I’ve looked into displaying an error on the assets summary page (using the incomplete tag) and also displaying the error on the review application page (`You need to complete all questions before you can submit the application`). However, ultimately, this proved difficult as the validations in the property model require global knowledge of all the properties and also leads to duplication of the validation logic 

I also discussed the following options: 
1. Potentially introducing a new attribute `home_address_id` where the id would be replaced when the form is saved so there will only ever be one home address
2. Hiding the home address question if the home address has been set 

But ultimately decided it is better if the home address question is explicitly answered by the user and the user is easily given the option to change their answer if they have previously made an error 

So in the end I opted to remove the save and back later button from these forms.  

The only potential issue I forsee with this is users thinking this is a bug but happy to discuss if there are concerns

## Link to relevant ticket

### Before changes:
Screenshot 2024-09-09 at 18 21 34

### After changes:
Screenshot 2024-09-09 at 18 20 53

### DIFF
--- a/app/forms/steps/capital/property_form.rb
+++ b/app/forms/steps/capital/property_form.rb
@@ -56,8 +56,7 @@ module Steps
       def home_address
         return true unless is_home_address&.yes?
 
-        errors.add(:is_home_address, :invalid) if is_home_address&.yes? &&
-                                                  crime_application.properties.home_address.count >= 1
+        errors.add(:is_home_address, :invalid) if crime_application.properties.home_address.count >= 1
       end
     end
   end
--- a/app/views/steps/capital/commercial_property/edit.html.erb
+++ b/app/views/steps/capital/commercial_property/edit.html.erb
@@ -24,7 +24,7 @@
       <% end %>
       <%= f.govuk_collection_radio_buttons :has_other_owners, YesNoAnswer.values, :value, inline: true, legend: { size: 's' } %>
 
-      <%= f.continue_button %>
+      <%= f.continue_button(secondary: false)  %>
     <% end %>
   </div>
 </div>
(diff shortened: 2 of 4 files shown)


app/views/steps/capital/commercial_property/edit.html.erb
```diff
@@ -24,7 +24,7 @@
       <% end %>
       <%= f.govuk_collection_radio_buttons :has_other_owners, YesNoAnswer.values, :value, inline: true, legend: { size: 's' } %>
 
-      <%= f.continue_button %>
+      <%= f.continue_button(secondary: false)  %>
```
hiboabd, line 27: I will also add a comment explaining why we have made the change to the form if we agree to go ahead with removing the button

app/forms/steps/capital/property_form.rb
```diff
@@ -56,8 +56,7 @@ def property_is_home_address?
       def home_address
         return true unless is_home_address&.yes?
 
-        errors.add(:is_home_address, :invalid) if is_home_address&.yes? &&
-                                                  crime_application.properties.home_address.count >= 1
+        errors.add(:is_home_address, :invalid) if crime_application.properties.home_address.count >= 1
```
hiboabd, line 59: This was just to remove the duplication of the logic from line 57

timpeat (approved):
Good decision.

hiboabd:
Closed as design have opted to hide the question

PR #1165, opened 2024-09-10
Using ping for liveness probe and add startup probe

timpeat:
## Description of change

- Use ping for livenessProbe on production deployment
- Add startup probe to production deployment

## Notes for reviewer
Applies changes tested on staging to production from:
https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/pull/1162

Using the healthcheck for the liveness probe was causing slow database connection to trigger restarts and pod crash looping. This change is for staging only and tests using ping instead of health for liveness. It also adds a startup probe. (e.g. https://mojdt.slack.com/archives/C056U956ESH/p1725882930025509 )

More discussion here: https://dsdmoj.atlassian.net/browse/CRIMAPP-345?focusedCommentId=499712

### DIFF
--- a/config/kubernetes/production/deployment.tpl
+++ b/config/kubernetes/production/deployment.tpl
@@ -47,14 +47,25 @@ spec:
           periodSeconds: 10
         livenessProbe:
           httpGet:
-            path: /health
+            path: /ping
             port: 3000
             httpHeaders:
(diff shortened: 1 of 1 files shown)


hiboabd (approved):
LGTM

PR #1175, opened 2024-09-12
Add system specs for partner means tested flows

hiboabd:
## Description of change
Adds system specs for three means tested routes with a partner (employed, not working, self employed)
For the time being, it verifies whether the submission is successful with some basic expected application information tested. We can look to test expectations in a subsequent pr like whether partner information is deleted when the partner answer has been updated

Also fixes a couple of small content/ux issues I noticed along the way

## Link to relevant ticket
No relevant ticket yet will write one up! 

## Notes for reviewer
One thing to note is each test takes a couple of seconds (~3) which will add up in time. We should look into what options we have to ensure this doesn't slow down the pipeline too much as we add more specs


spec/support/capybara_helpers.rb
```diff
@@ -22,6 +22,14 @@ def choose_answer(question, choice)
     end
   end
 
+  def choose_answer_all(question, choice)
```
hiboabd, line 25: I added this method to select all instances on form (where there are duplicates) and set a value. Alternatively, if we want to be specific we would need to look into adding ids to form elements so we can target specific elements

spec/support/shared_contexts/with_partner.rb
```diff
@@ -0,0 +1,125 @@
+RSpec.shared_context 'with partner' do
```
hiboabd, line 1: I should rename this to means tested with partner

app/views/steps/income/client/other_work_benefits/edit.html.erb
```diff
@@ -1,22 +1,23 @@
...
+        <p class="govuk-body">
+          <%= t('.page_text') %>
+        </p>
+        <ul class="govuk-body govuk-list--bullet govuk-!-margin-bottom-6">
+          <% t('.list').each_line do |line| %>
+            <li><%= line %></li>
+          <% end %>
+        </ul>
+
         <%= f.govuk_radio_button :applicant_other_work_benefit_received, YesNoAnswer::YES do %>
```
hiboabd, line 21: This was amended to work with the current spec helpers where the radio button is nested in the same element as the label. It has no front end impact

PR #1176, opened 2024-09-13
[CRIMAPP-1343] Unexpected 'no income' step

EdwinKruglov:
## Description of change
A regression has been introduced in the Income Assessment section of the application where the 'How do your client and their partner manage with no income?' question is asked when non-zero employment income has been provided.

## Link to relevant ticket
[CRIMAPP-1343](https://dsdmoj.atlassian.net/browse/CRIMAPP-1343)

## Notes for reviewer
Routing to the 'How do your client and their partner manage with no income?' question has been updated with client and partner self-employed checks where it was previously inconsistent.

The`manage_without_income_complete?` validation method has also been updated to check if the partner is self-employed. 

[CRIMAPP-1343]: https://dsdmoj.atlassian.net/browse/CRIMAPP-1343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/decisions/income_decision_tree.rb
+++ b/app/services/decisions/income_decision_tree.rb
@@ -59,7 +59,7 @@ module Decisions
       when :delete_dependant
         edit_dependants
       when :dependants_finished
-        determine_showing_no_income_page
+        after_dependants_finished
       when :manage_without_income
         edit(:answers)
       when :partner_employment_status
@@ -162,11 +162,7 @@ module Decisions
     end
 
     def after_partner_income_benefits
-      if crime_application.income&.all_income_over_zero?
-        edit(:answers)
-      else
-        edit(:manage_without_income)
-      end
+      determine_showing_no_income_page
     end
 
     def start_client_employment_journey # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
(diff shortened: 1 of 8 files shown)


app/services/decisions/income_decision_tree.rb
```diff
@@ -317,10 +317,14 @@ def edit_dependants(add_blank: false)
...
+      return edit(:partner_employment_status) if include_partner_in_means_assessment?
+
+      determine_showing_no_income_page
+    end
+
     def determine_showing_no_income_page
-      if include_partner_in_means_assessment?
-        edit(:partner_employment_status)
-      elsif crime_application.income&.all_income_over_zero? || crime_application.income&.client_self_employed?
+      if income&.all_income_over_zero? || income&.client_self_employed? || income&.partner_self_employed?
```
timpeat, line 327: Can we extract the logic in line 327 so that it can be used both here and in the IncomeAssessement::AnswersValidator

app/validators/income_assessment/answers_validator.rb
```diff
@@ -82,6 +82,7 @@ def manage_without_income_complete?
       return true if record.income_payments.present? || record.income_benefits.present?
```
timpeat, line 82: Can we update this method to use the same logic as used in the decision tree (see note in review) above.

timpeat (commented):
Looking good. Can we extract a method -- perhaps to the type of means assessment module -- which captures logic for when we show the 'no_income' page. Perhaps #insufficient_declared_income or something like that.

timpeat (approved):
Perfect! Thank you :)

PR #1179, opened 2024-09-18
[CRIMAPP-249] Maintain original datestamp values

mseedat-moj:
## Description of change
Tracks changes to key client details which are then used by Caseworkers to determine if there is potential dodgy application based on datestamp manipulation

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-249

## Notes for reviewer
- Note introduction of the `DateStampContext` and related `DateStampContextType`
- Slight departure with existint rehydration: `application_rehydration.rb` does a simple `date_stamp_context: parent.date_stamp_context` allowing the type to cast to the correct `DateStampContext` object. You can see in the `application_rehydration_spec.rb` that `date_stamp_context: an_instance_of(LaaCrimeSchemas::Structs::DateStampContext)` rather than the `DateStampContext` attribute value in this codebase. It's a bit confusing but perhaps exposes the way identical type names between Schema and Apply are used to make it look like Structs and ActiveRecord Models are somewhat interchangeable. The alternative is to just explicitly set the `DateStampContext` in `application_rehydration.rb` like most of the other complex attributes/relations are.
- `services/date_stamper.rb` is in charge of generating the `DateStampContext` rather than having a callback in `CrimeApplication` or a line in `ClientDecisionTree`. This means that there is a danger the `crime_application.date_stamp_context` could go out of sync e.g. if the date_stamp was wiped due to some rule being triggered.
- `crime_application.date_stamp_context` is immutable but not deletable/nillable.
- There is no validation check to see if the `date_stamp_context` exists prior to submission - should there be?

N/A

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2024_09_11_142608) do
+ActiveRecord::Schema[7.0].define(version: 2024_09_17_084151) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
   enable_extension "plpgsql"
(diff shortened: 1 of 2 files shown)


timpeat (approved):
Looks Great!

PR #1181, opened 2024-09-18
[CRIMAPP-801] Hide home address question when home address known

hiboabd:
## Description of change
PR to hide home address question in the property form when the home address has already been set for a property as requested by design. This fixes a qa issue where users would be able to save more than one property as the client's home address via the save and come back later button.

## Link to relevant ticket

### After changes:

**When adding a property and a home address has not been set yet**

Screenshot 2024-09-18 at 13 25 39

**When adding a property and a home address has been set yet**

Screenshot 2024-09-18 at 13 26 04

**When editing a property that is has been set as the home address already**

Screenshot 2024-09-18 at 13 25 21

### DIFF
--- a/app/forms/steps/capital/property_form.rb
+++ b/app/forms/steps/capital/property_form.rb
@@ -33,9 +33,7 @@ module Steps
       validates_with CapitalAssessment::PropertyOwnershipValidator
 
       def persist!
-        is_home_address_attr = display_home_address_question? ? is_home_address : YesNoAnswer::NO
-
-        record.update(attributes.merge(is_home_address: is_home_address_attr))
+        record.update(attributes)
       end
 
       def before_save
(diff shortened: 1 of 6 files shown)


app/forms/steps/capital/property_form.rb
```diff
@@ -32,8 +32,6 @@ class PropertyForm < Steps::BaseFormObject
 
       validates_with CapitalAssessment::PropertyOwnershipValidator
 
-      validate :home_address, unless: :property_is_home_address?
-
       def persist!
         record.update(attributes)
       end
```
hiboabd, line 37: I considered defaulting `is_home_address` to `no` if the question wasn't displayed but as it wasn't explicitly selected by the user i've left it to be `nil`

timpeat (approved):

hiboabd:
Closing as we will consider an alternative solution (see ticket for comments)

PR #1189, opened 2024-09-25
[CRIMAPP-825] Other charges

EdwinKruglov:
## Description of change
Add new pages and database tables and fields to handle any other charges the client or partner might have in progress.

## Link to relevant ticket
[CRIMAPP-825](https://dsdmoj.atlassian.net/browse/CRIMAPP-825)

image
image
image

[CRIMAPP-825]: https://dsdmoj.atlassian.net/browse/CRIMAPP-825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/presenters/summary/sections/client_other_charge.rb
+++ b/app/presenters/summary/sections/client_other_charge.rb
@@ -1,0 +1,44 @@
+module Summary
+  module Sections
+    class ClientOtherCharge < Sections::BaseSection
+      def show?
+        kase.present? && kase.client_other_charge_in_progress.present? && super
+      end
+
+      def answers # rubocop:disable Metrics/MethodLength,Metrics/AbcSize
+        [
+          Components::ValueAnswer.new(
+            :client_other_charge_in_progress, kase.client_other_charge_in_progress,
+            change_path: edit_steps_case_other_charge_in_progress_path(subject: 'client')
+          ),
+          Components::FreeTextAnswer.new(
+            :client_other_charge_charge, kase.client_other_charge&.charge,
+            change_path: edit_steps_case_other_charge_path(subject: 'client'),
+          ),
+          Components::FreeTextAnswer.new(
+            :client_other_charge_hearing_court_name, kase.client_other_charge&.hearing_court_name,
+            change_path: edit_steps_case_other_charge_path(subject: 'client'),
+          ),
+          Components::DateAnswer.new(
+            :client_other_charge_next_hearing_date, kase.client_other_charge&.next_hearing_date,
+            change_path: edit_steps_case_other_charge_path(subject: 'client'),
+          ),
+        ].select(&:show?)
+      end
+
+      def title
+        if kase.partner_other_charge_in_progress.present?
+          I18n.t(
+            :client_other_charge_explicit,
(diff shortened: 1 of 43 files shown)


app/presenters/summary/sections/client_other_charge.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      def answers # rubocop:disable Metrics/MethodLength,Metrics/AbcSize
+        [
+          Components::ValueAnswer.new(
+            :client_other_charge_in_progress, kase.client_other_charge_in_progress,
+            change_path: edit_steps_case_other_charge_in_progress_path(subject: 'client')
+          ),
+          Components::FreeTextAnswer.new(
+            :client_other_charge_charge, kase.client_other_charge&.charge,
+            change_path: edit_steps_case_other_charge_in_progress_path(subject: 'client'),
```
hiboabd, line 16: The routing here should be `edit_steps_case_other_charge_path` and the same for the questions below

app/presenters/summary/sections/partner_other_charge.rb
```diff
@@ -0,0 +1,30 @@
...
+
+      def answers # rubocop:disable Metrics/MethodLength,Metrics/AbcSize
+        [
+          Components::ValueAnswer.new(
+            :partner_other_charge_in_progress, kase.partner_other_charge_in_progress,
+            change_path: edit_steps_case_other_charge_in_progress_path(subject: 'partner')
+          ),
+          Components::FreeTextAnswer.new(
+            :partner_other_charge_charge, kase.partner_other_charge&.charge,
+            change_path: edit_steps_case_other_charge_in_progress_path(subject: 'partner'),
```
hiboabd, line 16: Same here `edit_steps_case_other_charge_path`

db/schema.rb
```diff
@@ -292,6 +294,17 @@
     t.index ["charge_id"], name: "index_offence_dates_on_charge_id"
   end
 
+  create_table "other_charges", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
hiboabd, line 297: Could we have expanded on the existing charges table?

app/presenters/summary/sections/client_other_charge.rb
EdwinKruglov: Well spotted!

db/schema.rb
EdwinKruglov: You're right - I considered it when I started the ticket and decided to go this way because I didn't want to bloat the existing table with attributes that current charges don't have. I remember thinking that `charges` had a lot more attributes than it actually does for some reason... Somehow "other charges" felt distinct enough to warrant its own table. Looking at it now, I think reusing the existing table might be the better way to go, as long as I distinguish "current" charges from "other" charges.

Yeah, I feel way less convinced now about the approach I took originally 😅 I'll update to reuse `charges`.

hiboabd (approved):
Great work!

PR #1195, opened 2024-09-27
[CRIMAPP-1313] Reset dwp_benefit_check if dob changes

mseedat-moj:
## Description of change
Rests the DWP Benefit check result and is_means_tested fields if the date of birth changes.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1313

### DIFF
--- a/app/forms/steps/client/details_form.rb
+++ b/app/forms/steps/client/details_form.rb
@@ -24,13 +24,20 @@ module Steps
         errors.add(:date_of_birth, :client_under_ten) if date_of_birth.beginning_of_day > 10.years.ago
       end
 
-      def persist!
+      def persist! # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
         return true unless changed?
 
         ::CrimeApplication.transaction do
+          # CRIMAPP-1313 prevents resubmitted application date_of_birth changes to
+          # maintain original application means passporting
           if crime_application.resubmission? && changed?(:date_of_birth)
-            applicant.update!(confirm_dwp_result: nil)
-            crime_application.update!(is_means_tested: nil)
+            if applicant.over_18_at_date_stamp? && Applicant.under_18?(Time.zone.now, date_of_birth)
+              crime_application.update!(is_means_tested: 'yes')
+              applicant.update!(confirm_dwp_result: nil)
+            elsif applicant.under_18_at_date_stamp? && Applicant.over_18?(Time.zone.now, date_of_birth)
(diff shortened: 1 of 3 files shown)


app/forms/steps/client/details_form.rb
```diff
@@ -24,15 +24,28 @@ def date_of_birth_ten_or_more_years_ago
...
+        ::CrimeApplication.transaction do
+          # CRIMAPP-1313 prevents resubmitted application date_of_birth changes to
+          # maintain original application means passporting
+          if crime_application.resubmission? && changed?(:date_of_birth)
+            if applicant.over_18_at_date_stamp? && Applicant.under_18?(Time.zone.now, date_of_birth)
+              crime_application.update!(is_means_tested: 'yes')
+              applicant.update!(confirm_dwp_result: nil)
+            elsif applicant.under_18_at_date_stamp? && Applicant.over_18?(Time.zone.now, date_of_birth)
+              applicant.update!(confirm_dwp_result: nil)
+              crime_application.update!(is_means_tested: nil)
```
arthurashman, line 39: I think the logic is correct, but the attributes updated are not what we'd want to update. Give me a sec to figure out which I think it would be

app/models/concerns/type_of_application.rb
```diff
@@ -21,7 +21,8 @@ def returned?
   end
 
   def resubmission?
-    initial? && parent_id.present?
+    parent_id.present? &&
+      (application_type_eql?(:initial) || application_type_eql?(:change_in_financial_circumstances))
```
arthurashman, line 25: Any reason not to use `initial?` and `cifc?` ?

app/forms/steps/client/details_form.rb
arthurashman: So:
`is_means_tested` we want to leave alone for both (this is the Is the application means tested Q)

If they were under 18 at submission, they will not have seen:
NINO or any of the NINO / DWP confimation pages, so they will be nil and do not need resetting
I wonder if the rest of what we would need then would be covered by running the means_passporter 'Passporting::MeansPassporter.new(crime_application).call'

Bc at the moment, the validator knows something is wrong, but the decision tree isn't routing the user to the pages that they need to fulfil - and that's bc the application still has the passport in the passport array. 

I _think_ that going the other way (over 18 to under 18) that the passporter will be run and populate the initially empty array with the accurate passport. And I think that then handles the routing being correct.

What hasn't happened that I expected to is that the serialiser hasn't taken out the partner qs playback on the CYA.

Screenshot below, left is resubmission going from over18 to under18, right is initial sub uder18
image

PR #1200, opened 2024-10-01
[CRIMAPP-249] Show changed date stamp value tag in summary

mseedat-moj:
## Description of change
Displays a 'Changed after date stamp' tag for date stamp details that have changed.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-249

## Notes for reviewer
- Follows on from #1199 by adding the red 'Changed after date stamp' tag
- However departs from design by showing the tag under the actual value rather than on the right hand side where the 'Change' link is usually displayed
- This is because the x-gov govuk_summary_list `.with_action` only displays links. This would mean outputting the tag as an inline-block link which is linked to `#` (i.e. nowhere). That felt like a bad UX/accessibility
- The alternative is to override the `GovukComponent::SummaryListComponent::ActionComponent#render?` and `#call` methods so that `with_action` would output strings even if `href == nil`. That can be discussed if worth pursuing.
- The same issue does not arise in Review because the HTML is output explicitly in `.html.erb` files rather than generated using view components.

### Before changes:
Screenshot 2024-10-01 at 15 34 06

### After changes:

Screenshot 2024-10-01 at 22 35 56

#### When some date stamp context fields are missing for whatever reason

Screenshot 2024-10-02 at 12 04 06

#### When there is no date stamp context for historical applications
Screenshot 2024-10-02 at 12 04 16

Create a new application by entering client details, progressing the application until it is date stamped, then changing the applicant's first name or last name or date of birth. Then view the summary.

### DIFF
--- a/app/presenters/summary/sections/date_stamp_context.rb
+++ b/app/presenters/summary/sections/date_stamp_context.rb
@@ -9,29 +9,35 @@ module Summary
         false
       end
 
-      # rubocop:disable Metrics/MethodLength
+      # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
       def answers
-        [
+        answers = [
           Components::DateAnswer.new(
             :date_stamp, date_stamp_context.date_stamp,
             i18n_opts: { format: :datestamp },
+            show: date_stamp_context.date_stamp.present?,
           ),
           Components::ChangedDateStampAnswer.new(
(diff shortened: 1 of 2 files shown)


hiboabd (approved):
LGTM

PR #1230, opened 2024-10-28
[CRIMAPP-1389] Hide nino answer when not required

hiboabd:
## Description of change
Hides the nino answer row for appeal no changes and under 18 applications as the nino question is not asked
Previously the value `Not provided` was being displayed
The contact details section is also hidden for appeal no changes applications as this screen is not displayed whilst filling in an appeal no changes application

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1389

### Before changes:
Screenshot 2024-10-28 at 12 14 14

Screenshot 2024-10-28 at 13 14 32

### After changes:
Screenshot 2024-10-28 at 12 14 40

Screenshot 2024-10-28 at 13 09 41


spec/presenters/summary/sections/client_details_spec.rb
```diff
@@ -166,6 +170,22 @@
         expect(answers.count).to eq(7)
       end
     end
+
+    context 'when application is a appeal no changes application' do
+      let(:appeal_no_changes?) { true }
+
+      it 'has the correct rows' do
+        expect(answers.count).to eq(7)
```
EdwinKruglov, line 178: Is it worth asserting that specific questions/answers are present in this case? I think that would make for a more robust test than just relying on the count.

hiboabd: That's a good point, I've added checks to the question labels to assert the expected questions are shown

app/presenters/summary/sections/contact_details.rb
```diff
@@ -2,7 +2,7 @@ module Summary
   module Sections
     class ContactDetails < Sections::BaseSection
       def show?
-        applicant.present? && super
+        applicant.present? && !crime_application.appeal_no_changes? && super
```
hiboabd, line 5: I also spotted that the contact details section is displayed for appeal no changes applications when this wasn't asked and would enable providers to route to that question from the review screen when it's not meant to be part of the flow. I've added screenshots to the ticket

EdwinKruglov (approved):

PR #1240, opened 2024-11-06
[CRIMAPP-1453] Fix error with incomplete records

hiboabd:
## Description of change
Fixes a couple of sentry errors linked in the ticket. 

Users are being thrown errors on the evidence upload and review page when there are incomplete employment, Investment, co-defendant, saving and national saving certificate records. This is because the application's data is used to recreate a draft version of the application based on the schema. For some attributes, nil values were not permitted and so errors were being thrown. This is fixed at the schema level and with some `nil` value handling

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1453

## Notes for reviewer
Note: this means there will be greater reliance on validation to catch incomplete records before an app is submitted

Start an application and create incomplete employment, Investment, co-defendant, saving and national saving certificate records. Check that you can view the evidence upload and review screens and no error is presented

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,7 @@ gem 'laa-criminal-applications-datastore-api-client',
 
 gem 'laa-criminal-legal-aid-schemas',
     github: 'ministryofjustice/laa-criminal-legal-aid-schemas',
-    branch: 'CRIMAPP-1453-fix-sentry-bug'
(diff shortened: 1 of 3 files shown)


timpeat (approved):
THANK YOU!

PR #1271, opened 2024-11-25
CRIMAPP-1499 use post request for search.

timpeat:
## Description of change

Use POST for search requests.

## Link to relevant ticket

https://dsdmoj.atlassian.net/browse/CRIMAPP-1499

## Notes for reviewer

This is to avoid exposing PII in the search URL


app/controllers/application_searches_controller.rb
```diff
@@ -34,6 +34,10 @@ def search_params
     )
   end
 
+  def sorted_filter_params
+    { filter: @filter.params, sorting: @sorting.params }
+  end
+
```
timpeat, line 40: These params are used by the pagination buttons. (Note: they exclude none editable params.)

hiboabd (approved):

PR #1303, opened 2025-01-08
[CRIMAPP-1516] Property ownership validation

EdwinKruglov:
## Description of change
This changes adds validation in the capital section to check whether the client's usual residence is declared as a property. A new question is asked to allow the user to either add a new property with a matching home address or change where the client usually lives.

The changes are behind the feature flag `property_ownership_validation`.

## Link to relevant ticket
[CRIMAPP-1516](https://dsdmoj.atlassian.net/browse/CRIMAPP-1516)

### After changes:
image

1. Create an application and when asked 'Where does your client usually live?' answer 'In a property they own', 'In a property their partner owns', or 'In a property they and their partner both own'.
2. At the capital section, when asked about which assets the client (and/or partner) owns, answer 'They do not own any of these assets'.
3. You should see the new question.

The new question should also be triggered after answering 'No' to 'Do you want to add another asset?' if there is no asset that matches the home address.
The new question (and validation) shouldn't be triggered if the client has a partner and they live in a property owned by the partner and there's a conflict of interest.

[CRIMAPP-1516]: https://dsdmoj.atlassian.net/browse/CRIMAPP-1516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/steps/capital/usual_property_details/edit.html.erb
+++ b/app/views/steps/capital/usual_property_details/edit.html.erb
@@ -1,0 +1,22 @@
+<% title t('.heading', residence_ownership: t(".residence_ownership.#{@form_object.residence_ownership}")) %>
+<% step_header %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_error_summary(@form_object) %>
+
+    <span class="govuk-caption-xl"><%= t('steps.capital.caption') %></span>
+
+    <h1 class="govuk-heading-xl">
+      <%= t('.heading', residence_ownership: t(".residence_ownership.#{@form_object.residence_ownership}")) %>
+    </h1>
+
+    <p class="govuk-body"><%= t('.info', residence_ownership: t(".residence_ownership.#{@form_object.residence_ownership}")) %></p>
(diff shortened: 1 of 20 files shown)


app/views/steps/capital/usual_property_details/edit.html.erb
```diff
@@ -0,0 +1,27 @@
...
+
+    <p class="govuk-body"><%= t('.info', residence_ownership: t(".residence_ownership.#{@form_object.residence_ownership}")) %></p>
+    <%= simple_format(@form_object.home_address, class: 'govuk-body') %>
+
+    <%= step_form @form_object do |f| %>
+      <%= f.govuk_radio_buttons_fieldset :usual_property_details, legend: { text: t('.next_action') } do %>
+        <% @form_object.choices.each_with_index do |option, index| %>
+          <%= f.govuk_radio_button :action, option, link_errors: index.zero? %>
+        <% end %>
+      <% end %>
```
hiboabd, line 22: The error message is missing under the question 'What do you want to do next?'

hiboabd (approved):
Looks good, just left one comment regarding the error message presentation

PR #1333, opened 2025-02-10
Update validation for property form 

lucas-shaw:
## Description of change
The field "Does anyone else own part of the property?" now displays the error if the total percentages equal 100%. Currently, the percentage fields both display the error.

The error messages for the applicant and partner percentages have been updated for clarity.

This removes repetition from the error messages.

## Link to relevant ticket
https://dsdmoj.atlassian.net/jira/software/c/projects/CRIMAPP/boards/1375?selectedIssue=CRIMAPP-1598

### Before changes:
Screenshot 2025-02-10 at 16 01 07
Screenshot 2025-02-10 at 16 01 31

### After changes:
Screenshot 2025-02-10 at 15 56 43
Screenshot 2025-02-11 at 11 25 37

Enter values that equal 100 for applicant and partners ownership percentage
Select "Yes" for other owner


app/validators/capital_assessment/property_ownership_validator.rb
```diff
@@ -5,37 +5,46 @@ class PropertyOwnershipValidator < ActiveModel::Validator
...
+
+    def other_owners?
       form.has_other_owners&.yes?
     end
 
-    def error_key(form)
-      return :invalid_when_other_owners if other_owners?(form)
+    def error_key
+      return :invalid_when_other_owners if other_owners?
+      return :invalid_with_no_partner unless partner_included?
```
hiboabd, line 48: This method might only be required for the `percentage_applicant_owned`  attribute when you are distinguishing between the error message when there is a partner vs. no partner. You could hardcode the error key for the other attributes

config/locales/en/errors.yml
```diff
@@ -966,15 +966,16 @@ en:
               blank: Enter how much is left to pay on the mortgage
               not_a_number: How much there is left to pay on the mortgage must be a number, like 50
             percentage_applicant_owned:
-              invalid: Percentages entered need to total 100%
+              invalid: Percentages entered need to total 100% - check percentage owned by partner
```
hiboabd, line 969: This should say `by client` instead of partner, double check the others in this file as there are a couple of mixups

hiboabd (commented):
Nice work! a couple of comments from me

hiboabd (approved):

PR #1336, opened 2025-02-12
[CRIMAPP-1641] has_no_properties not updated in property validation flow

EdwinKruglov:
## Description of change
This fixes an issue where `capital_details.has_no_properties` is not updated if the user chooses to create a residential property as a result of a failed property ownership validation. This causes their assets to not show on Review, where `has_no_properties` is used to determine whether to show their assets.

## Link to relevant ticket
[CRIMAPP-1641](https://dsdmoj.atlassian.net/browse/CRIMAPP-1641)

[CRIMAPP-1641]: https://dsdmoj.atlassian.net/browse/CRIMAPP-1641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/capital.rb
+++ b/app/models/capital.rb
@@ -71,9 +71,10 @@ class Capital < ApplicationRecord
     super if requires_full_capital?
   end
 
-  def usual_property_details_required?
+  def usual_property_details_required? # rubocop:disable Metrics/AbcSize
     return false unless FeatureFlags.property_ownership_validation.enabled?
     return false unless requires_full_capital?
+    return false if crime_application.applicant.residence_type.blank?
 
     residence_type = ResidenceType.new(crime_application.applicant.residence_type)
     if residence_type.owned?
--- a/app/models/capital.rb
+++ b/app/models/capital.rb
@@ -52,7 +52,10 @@ class Capital < ApplicationRecord
   end
 
   def has_no_properties
-    super if requires_full_capital?
+    return unless requires_full_capital?
+    return unless properties.empty?
+
+    super
   end
 
   def has_no_savings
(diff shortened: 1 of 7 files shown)


spec/system/applying/property_ownership_validation_spec.rb
```diff
@@ -0,0 +1,301 @@
+require 'rails_helper'
+
+RSpec.describe 'Apply for Criminal Legal Aid with cross-question property ownership validation' do
```
hiboabd, line 3: Good call!

app/models/capital.rb
```diff
@@ -71,9 +71,10 @@ def partner_has_premium_bonds
     super if requires_full_capital?
   end
 
-  def usual_property_details_required?
+  def usual_property_details_required? # rubocop:disable Metrics/AbcSize
     return false unless FeatureFlags.property_ownership_validation.enabled?
     return false unless requires_full_capital?
+    return false if crime_application.applicant.residence_type.blank?
```
hiboabd, line 77: Why was this added? 🤔

hiboabd (approved):
LGTM just one question. Also it would be good to to create a application with the bug in staging before merging this in to check this fixes it

app/models/capital.rb
EdwinKruglov: There was an instance where a user skipped ahead and go to this question before answering the `residence_type` question, and the page threw an error because of this. This doesn't fix the potentially bigger issue of users being able to jump ahead like that, but thought I'd add this to prevent it from happening again. Here's more about what happened: https://mojdt.slack.com/archives/C056U956ESH/p1738932998077209

timpeat (approved):
🙇

PR #1341, opened 2025-02-17
Remove trailing .00's throughout the service

lucas-shaw:
## Link to relevant ticket
https://dsdmoj.atlassian.net/jira/software/c/projects/CRIMAPP/boards/1375?selectedIssue=CRIMAPP-1592

### After changes:
Screenshot 2025-02-17 at 15 10 08

### DIFF
--- a/app/presenters/summary/components/money_answer.rb
+++ b/app/presenters/summary/components/money_answer.rb
@@ -2,7 +2,7 @@ module Summary
   module Components
     class MoneyAnswer < BaseAnswer
       def answer_text
-        number_to_currency(value.to_s) if value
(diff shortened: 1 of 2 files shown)


EdwinKruglov (commented):
`number_to_currency` and `number_to_percentage` both accept the `strip_insignificant_zeros` option. I think that would be a simpler change than calling `sub` on the string. Source: https://api.rubyonrails.org/v8.0/classes/ActiveSupport/NumberHelper.html#method-i-number_to_percentage

**Edit:** this would remove _all_ zeros after the decimal separator, e.g. .50 would become .5, which might not be desirable.

EdwinKruglov (approved):

PR #1353, opened 2025-02-26
[CRIMAPP-1659] age passport always considers date stamp

timpeat:


app/models/person.rb
```diff
@@ -21,12 +19,6 @@ def has_passporting_benefit?
     BenefitType.passporting.map(&:value).include?(benefit_type&.to_sym)
   end
 
-  def over_18_at_date_stamp?
-    datum = crime_application&.date_stamp || Time.zone.now
-
-    datum.in_time_zone('London').to_date - 18.years >= date_of_birth
-  end
-
```
timpeat, line 29: This method was unused.

app/models/concerns/person_with_dob.rb
```diff
@@ -1,9 +0,0 @@
-module PersonWithDob
```
timpeat, line 1: This method does not consider the date_stamp it's use was the source of the regression.  There term under18 has been deprecated in favour of the already existing `age_passported?`

spec/models/person_spec.rb
```diff
@@ -86,63 +86,6 @@
     end
   end
 
-  describe '#over_18_at_date_stamp?' do
```
timpeat, line 89: Method removed from Person

app/services/passporting/means_passporter.rb
```diff
@@ -27,7 +27,7 @@ def not_means_tested?
     end
 
     def age_passported?
-      parent_was_age_passported? || applicant_under18?
+      parent_was_age_passported? || age_passported_at_datestamp_or_now?
     end
```
timpeat, line 32: 👇  I believe `parent_was_age_passported?` is not longer required however this PR is already big enough, so will ticket up looking into that another time.

app/validators/partner_details/answers_validator.rb
```diff
@@ -3,8 +3,8 @@ class AnswersValidator < BaseAnswerValidator
     include TypeOfMeansAssessment
 
     def applicable?
+      return false if partner_information_not_required?
```
timpeat, line 6: 👇 I'm not sure why we have the `return false unless record` guard clause here, but again this is a big PR for a bug fix.

app/validators/sections_completeness_validator.rb
```diff
@@ -30,10 +30,11 @@ def validate # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Me
   delegate :client_details_complete?, :passporting_benefit_complete?, to: :crime_application
 
   def partner_detail_complete?
-    return true if applicant&.under18? || not_means_tested? || appeal_no_changes?
```
timpeat, line 33: Uses the validator#applicable instead of repeating the logic in this class 👉

EdwinKruglov (approved):

PR #1365, opened 2025-03-11
announce table label and result count

lucas-shaw:
## Description of change
When entering the table, the screenreader should announce this and how many search results there are.

## Link to relevant ticket
https://dsdmoj.atlassian.net/jira/software/c/projects/CRIMAPP/boards/1375?selectedIssue=CRIMAPP-1532

### Before changes:
Screenshot 2025-03-11 at 08 58 37

### After changes:
Screenshot 2025-03-11 at 08 56 20

Activate MacOS Voiceover using CMD + Fn + F5
You can tab into the "copy reference number" link then press CTRL + OPT + SPACE (MacOs)

### DIFF
--- a/app/views/application_searches/search.html.erb
+++ b/app/views/application_searches/search.html.erb
@@ -21,7 +21,7 @@
 
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full app-table-container">
-      <table class="govuk-table app-table" aria-roledescription="." aria-label="<%= t('search.label') %> <%= t('search.total', count: @search.total) %>">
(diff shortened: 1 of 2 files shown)


app/views/application_searches/search.html.erb
```diff
@@ -21,7 +21,7 @@
 
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full app-table-container">
-      <table class="govuk-table app-table">
+      <table class="govuk-table app-table" aria-roledescription="." aria-label="<%= t('search.table_label') %> <%= t('search.total', count: @search.total) %>">
```
lucas-shaw, line 24: Added `aria-roledescription="."` because otherwise the announcement appends the default "...table" at the end, which would read "Entering application search results table, n search results **_table_**..."

PR #1366, opened 2025-03-11
[CRIMAPP-1676] Disregard unrequired benefit check values when result has changed

hiboabd:
## Description of change

Caseworkers in review reported an issue where they could not determine whether a benefit check had been performed for an application. This occurred as the application's passporting benefit check result had changed from undetermined to confirmed during the drafting of the application. However, the answers to the questions that were answered as part of the undetermined flow (e.g. has benefit evidence) were still being used to calculate the benefit check status on submission. This was resulting in a final benefit check status of `nil` which impacted the view of the passporting benefit check result on apply and review

This PR fixes this and ignores the values for these fields if a confirmed result is obtained when calculating the benefit check status to be presented in apply and review.  

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-1676

## Notes for reviewer
I have added a helper method to mock the benefit check status result in the system specs. We could now add system specs for all of the benefit check routes
 
### Before changes:

(Note screenshot taken from ticket)
image (14)

### After changes:

Screenshot 2025-03-11 at 09 27 41

The scenarios that need testing are: 
1. The benefit checker returned an undetermined result, the confirm dwp result, confirm client details and has benefit evidence question was answered and then the passporting benefit checker was re-ran later with a confirmed result. `Note: this will require updating the answers in the rails console`
2. The benefit checker was down, the has benefit evidence question was answered and then the passporting benefit checker was re-ran later with a confirmed result. `Note: this requires turning the mock benefit checker service off and then on via your local env.development.local file`

### DIFF
--- a/spec/support/capybara_helpers.rb
+++ b/spec/support/capybara_helpers.rb
@@ -196,4 +196,12 @@ module CapybaraHelpers # rubocop:disable Metrics/ModuleLength
 
     visit completed_crime_application_path(application_id)
   end
+
+  # mock_result format 'Yes' or 'No'
+  def mock_benefit_check(mock_result)
+    fixture_name = mock_result == 'Yes' ? 'successful.xml' : 'unsuccessful.xml'
+
+    stub_request(:post, 'http://benefit-checker/?wsdl')
+      .to_return(body: file_fixture("benefit_checker_responses/#{fixture_name}").read)
+  end
 end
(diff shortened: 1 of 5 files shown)


timpeat (approved):
Looks great. Thank you :)

spec/support/capybara_helpers.rb
```diff
@@ -196,4 +196,12 @@ def return_submitted_application
 
     visit completed_crime_application_path(application_id)
   end
+
+  # mock_result format 'Yes' or 'No'
+  def mock_benefit_check(mock_result)
+    fixture_name = mock_result == 'Yes' ? 'successful.xml' : 'unsuccessful.xml'
```
timpeat, line 202: ❤️

PR #1372, opened 2025-03-14
[CRIMAPP-1634] Ensure income payment completness

EdwinKruglov:
## Description of change
This change fixes an issue where the user may leave the employment income question in an incomplete state and then face errors when attempting to revisit the question or access the evidence upload page.

Incomplete `IncomePayment` records will now be excluded from calculations and will not be serialized, which causes an exception when accessing the evidence upload page.

## Link to relevant ticket
[CRIMAPP-1634](https://dsdmoj.atlassian.net/browse/CRIMAPP-1634)

[CRIMAPP-1634]: https://dsdmoj.atlassian.net/browse/CRIMAPP-1634?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/income.rb
+++ b/app/models/income.rb
@@ -31,6 +31,7 @@ class Income < ApplicationRecord
     # disregard payments for people not included in means assessemnt
     # as well as obsolete payment types
     scope = crime_application.income_payments
+                             .completed
                              .owned_by(ownership_types)
                              .not_of_type(obsolete_employed_income_payment_types)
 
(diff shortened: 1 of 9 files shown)


app/models/income.rb
```diff
@@ -40,7 +40,7 @@ def income_payments
                           not_working_owners, employed_income_payment_types)
     end
 
-    @income_payments = scope
+    @income_payments = scope.select(&:complete?)
```
EdwinKruglov, line 43: Not a big fan of calling an array method on what is otherwise a query different of scopes. This would prevent further scope chaining if we wanted it. Happy to hear other ideas!

timpeat (approved):

PR #1397, opened 2025-04-11
CRIMAPP-1712 only check clamav on startup probe

timpeat:
## Description of change

- Web app only checks clamav working on startupProbe.
- Remove clamav from readynesProbe

## Notes for reviewer

Moving to startup probe to ensure requests continue to the app if clamav down and web app has already started.

### DIFF
--- a/config/kubernetes/staging/deployment.tpl
+++ b/config/kubernetes/staging/deployment.tpl
@@ -58,7 +58,7 @@ spec:
           periodSeconds: 10
         startupProbe:
           httpGet:
-            path: /ping
+            path: /startupz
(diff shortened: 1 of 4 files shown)


config/kubernetes/staging/deployment.tpl
```diff
@@ -58,7 +58,7 @@ spec:
           periodSeconds: 10
         startupProbe:
           httpGet:
-            path: /ping
+            path: /startz
```
EdwinKruglov, line 61: Is the correct path? Should it not be `/startupz`?

PR #1430, opened 2025-06-10
[CRIMAPP-1776] WIP display undetermined status

hiboabd:


app/services/dwp/benefit_check_service.rb
```diff
@@ -18,7 +19,9 @@ def self.passporting_benefit?(applicant)
       result = call(applicant)
       return result if result.nil?
 
-      result[:benefit_checker_status].casecmp('yes').zero?
+      return 'Undetermined' unless result[:benefit_checker_status].in?(BENEFIT_CHECKER_RESPONSES)
```
hiboabd, line 22: Defaults to undetermined for other statuses e.g. deceased

app/services/dwp/benefit_check_status_service.rb
```diff
@@ -22,24 +22,28 @@ def call
...
-    def undetermined_status
-      return BenefitCheckStatus::UNDETERMINED.to_s if benefit_evidence_forthcoming?
+    def undetermined
+      benefit_check_subject.dwp_response == 'Undetermined' || (benefit_check_subject.confirm_dwp_result == 'no' && benefit_evidence_forthcoming?)
+    end
 
-      BenefitCheckStatus::NO_RECORD_FOUND.to_s if means_assessment_as_benefit_evidence?
+    def not_in_receipt
+      benefit_check_subject.dwp_response == 'No' || (benefit_check_subject.confirm_dwp_result == 'no' && means_assessment_as_benefit_evidence?)
     end
```
hiboabd, line 40: Would it be better have a method for `dwp_response` on the model that would calculate this? My only concern would be that if the app was returned it would be submitted with the calculated attribute `dwp_response` when that was not actually generated from the checker 🤔

app/services/decisions/dwp_decision_tree.rb
```diff
@@ -113,14 +113,16 @@ def determine_dwp_result_page(person)
 
       DWP::UpdateBenefitCheckResultService.call(person)
```
timpeat, line 114: Can we explore moving this update out of the decision tree?

PR #1581, opened 2025-10-29
Fix incorrect Welsh currency formatting (.000 decimals)

lucas-shaw:
## Description of change
In Welsh, 000 decimal number amounts should not be shown. e.g £12,345.000. In local development, this issue does not seem to exist (see ticket).

In Staging/Prod, the cause might be a result of missing precision: 2 being properly defined. `config.i18n.fallbacks = true` allows missing Welsh keys (such as cy.number.format.precision) to fall back to the English locale. But the English locale yml didn’t explicitly define `precision` either, so it results in the additional .000

## Link to relevant ticket
https://dsdmoj.atlassian.net/jira/software/c/projects/CRIMAPP/boards/1375?assignee=712020%3A3d805bbe-8fa6-48b0-b8bd-f2d3612405d6&selectedIssue=CRIMAPP-1966


hiboabd (approved):
🤞

PR #1583, opened 2025-11-13
[CRIMAPP-1980] Fix application search when switching locales

lucas-shaw:
## Link to relevant ticket
https://dsdmoj.atlassian.net/jira/software/c/projects/CRIMAPP/boards/1375?selectedIssue=CRIMAPP-1980


config/routes.rb
```diff
@@ -317,6 +317,7 @@ def show_step(name)
 
   resource :application_searches, path: 'application-searches', only: [:new] do
     post :search, on: :collection
+    get :search, on: :collection
```
timpeat: To avoid having personal information exposed in the query, we should avoid GET requests on search if possible.

timpeat (changes requested):

PR #1603, opened 2025-12-16
[CRIMAPP-2049] Ensure nino is not mandatory for non means applications

hiboabd:
## Description of change

Fixes bug where non means applications could not be submitted where the NINO option was selected but the NINO was not provided by updating the logic in the section completeness validator. See ticket for more details.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-2049

### DIFF
--- a/app/views/home/index.en.html.erb
+++ b/app/views/home/index.en.html.erb
@@ -5,7 +5,7 @@
     <%= render partial: 'shared/flash_banner' %>
 
     <h1 class="govuk-heading-xl"><%= service_name %></h1>
-    <p class="govuk-body">This service is available from 7am until 9:30pm.</p>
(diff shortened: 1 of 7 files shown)


app/validators/client_details/answers_validator.rb
```diff
@@ -58,7 +58,14 @@ def has_nino_complete?
       return false if applicant.has_nino.blank? && applicant.has_arc.blank?
       return true if applicant.has_nino == 'no'
```
timpeat: Was there a reason the Provider did not state 'no' to the has_nino question? This would have enabled them to proceed I'd thought?

timpeat (approved):

PR #1707, opened 2026-04-16
Add frame and form actions to CSP

timpeat:
## Description of change

Add explicit CSP for frame_ancestors and form_action 

## Link to relevant ticket

https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/issues/1655

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -14,15 +14,18 @@ Rails.application.configure do
     policy.font_src    :self
 
     if FeatureFlags.google_analytics.enabled?
-      policy.connect_src :self, 'https://ga.jspm.io', 'https://*.google-analytics.com', 'https://*.analytics.google.com', 'https://*.googletagmanager.com'
+      policy.connect_src :self, 'https://ga.jspm.io', 'https://*.google-analytics.com',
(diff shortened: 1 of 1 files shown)


config/initializers/content_security_policy.rb
```diff
@@ -14,23 +14,26 @@
     policy.font_src    :self
 
     if FeatureFlags.google_analytics.enabled?
-      policy.connect_src :self, 'https://ga.jspm.io', 'https://*.google-analytics.com', 'https://*.analytics.google.com', 'https://*.googletagmanager.com'
+      policy.connect_src :self, 'https://ga.jspm.io', 'https://*.google-analytics.com',
+                         'https://*.analytics.google.com', 'https://*.googletagmanager.com'
```
AndyDouglasGit, line 18: just a quick, non blocking, question -  this looks to be the same. Do we have some sort of linting that is being applied sometimes and not othertimes not that is causing changes?

AndyDouglasGit (approved):

config/initializers/content_security_policy.rb
timpeat: Yes, our linter is not applied on all files but my editor applies linting on all files

PR #1713, opened 2026-04-20
Switch from Renovate to Dependabot

EdwinKruglov:
## Description of change
- switch from Renovate to Dependabot for all dependency updates as Renovate continues to not function properly
- add timezones to all scheduled workflows


.github/dependabot.yml
EdwinKruglov, line 1: Each check will run weekly on Monday at a random time by default. Updates to security vulnerabilities will continue to be raised immediately.

timpeat (approved):
Thank you :)

PR #1736, opened 2026-04-27
CRIMAPP-2190 only show and submit relevant details on crown court appeals

timpeat:
## Description of change

Appeal to Crown Court--fix redundant data appearing on Crime Review and CYA screens

When an Appeal to Crown Court application was returned to a provider and they changed a key earlier answer (e.g. "was there a previous application?"), redundant data from the previous journey continued to appear on the Check Your Answers and Review Application screens. This affected several scenarios where the case type, passporting status, or means assessment path had effectively changed on return.

Responsibility for deciding which fields were relevant was spread across forms, presenters, Crime Review and validators. Forms would reset dependent fields on save, but only for the specific step being edited--if a provider changed an earlier answer without re-visiting downstream steps, stale data remained visible and was included in the submission.

Attribute relevance logic has been centralised in the model layer. Conditionally relevant attributes now return nil when their controlling condition is not met, regardless of what is stored in the database. This means the CYA and review screens, the submission serializer, and validators all see a consistent view of the application state without each needing to replicate the relevance rules. This approach has been used elsewhere, and whilst not perfect is an improvement on resetting attributes or having logic in the view layer.

This approach is documented in ADR 0004.

## Link to relevant ticket

[CRIMAPP-2190](https://dsdmoj.atlassian.net/browse/CRIMAPP-2190)

## Notes for reviewer

Once deployed, the conditional display logic in the review service that caused the original bug can be removed (see laa-review-criminal-legal-aid#1021).

[CRIMAPP-2190]: https://dsdmoj.atlassian.net/browse/CRIMAPP-2190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -1,6 +1,7 @@
 feature_flags:
   developer_tools:
-    local: true
+    development: true
+    test: false
     staging: false
     production: false
   google_analytics:
--- a/app/forms/steps/client/case_type_form.rb
+++ b/app/forms/steps/client/case_type_form.rb
@@ -25,20 +25,9 @@ module Steps
       end
 
       def persist!
-        return true unless changed?
+        return unless changed?
 
-        kase.update(
-          attributes.merge(
-            # The following are dependent attributes that need to be reset
-            appeal_lodged_date: nil,
-            appeal_with_changes_details: nil,
-            appeal_original_app_submitted: nil,
-            appeal_financial_circumstances_changed: nil,
-            appeal_reference_number: nil,
-            appeal_maat_id: nil,
-            appeal_usn: nil
-          )
-        )
+        kase.update(attributes.merge)
       end
     end
   end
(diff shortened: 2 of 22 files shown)


config/settings.yml
```diff
@@ -1,6 +1,7 @@
 feature_flags:
   developer_tools:
-    local: true
+    development: true
```
EdwinKruglov, line 3: Was the change here from `local` to `development` intentional?

EdwinKruglov (approved):
This is brilliant! Thanks Tim 🙏

config/settings.yml
timpeat: Yes intentional :)

I did not want the dev tools available in the test environment . This turns them off in test and keeps them in development .

PR #1753, opened 2026-05-15
CRIMAPP-2143 court name and offence type always english

AndyDouglasGit:
## Description of change

Court name and offence type elements will always have lang attribute set to en.

## Link to relevant ticket

https://dsdmoj.atlassian.net/jira/software/c/projects/CRIMAPP/boards/1375?selectedIssue=CRIMAPP-2143

## Notes for reviewer

- This is here for critique, this will stop me progressing too much in wrong direction.
- I need to find out if offence class is also always english (this was something new when I merged main into my branch)
- I'm about to add a test for court hearing answer

image

image

### Before changes:

These elements had no lang attribute; therefore utilising the lang attribute cascaded from the HTML element

Review an application with court hearing(s) and offence(s) in it.

### DIFF
--- a/app/presenters/summary/components/free_text_answer.rb
+++ b/app/presenters/summary/components/free_text_answer.rb
@@ -1,6 +1,7 @@
 module Summary
   module Components
     class FreeTextAnswer < BaseAnswer
+      # ?
       def answer_text
         simple_format(value.presence || absence_answer)
       end
--- a/app/presenters/summary/components/free_text_answer.rb
+++ b/app/presenters/summary/components/free_text_answer.rb
@@ -1,7 +1,6 @@
 module Summary
   module Components
     class FreeTextAnswer < BaseAnswer
-      # ?
       def answer_text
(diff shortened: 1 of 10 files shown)


app/presenters/summary/components/free_text_answer.rb
```diff
@@ -1,6 +1,7 @@
 module Summary
   module Components
     class FreeTextAnswer < BaseAnswer
+      # ?
```
AndyDouglasGit: oops - will remove

EdwinKruglov (commented):
I think this is going in the right direction 👍 The only thing I’d say is the naming of `CourtNameTypeAnswer` should be different. Since all this class does is wrap the answer value in a `p` tag with a `lang` attribute, it’s quite a general class which could easily be used elsewhere.

`FreeTextAnswer` also handles the absence of an answer with `absence_answer` - I would make sure that this isn’t needed in your new class.

EdwinKruglov (approved):

PR #1777, opened 2026-05-29
CRIMAPP-2130 announce sortable column headers

AndyDouglasGit:
## Description of change

The sortable column headers in the tables should work better with screen readers. 

This is done with the following covered in the PR:

- For each tab (In Progress, Submitted, Returned, Decided), the displayed table contains a visually hidden <caption> element that states “Column headers with buttons are sortable."
- Voiceover for each button clearly states its purpose in the form of "Sort by <Fieldname>" (strings added for English and Welsh)
( Screen reading tools should no longer describe the icon in the form of “down pointing black pointer name up pointing black triangle” etc)
- Each sortable column heading is now a link rather than a button - to the user this appears identical and has the same functionality but is semantically clearer.

## Link to relevant ticket

https://dsdmoj.atlassian.net/browse/CRIMAPP-2130

### Before changes:

image

Using voiceover (command F5) - note that the arrows are voiced i.e. "“down pointing black pointer" and this makes it very disruptive
image

### After changes:

Note:
1.  the **caption** element
2.  forms have been removed from the header
3.  form buttons are now links

image

Note:

- Voice over no longer mentions the icons
- Voice over now states the header is for sorting

image

In **Apply** select **Your applications**

- For regression select different tabs and sort by different fields to ensure that table functionality has not regressed since the change
- Inspect an element on the table and examine the console to see that a **Caption** exists explaining that the columns are sortable.
- Inspect an element on the table and examine the console to see that the forms have been removed and that the headers are now links.
- Press command f5 (mac) to get voice over. Tab through the sortable headers and voice will be "sort by <column name>"
- Repeat the above with search results


timpeat (approved):
👏

PR #1793, opened 2026-06-11
CRIMAPP-2167: Add frozen income or assets subject question

amazimbe:
## Description of change
Add a follow-up question to capture whether frozen income or assets
belong to the client, partner, or both.

Includes:
- income and capital model changes
- Apply routes, controllers, forms, and views
- Review support for displaying the answer
- decision tree updates
- translations
- controller, form, and workflow tests

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-2167
### Before changes:

image-20260413-125548
image-20260413-125548

### After changes:

Screenshot 2026-06-08 at 10 31 06

Screenshot 2026-06-18 at 14 41 17

### Prerequisites

Create or use an application that:

* Requires a means assessment.
* Includes a partner in the means assessment.

### Income journey

1. Navigate to **Income > Does your client [or their partner] have any income, savings or assets under a restraint or freezing order?**
2. Select **Yes** and continue.
3. Verify the new question **"Who does the restraint or freezing order relate to?"** is displayed.
4. Verify the available options are:

   * The client
   * Their partner
   * Both
5. Select each option in turn and continue.
6. Verify the selected answer is displayed correctly on the Income review page.

### Capital journey

1. Navigate to **Capital > Does your client [or their partner] have any income, savings or assets under a restraint or freezing order?**
2. Select **Yes** and continue.
3. Verify the new question **"Who does the restraint or freezing order relate to?"** is displayed.
4. Select each option in turn and continue.
5. Verify the selected answer is displayed correctly on the Capital review page.

### Validation

1. Select **Yes** to the restraint/freezing order question.
2. Continue without answering the follow-up question.
3. Verify a validation error is shown.

### Regression checks

1. Create or use an application where a partner is **not** included in the means assessment.
2. Select **Yes** to the restraint/freezing order question.
3. Verify the new follow-up question is **not** displayed and the journey continues as before.

### DIFF
--- a/app/services/decisions/capital_decision_tree.rb
+++ b/app/services/decisions/capital_decision_tree.rb
@@ -52,6 +52,8 @@ module Decisions
       when :partner_trust_fund
         after_trust_fund
       when :frozen_income_savings_assets_capital
+        after_frozen_income_savings_assets_capital
+      when :frozen_income_or_assets_subject
         edit(:answers)
       when :answers
         edit('/steps/evidence/upload')
@@ -162,6 +164,15 @@ module Decisions
         return edit(:property_owners, property_id: property)
       end
       edit(:properties_summary)
+    end
+
+    def after_frozen_income_savings_assets_capital
+      if crime_application.capital.has_frozen_income_or_assets == YesNoAnswer::YES.to_s &&
+         include_partner_in_means_assessment?
+        edit(:frozen_income_or_assets_subject)
+      else
+        edit(:answers)
+      end
     end
 
     def create_property_owner?
--- a/app/services/decisions/income_decision_tree.rb
+++ b/app/services/decisions/income_decision_tree.rb
@@ -32,6 +32,8 @@ module Decisions
         after_income_before_tax
       when :frozen_income_savings_assets
         after_frozen_income_savings_assets
+      when :frozen_income_or_assets_subject
+        after_frozen_income_or_assets_subject
       when :client_owns_property
         after_client_owns_property
       when :has_savings
@@ -220,7 +222,12 @@ module Decisions
       edit(:frozen_income_savings_assets)
     end
 
+    def after_frozen_income_or_assets_subject
+      after_extent_of_means_determined
+    end
+
     def after_frozen_income_savings_assets
+      return edit(:frozen_income_or_assets_subject) if has_frozen_assets? && include_partner_in_means_assessment?
       return after_extent_of_means_determined if extent_of_means_assessment_determined?
 
       edit(:client_owns_property)
--- a/config/locales/en/errors.yml
+++ b/config/locales/en/errors.yml
@@ -546,6 +546,12 @@ en:
           attributes:
             has_frozen_income_or_assets:
               blank: Select yes if %{subject} has any income, savings or assets under a restraint or freezing order
+        steps/income/frozen_income_or_assets_subject_form:
+          &FROZEN_INCOME_OR_ASSETS_SUBJECT_VALIDATION
+          attributes:
+            frozen_income_or_assets_subject:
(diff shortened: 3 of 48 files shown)


app/services/decisions/capital_decision_tree.rb
```diff
@@ -164,6 +166,14 @@ def after_property_address
       edit(:properties_summary)
     end
 
+    def after_frozen_income_savings_assets_capital
+      if crime_application.capital.has_frozen_income_or_assets == YesNoAnswer::YES.to_s
+        edit(:frozen_income_or_assets_subject)
```
hiboabd: There is an additional condition to display the screen if the applicant has a partner. It should not be shown if the applicant does not have a partner

app/services/decisions/income_decision_tree.rb
```diff
@@ -220,7 +222,12 @@ def after_income_before_tax
       edit(:frozen_income_savings_assets)
     end
 
+    def after_frozen_income_or_assets_subject
+      after_extent_of_means_determined
+    end
+
     def after_frozen_income_savings_assets
+      return edit(:frozen_income_or_assets_subject) if has_frozen_assets?
```
hiboabd: Same here. It's if the client has frozen assets AND the client has a partner

config/locales/en/steps.yml
```diff
@@ -656,6 +659,9 @@ en:
             one: You have added %{count} asset
             other: You have added %{count} assets
       frozen_income_savings_assets_capital: *FROZEN_INCOME_ASSETS_PAGE_TEXT
+      frozen_income_or_assets_subject:
+        edit:
+          page_title: Who does the restraint or freezing order relate to?
```
hiboabd: Rather than duplicating this from 302–304. Could we extract this into a YAML anchor and reuse it here, similar to *FROZEN_INCOME_ASSETS_PAGE_TEXT, so the page title is defined in one place?

config/locales/en/helpers.yml
```diff
@@ -1284,6 +1289,11 @@ en:
           other_relationship: What is their relationship?
       steps_capital_frozen_income_savings_assets_form:
         has_frozen_income_or_assets_options: *YESNO
+      steps_capital_frozen_income_or_assets_subject_form:
+        frozen_income_or_assets_subject_options:
+          client: The client
+          partner: Their partner
+          client_and_partner: Both the client and their partner
```
hiboabd: Rather than duplicating this from 875–878. Could we extract this into a YAML anchor and reuse it here, similar to has_frozen_income_or_assets_options: *YESNO, so the options are defined in one place?

config/locales/en/errors.yml
```diff
@@ -1125,7 +1130,11 @@ en:
           attributes:
             add_investment:
               inclusion: Select yes if you want to add another investment
-
+        steps/capital/frozen_income_or_assets_subject_form:
+          attributes:
+            frozen_income_or_assets_subject:
+              blank: Select whether the income, savings or assets under a restraint or freezing order belong to your client or their partner
+              inclusion: Select whether the income, savings or assets under a restraint or freezing order belong to your client or their partner
```
hiboabd: Rather than duplicating this from 549–553. Could we extract this into a YAML anchor and reuse it here so the error text is defined in one place? Similar to &FROZEN_INCOME_ASSETS_VALIDATION

db/schema.rb
hiboabd, line 1: There's a lot of changes in this file. I think you may need to rerun your migrations and recommit the file

hiboabd (commented):
Nice work so far, i've left some comments and there's some additional changes it would be good to talk through

EdwinKruglov (approved):

PR #1829, opened 2026-07-08
CRIMAPP-2219: Extend CookieOverflow diagnostics for omniauth.params

amazimbe:
## Description of change
The initial CookieOverflow investigation identified that the session was not being bloated by the OpenID Connect security values (state, nonce or PKCE verifier) as originally suspected.

Instead, the vast majority of the session payload was contained within the `omniauth.params` session entry, which is populated by OmniAuth from the incoming request's query parameters.

Extend the temporary diagnostic middleware to log the structure of `omniauth.params`, including the parameter names, value types and estimated sizes, while continuing to avoid logging the parameter values themselves.

This additional instrumentation will help determine which query parameter(s) are responsible for the oversized session and whether the root cause lies in the application, the OpenID Connect strategy, or the broader authentication flow.

## Link to relevant ticket
CRIMAPP-2219

### DIFF
--- a/spec/middleware/session_cookie_overflow_logger_spec.rb
+++ b/spec/middleware/session_cookie_overflow_logger_spec.rb
@@ -7,7 +7,17 @@ RSpec.describe SessionCookieOverflowLogger do
   subject(:middleware) { described_class.new(app, logger:) }
 
   let(:logger) { instance_spy(Logger) }
-  let(:session) { { 'large_key' => 'x' * 100, 'small_key' => 'abc' } }
+
+  let(:session) do
+    {
+      'large_key' => 'x' * 100,
+      'small_key' => 'abc',
+      'omniauth.params' => {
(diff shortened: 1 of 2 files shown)


spec/middleware/session_cookie_overflow_logger_spec.rb
```diff
@@ -7,7 +7,18 @@
   subject(:middleware) { described_class.new(app, logger:) }
 
   let(:logger) { instance_spy(Logger) }
-  let(:session) { { 'large_key' => 'x' * 100, 'small_key' => 'abc' } }
+  # let(:session) { { 'large_key' => 'x' * 100, 'small_key' => 'abc' } }
```
EdwinKruglov: Can this commented-out line be removed?

EdwinKruglov (dismissed):

EdwinKruglov (approved):

PR #1863, opened 2026-07-22
CRIMAPP-2137 Unify multi-part date error messaging and summary links

amazimbe:
## Description of change
Improve how date validation errors are rendered inline and in the GOV.UK error summary so users consistently see segment-specific feedback.

- Add DateFieldErrors presenter module logic to:
  - map recognised date error types to day/month/year segments
  - build joined inline messages when multiple segments are invalid
  - generate segment-targeted summary links for one or more recognised segment errors
  - preserve default summary behaviour when no recognised segment errors exist
- Add DateFieldErrorSummaryPresenter and wire it into StepsHelper#govuk_error_summary
- Update FormBuilderHelper#date_input to inject combined inline error message at the error-collection level (before render), and restore the full original errors collection afterwards via copy! to avoid ordering/type/option drift
- Update MultiparamDateValidator to avoid adding generic :invalid when attribute-specific errors already exist
- Add/extend specs to cover:
  - multiple invalid date segments inline message
  - full error restoration after rendering
  - single invalid segment keeping normal inline behaviour
  - summary links for single day/month/year segment errors
  - summary links for multiple segment errors

Verification:
- COVERAGE=false bundle exec rspec spec/helpers/steps_helper_spec.rb spec/forms/steps/case/is_client_remanded_form_spec.rb

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-2137

### After changes:
Screenshot 2026-07-22 at 17 23 50

Updated todo list

Manually test by opening a form with a required date field (for example, *Is client remanded?*), selecting the option that makes the date mandatory, and submitting invalid combinations: first make all parts invalid (e.g. day 32, month 13, short year) and confirm the inline error shows combined messages while the error summary shows separate links to day/month/year inputs; then test single-part failures (only day invalid, only month invalid, only year invalid) and confirm the summary link goes directly to the specific input and inline messaging remains normal for single errors; finally test a non-date validation error on the same page to confirm standard summary behavior is unchanged.

### DIFF
--- a/app/helpers/form_builder_helper.rb
+++ b/app/helpers/form_builder_helper.rb
@@ -11,10 +11,34 @@ module FormBuilderHelper
   end
 
   def date_input(attribute_name, opts = {}, &block)
-    govuk_date_field(attribute_name, segments:, segment_names:, **opts, &block)
+    combined_error_message = DateFieldErrors.joined_inline_message(object, attribute_name)
+
+    return govuk_date_field(attribute_name, segments:, segment_names:, **opts, &block) unless combined_error_message
+
+    with_inline_error_message(attribute_name, combined_error_message) do
+      govuk_date_field(attribute_name, segments:, segment_names:, **opts, &block)
+    end
   end
 
   private
+
+  def with_inline_error_message(attribute_name, message)
+    errors = object.errors
+    return yield if errors.where(attribute_name).blank?
+
+    with_restored_errors(errors) do
+      errors.delete(attribute_name)
+      errors.add(attribute_name, message)
+      yield
+    end
+  end
+
+  def with_restored_errors(errors)
+    original_errors = errors.dup
+    yield
+  ensure
+    errors.copy!(original_errors)
+  end
 
   def submit_button(i18n_key, opts = {}, &block)
     govuk_submit I18n.t("helpers.submit.#{i18n_key}"), **opts, &block
--- a/app/presenters/date_field_error_summary_presenter.rb
+++ b/app/presenters/date_field_error_summary_presenter.rb
@@ -1,0 +1,9 @@
+class DateFieldErrorSummaryPresenter
+  def initialize(form_object)
+    @form_object = form_object
+  end
+
+  def formatted_error_messages
+    DateFieldErrors.error_summary_messages(@form_object)
+  end
+end
--- a/app/presenters/date_field_errors.rb
+++ b/app/presenters/date_field_errors.rb
@@ -1,0 +1,52 @@
+module DateFieldErrors
+  ERROR_SEGMENTS = {
+    blank_day: :day,
+    invalid_day: :day,
+    blank_month: :month,
+    invalid_month: :month,
+    blank_year: :year,
+    invalid_year: :year,
+  }.freeze
+  DATE_INPUT_SEGMENTS = { day: '3i', month: '2', year: '1i' }.freeze
+
+  module_function
+
+  def error_summary_messages(form_object)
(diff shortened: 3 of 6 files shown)


app/helpers/form_builder_helper.rb
```diff
@@ -11,11 +11,35 @@ def continue_button(primary: :save_and_continue, secondary: :save_and_come_back_
   end
 
   def date_input(attribute_name, opts = {}, &block)
-    govuk_date_field(attribute_name, segments:, segment_names:, **opts, &block)
+    combined_error_message = DateFieldErrors.joined_inline_message(object, attribute_name)
```
amazimbe, line 14: This keeps the standard `govuk_date_field` path for normal cases, but when multiple date-segment errors are present it temporarily injects a combined inline message at the error-collection level so the rendered field shows unified feedback, then restores the original errors immediately after rendering.

app/helpers/form_builder_helper.rb
```diff
@@ -11,11 +11,35 @@ def continue_button(primary: :save_and_continue, secondary: :save_and_come_back_
...
+    return govuk_date_field(attribute_name, segments:, segment_names:, **opts, &block) unless combined_error_message
+
+    with_inline_error_message(attribute_name, combined_error_message) do
+      govuk_date_field(attribute_name, segments:, segment_names:, **opts, &block)
+    end
   end
 
   private
 
+  def with_inline_error_message(attribute_name, message)
```
amazimbe, line 25: temporarily swaps in a combined message for the date field while rendering, then always puts the original errors back afterwards.

app/presenters/date_field_error_summary_presenter.rb
```diff
@@ -0,0 +1,9 @@
+class DateFieldErrorSummaryPresenter
```
amazimbe, line 1: lets the GOV.UK error summary use our custom date-error mapping, so date segment errors can be shown as separate summary links to day, month, and year inputs.

app/presenters/date_field_errors.rb
```diff
@@ -0,0 +1,52 @@
+module DateFieldErrors
```
amazimbe, line 1: centralises date-error handling by mapping validator error types to day/month/year segments, building combined inline messages, and generating segment-specific error-summary links while keeping default behavior for non-date errors.

app/presenters/date_field_errors.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+  end
+
+  def date_segment_id(form_object, attribute, segment)
+    [
+      form_object.model_name.param_key,
+      attribute,
+      DATE_INPUT_SEGMENTS.fetch(segment),
+    ].join('_')
+  end
```
mooktakim-moj, line 45: One issue i found was the error link for the day error didn't link to the day input field.
The `day` input sets `link_errors: true` in gov formbuilder, the others have `link_errors: false`.

Possible fix:

```ruby
  def date_segment_id(form_object, attribute, segment)
    # Use the existing anchor id for the day field
    return field_error_id(form_object, attribute) if segment == :day

    [
      form_object.model_name.param_key,
      attribute,
      DATE_INPUT_SEGMENTS.fetch(segment),
    ].join('_')
  end

  def field_error_id(form_object, attribute)
    [form_object.model_name.param_key, attribute, 'field-error'].join('-').tr('_', '-')
  end
```

amazimbe: Applied, thanks.

mooktakim-moj (approved):

PR #1869, opened 2026-07-27
Apply: Update Banner to utilise MOJ header

AndyDouglasGit:
## Description of change

image

Utilise a gem for the local header to replace than the local written version. 

This allows us to upgrade our gov and moj dependencies.

## Link to relevant ticket

https://dsdmoj.atlassian.net/jira/software/c/projects/CRIMAPP/boards/1375?selectedIssue=CRIMAPP-2281

## Notes for reviewer

1. The header should function as the previous version and appear basically the same (there will be minor nuances in sizing etc) and:
2. The organisation name is no longer underlined upon hover (this brings the app inline with others)
3. The bottom border of the header should have a different color depending on environment:
        dev = green
        staging = orange
        live = blue (this was inadvertently removed when the local header was created)

Run it; click on the various options on the header.

### DIFF
--- a/app/views/application/_header.html.erb
+++ b/app/views/application/_header.html.erb
@@ -1,21 +1,6 @@
-<%= govuk_header(homepage_url: root_path, classes: "govuk-header-#{HostEnv.env_name}") do |header|
-  header.with_custom_logo do %>
-    <%= render partial: 'layouts/moj_crest' %>
-  <% end %>
-  <% header.with_product_name(name: t('layouts.header.organisation_name')) %>
-
+<%= moj_header(organisation_name: t('layouts.header.organisation_name'), url: '#') do |header| %>
   <% if signed_in? %>
-    <div class="govuk-header__content">
-      <nav class="govuk-header__navigation" aria-label="Menu">
-        <ul class="govuk-header__navigation-list">
-          <li class="govuk-header__navigation-item app-header__auth-user" aria-role="paragraph">
-            <%= current_provider.display_name %>
-          </li>
-          <li class="govuk-header__navigation-item">
-            <%= link_to t('layouts.header.sign_out'), provider_omniauth_logout_path, class: 'govuk-header__link', aria: { role: 'button' } %>
-          </li>
-        </ul>
-      </nav>
-    </div>
+    <%= header.with_navigation_item(text: current_provider.display_name, href: '#', current: true) %>
+    <%= header.with_navigation_item(text: t('layouts.header.sign_out'), href: provider_omniauth_logout_path) %>
   <% end %>
 <% end %>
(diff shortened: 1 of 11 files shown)


app/views/application/_header.html.erb
```diff
@@ -1,21 +1,6 @@
-<%= govuk_header(homepage_url: root_path, classes: "govuk-header-#{HostEnv.env_name}") do |header|
-  header.with_custom_logo do %>
-    <%= render partial: 'layouts/moj_crest' %>
-  <% end %>
-  <% header.with_product_name(name: t('layouts.header.organisation_name')) %>
-
+<%= moj_header(organisation_name: t('layouts.header.organisation_name'), url: '#') do |header| %>
```
amazimbe: url: '#'. The gem uses this directly as the href, so clicking the organisation name no longer navigates anywhere. The old implementation used homepage_url: root_path.

app/views/steps/provider_step/_header.html.erb
```diff
@@ -1,18 +1,4 @@
...
-        <li class="govuk-header__navigation-item app-header__auth-user">
-          <%= current_provider.display_name %>
-        </li>
-        <li class="govuk-header__navigation-item">
-          <%= link_to t('layouts.header.sign_out'), provider_omniauth_logout_path, class: 'govuk-header__link', aria: { role: 'button' } %>
-        </li>
-      </ul>
-    </nav>
-  </div>
+<%= moj_header(organisation_name: t('layouts.header.organisation_name'), url: '#') do |header| %>
```
amazimbe: same issue with #

app/views/application/_header.html.erb
```diff
@@ -1,21 +1,6 @@
...
-          <li class="govuk-header__navigation-item app-header__auth-user" aria-role="paragraph">
-            <%= current_provider.display_name %>
-          </li>
-          <li class="govuk-header__navigation-item">
-            <%= link_to t('layouts.header.sign_out'), provider_omniauth_logout_path, class: 'govuk-header__link', aria: { role: 'button' } %>
-          </li>
-        </ul>
-      </nav>
-    </div>
+    <%= header.with_navigation_item(text: current_provider.display_name, href: '#', current: true) %>
```
amazimbe: I think header.with_navigation_item(..., current: true) causes the gem to emit aria-current="page" on the provider's display name link — every page will announce this as "current page" to screen readers. The old implementation rendered it as plain text with no ARIA role.

Probably pass current: false (or omit it) on the display-name navigation item.

amazimbe (changes requested):
Nothing major just the donuts about # in url and current: true. Please confirm.

app/views/application/_header.html.erb
AndyDouglasGit: used root_path - and this emulates previous behaviour exactly

app/views/steps/provider_step/_header.html.erb
AndyDouglasGit: used root_path - and this emulates previous behaviour exactly

app/views/application/_header.html.erb
AndyDouglasGit: You are correct - this was taken from the example in the docs but I agree with your point as current page is not useful.

amazimbe (approved):

PR #1879, opened 2026-07-29
CRIMAPP-2134 Announce currency symbol to screen reader users

amazimbe:
## Description of change
The pound (£) prefix on monetary input fields was marked aria-hidden="true", so screen reader users heard only the field label with no indication that the value should be entered in pounds.

Override govuk_number_field in FormBuilderHelper to append a visually hidden span containing "Amount in pounds" (Welsh: "Swm mewn punnoedd") to the label whenever prefix_text: '£' is used. The span is injected via label_with_currency_hint, which resolves the label text through the standard I18n lookup before merging the hidden span in. All other label options (tag, size, etc.) are preserved.

The fix is centralised in the form builder helper so all currency fields across the service benefit automatically, without changes to individual views.

## Link to relevant ticket
https://dsdmoj.atlassian.net/browse/CRIMAPP-2134

### After changes:
Screenshot 2026-07-29 at 15 13 40

Navigate to any page with a monetary input field — for example, the savings account balance page under Capital, or any employment income or outgoings payment page. Enable a screen reader (VoiceOver on macOS, NVDA or JAWS on Windows, or TalkBack on Android) and tab to the currency input. The screen reader should announce the field label followed by "Amount in pounds" before reading the input role and any error. Confirm that the visible £ prefix is still displayed but not read out a second time.

### DIFF
--- a/app/helpers/form_builder_helper.rb
+++ b/app/helpers/form_builder_helper.rb
@@ -20,6 +20,16 @@ module FormBuilderHelper
     end
   end
 
+  # Overrides the GOV.UK form builder's govuk_number_field to inject a
+  # visually hidden "Amount in pounds" span into the label when prefix_text
+  # is '£'. This ensures screen reader users receive the same currency context
+  # that sighted users get from the visible pound symbol prefix.
+  def govuk_number_field(attribute_name, prefix_text: nil, label: {}, **, &block)
+    label = label_with_currency_hint(attribute_name, label) if prefix_text == '£' && label.is_a?(Hash)
+
+    super
+  end
+
   private
 
   def with_inline_error_message(attribute_name, message)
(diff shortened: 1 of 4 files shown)


app/helpers/form_builder_helper.rb
```diff
@@ -20,6 +20,16 @@ def date_input(attribute_name, opts = {}, &block)
     end
   end
 
+  # Overrides the GOV.UK form builder's govuk_number_field to inject a
+  # visually hidden "Amount in pounds" span into the label when prefix_text
+  # is '£'. This ensures screen reader users receive the same currency context
+  # that sighted users get from the visible pound symbol prefix.
+  def govuk_number_field(attribute_name, prefix_text: nil, label: {}, **, &block)
```
mooktakim-moj, line 27: I almost feel like there should be a `govuk_currency_field` or `govuk_money_field` 😅

mooktakim-moj (approved):

app/helpers/form_builder_helper.rb
amazimbe: I know what you mean but it looks like a conscious design decision to go with govuk_number_field with a prefix_text option that can be used for £ prefix.

PR #1892, opened 2026-08-05
Dependabot config: group each ecosystem into one PR

AndyDouglasGit:
## Description of change

We’ve been seeing a high volume of Dependabot PRs, which is becoming difficult to manage.
This change groups updates by ecosystem (e.g. npm, GitHub Actions) into a single PR, rather than creating one PR per dependency.

It’s a small change, but should help reduce noise and improve overall workflow. 

Unlike what is currently done on review  this will group together all minor and patch versions with the major versions being placed into separate PRs

## Link to relevant ticket

No ticket

## Notes for reviewer

This has already been done for review.


.github/dependabot.yml
```diff
@@ -6,24 +6,40 @@ updates:
...
     directory: "/"
     schedule:
       interval: "weekly"
     cooldown:
       default-days: 7
     open-pull-requests-limit: 10
+    groups:
+      docker:
+        patterns:
+          - "*"
```
amazimbe: newline missing here too.

.github/dependabot.yml
```diff
@@ -6,24 +6,40 @@ updates:
...
+      github-actions:
+        patterns:
+          - "*"
   - package-ecosystem: "bundler"
     directory: "/"
     schedule:
       interval: "weekly"
     cooldown:
       default-days: 7
     open-pull-requests-limit: 10
```
amazimbe, line 22: For nom and bundler probably best to group only minor and patch upgrades. Major upgrades can remain in individual PRs because they are more likely to contain breaking changes and are easier to review and troubleshoot separately.
groups:
  bundler-minor-and-patch:
    patterns:
      - "*"
    update-types:
      - "minor"
      - "patch"

amazimbe (commented):
I recommend grouping only minor and patch upgrades and leaving major upgrades as individual PRs. Major upgrades are more likely to introduce breaking changes, and combining several of them would make review and CI failures harder to diagnose.

It may also be worth adding separate groups for security updates using applies-to: security-updates, as the groups added here apply only to regular version updates by default. Security updates would otherwise continue to generate individual PRs.

```
groups:
  bundler-security-updates:
    applies-to: security-updates
    patterns:
      - "*"
```

amazimbe (approved):

PR #1895, opened 2026-08-06
CRIMAPP-2139 Move focus to inputs when activating error summary links

mooktakim-moj:
## Description of change

Fixes an accessibility defect where activating a GOV.UK error summary link did not move keyboard/screen-reader focus to the input that needs correcting. The summary link's `href` pointed at a `…-<attribute>-field-error` id that was never rendered on the page, leaving the user at the top of the page.

The GOV.UK formbuilder only renders that anchor id when a radio/checkbox is built with `link_errors: true`. The affected pages either omitted it or attached the error to an attribute (`:base`) with no matching element.

Changes:
- **NINO** (`steps/shared/nino/edit`) — add `link_errors: true` to the first radio in the `has_arc_or_nino` group.
- **Cannot check benefit status** (`steps/dwp/cannot_check_benefit_status/edit`) — add `link_errors: index.zero?` to the first `will_enter_nino` radio.
- **Deductions** (client + partner) — the "nothing selected" validation error was added to `:base`, whose summary link (`#…-base-field-error`) had no target. Moved it to `:types` in `DeductionsValidator` (shared by both forms) and set `link_errors: index.zero?` on the first checkbox so the link resolves to it. Renamed the paired `base:` → `types:` locale keys (en + cy) so the message still resolves. (The nested amount/frequency summary links already resolved correctly.)

Tests:
- New view specs for all four templates asserting every error-summary link target exists on the page (verified they fail without the fix).
- Updated the two deductions form specs (`of_kind?(:base, …)` → `of_kind?(:types, …)`).
- RuboCop, erb_lint, and the affected unit/view/controller specs pass, plus the NINO and employed-with-partner system specs.

## Link to relevant ticket

https://dsdmoj.atlassian.net/browse/CRIMAPP-2139

## Notes for reviewer

- Focus/screen-reader behaviour is inherently a JS/AT path not exercised by automated tests; the new view specs assert the structural precondition (the summary link's `href` matches a real element id on the page), which is what enables the browser to move focus.
- Out of scope: the same `:base` + `link_errors: true` pattern exists on other checkbox pages (income payments/benefits, outgoings payments) and is a candidate for a follow-up ticket.

### After changes:

_No visual change — the error summary and inline errors look the same; only the link target/focus behaviour changes._

1. Start (or resume) an application and go to the **"Does your client have a National Insurance number?"** page.
2. Leave the question unanswered and select **Save and continue**.
3. In the error summary, activate the **"Select whether…"** link (click, or Tab to it and press Enter).
4. Confirm the page scrolls to and focus moves to the first radio button in the group.
5. Repeat for the **"We cannot check your client's benefits status"** page (leave the yes/no question blank).
6. Repeat for the **Deductions** page (client and partner): submit with nothing selected and confirm the summary link moves focus to the first deduction checkbox.

### DIFF
--- a/app/views/steps/dwp/cannot_check_benefit_status/edit.html.erb
+++ b/app/views/steps/dwp/cannot_check_benefit_status/edit.html.erb
@@ -28,8 +28,8 @@
   <div class="govuk-grid-column-two-thirds">
     <%= step_form @form_object do |f| %>
     <div class="govuk-!-margin-bottom-7">
-      <% @form_object.choices.each do |option| %>
-        <%= f.govuk_radio_button :will_enter_nino, option %>
+      <% @form_object.choices.each_with_index do |option, index| %>
+        <%= f.govuk_radio_button :will_enter_nino, option, link_errors: index.zero? %>
       <% end %>
     </div>
 
--- a/app/validators/deductions_validator.rb
+++ b/app/validators/deductions_validator.rb
@@ -13,7 +13,7 @@ class DeductionsValidator < ActiveModel::Validator
 
     return unless record.types.empty?
 
-    record.errors.add(:base, :none_selected) if record.employment.has_no_deductions.blank?
+    record.errors.add(:types, :none_selected) if record.employment.has_no_deductions.blank?
   end
 
   private
--- a/app/views/steps/income/client/deductions/edit.html.erb
+++ b/app/views/steps/income/client/deductions/edit.html.erb
@@ -9,8 +9,8 @@
     <%= step_form @form_object do |f| %>
       <%= f.govuk_check_boxes_fieldset :deductions, legend: { tag: 'h1', size: 'xl' } do |x| %>
 
-        <% @form_object.ordered_deductions.each do |type| %>
-          <%= f.govuk_check_box :types, type, link_errors: true do %>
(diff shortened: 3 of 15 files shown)


app/views/steps/dwp/cannot_check_benefit_status/edit.html.erb
```diff
@@ -28,8 +28,8 @@
   <div class="govuk-grid-column-two-thirds">
     <%= step_form @form_object do |f| %>
     <div class="govuk-!-margin-bottom-7">
-      <% @form_object.choices.each do |option| %>
-        <%= f.govuk_radio_button :will_enter_nino, option %>
+      <% @form_object.choices.each_with_index do |option, index| %>
```
amazimbe, line 31: This fixes the will_enter_nino error-summary target on this page, but the same radio group still exists unchanged in cannot_submit_without_nino/edit.html.erb. That template still renders f.govuk_radio_button :will_enter_nino, option without link_errors, so submitting it blank will keep producing an error-summary link to the missing ...-field-error anchor. Perhaps apply the same each_with_index / link_errors: index.zero? fix there as well (and ideally cover it with a matching view spec).

amazimbe (commented):
Just that one file I believe needs updating too.

amazimbe (approved):

PR #1924, opened 2026-08-25
CRIMAPP-2276 Add slipstream_audit_selections table and model

amazimbe:
## Description of change
Introduces persistent tracking of applications selected for slipstream audit.

This table captures:
- crime_application_id: FK to the application (unique constraint prevents re-selection)
- selected_at: timestamp when the selection occurred
- sample_rate: snapshot of the sample rate in effect at selection time (for audit trail)

The unique index on crime_application_id ensures that:
- Once an application is selected, it cannot be re-selected
- The selection persists even if the provider edits the application
- On resubmission after being returned, the selection is retained

Model includes dependent: :destroy to clean up selections when applications
are purged from the system.

Introduces persistent tracking of crime applications selected for slipstream
audit. The unique constraint on crime_application_id prevents re-selection and
ensures the selection persists across provider edits and resubmissions.

Includes sample_rate snapshot at time of selection for audit trail purposes.

## Link to relevant ticket
https://dsdmoj.atlassian.net/jira/software/c/projects/CRIMAPP/boards/1375?selectedIssue=CRIMAPP-2346


db/migrate/20260824174246_create_slipstream_audit_selections.rb
```diff
@@ -0,0 +1,10 @@
+class CreateSlipstreamAuditSelections < ActiveRecord::Migration[8.1]
+  def change
+    create_table :slipstream_audit_selections, id: :uuid, default: -> { 'gen_random_uuid()' } do |t|
```
EdwinKruglov: I think we might want to include the actual selection outcome - whether this application was selected for the audit or not. Otherwise there are two ways to interpret the absence of this record for a given application:
1. The selection check has not been run on that application
2. The selection check was run and the application was not selected

We'll want to be able to differentiate between those two states.

amazimbe: @EdwinKruglov I think we should be careful with the word selection(s) here as it could have a double meaning.

Correct me if I'm wrong, my understanding is that slipstream_audit_selections stores applications that have passed random selection. However, some of these might not end up in the slipstream audit as they could fail other eligibility checks. I'm proposing to add a boolean column confirmed. A true value here would mean the selection has passed all eligibility checks and that application will go for slipstream auditing.

On the naming, I thing slipstream_audit_candidates is a better name and I'll change it to that if no one objects. It's more precise about what the table actually represents — these are candidates for slipstream auditing, not selections. The word "selections" implied they were all chosen, which is clearly not the case.

EdwinKruglov (dismissed):

mooktakim-moj (approved):

